PR #2, opened 2018-06-08
Feature pick

rfhaque:
Pick implementation


src/ResourceManager.cpp
```diff
@@ -211,6 +211,16 @@ void ResourceManager::copy(void* dst_ptr, void* src_ptr, size_t size)
   op->transform(src_ptr, &dst_ptr, src_alloc_record, dst_alloc_record, size);
 }
 
+void ResourceManager::transfer(void* dst_ptr, void* src_ptr, size_t size, std::shared_ptr<umpire::strategy::AllocationStrategy>& dst_alloc_strategy, std::shared_ptr<umpire::strategy::AllocationStrategy>& src_alloc_strategy)
```
davidbeckingsale: Why do we need an extra operation?

PR #3, opened 2018-06-29
Moved source files from src/ to src/umpire to make headers work out w/ submodules

ajkunen:
Moved source files from src/ to src/umpire to make headers work out without COPY_HEADERS

fixed a number of pathnames to work when brigning chai+umpire into other projects as nested submodules


davidbeckingsale:
@ajkunen I have a branch with this capability and no conflicts.

PR #5, opened 2018-07-17
Feature/cuda const

nanzifan:
Add CUDA constant memory support memory allocator named DEVICE_CONST.
Add one example that use constant memory allocator.

### DIFF
--- a/src/umpire/ResourceManager.cpp
+++ b/src/umpire/ResourceManager.cpp
@@ -162,16 +162,6 @@ ResourceManager::getAllocator(resource::MemoryResourceType resource_type)
   return Allocator(m_memory_resources[resource_type]);
 }
 
-void
-ResourceManager::registerAllocator(const std::string& name, Allocator allocator)
-{
-  if (isAllocator(name)) {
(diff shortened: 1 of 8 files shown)


src/umpire/ResourceManager.cpp
```diff
@@ -150,6 +162,16 @@ ResourceManager::getAllocator(resource::MemoryResourceType resource_type)
   return Allocator(m_memory_resources[resource_type]);
 }
 
+void
+ResourceManager::registerAllocator(const std::string& name, Allocator allocator)
```
davidbeckingsale: Why have you moved this method?

davidbeckingsale: Okay - could you fix this please?

nanzifan: No problem.

davidbeckingsale: Thanks :)

nanzifan: I put them back to original place. ^.^

PR #10, opened 2018-08-22
fix #8

### LINKED ISSUE #8: thread safe DynamicPool for cuda unified memory
Hi,

I'm trying to use Umpire to manage CUDA unified memory. The program is multi-threaded, so I try to make a thread-safe allocator.  
Here is my code
```
auto um_dynamic_pool = rm.makeAllocator<umpire::strategy::DynamicPool>(
          "UMDynamicPool", rm.getAllocator("UM"));
auto thread_safe_um_dynamic_pool =
          rm.makeAllocator<umpire::strategy::ThreadSafeAllocator>(
              "ThreadSafeUMDynamicPool", rm.getAllocator("UMDynamicPool"));
```
The `thread_safe_um_dynamic_pool` allocator is stored in a global class and used by all threads to allocate and deallocate cuda unified memory.

However, I found that my program will stochastically throw an error when deallocating unified memory when using multiple threads.
Sometimes it throws the error here https://github.com/LLNL/Umpire/blob/develop/src/umpire/util/AllocationMap.cpp#L121
sometimes here
https://github.com/LLNL/Umpire/blob/develop/src/umpire/util/AllocationMap.cpp#L75

I try to generate a nice log file by lock the `logMessage` function, but the lock hides the errors and I can't trigger the same error again. 

I attached the original log file, but the output is in a mess because of multi-threading. Sorry about the inconvenience.
[umpire_log.txt](https://github.com/LLNL/Umpire/files/2295187/umpire_log.txt)

So, am I doing the correct thing if I want to use a thread-safe dynamic pool for CUDA unified memory?
Would you take a look at this in case there is a real issue?

Thanks,

pchong90:
please review


src/umpire/util/AllocationMap.cpp
```diff
@@ -87,25 +87,34 @@ AllocationMap::remove(void* ptr)
 AllocationRecord*
 AllocationMap::findRecord(void* ptr)
 {
-  auto record = m_records.atOrBefore(reinterpret_cast<uintptr_t>(ptr));
 
-  if (record.value) {
-    void* parent_ptr = reinterpret_cast<void*>(record.key);
-    auto alloc_record =
-      reinterpret_cast<Entry>(record.value->back());
+  AllocationRecord* alloc_record;
```
evaleev: should initialize to nullptr? seems that previous behavior was to return nullptr if not the record was not found

davidbeckingsale:
Looks good, thanks for the fix!

PR #13, opened 2018-08-27
Enable code coverage for codecov

davidbeckingsale:


davidbeckingsale:
@mcfadden8 - our AllocationMap "performance" test is failing. Do you think we should just get rid of it, or skip it for Travis builds?

mcfadden8:
Why is the performance test failing?  I'd like to understand this better and understand whether there are viable alternatives.

codecov.yml
```diff
@@ -0,0 +1,3 @@
+ignore:
+  - "src/umpire/tpl"
```
mcfadden8: Won't this also ignore simPool?  I think we want to add that right?

tests/unit/util/CMakeLists.txt
```diff
@@ -32,12 +32,15 @@ blt_add_test(
   NAME allocation_map_tests
   COMMAND allocation_map_tests)
 
-blt_add_executable(
-  NAME allocation_map_performance_tests
-  SOURCES allocation_map_performance_tests.cpp
-  DEPENDS_ON umpire_util gtest
-  OUTPUT_DIR ${UMPIRE_TEST_OUTPUT_DIR})
+# TODO: Make this test a relative performance test
```
mcfadden8: Can we only disable this when running coverage tests?

davidbeckingsale: That was previously done - it didn't help.

codecov.yml
davidbeckingsale: Yes. I can change that.

davidbeckingsale:
👍

PR #21, opened 2018-10-29
Add noexcept specifier to functions that don't throw

davidbeckingsale:


davidbeckingsale:
@mcfadden8 - I think this is good to go 😄

PR #23, opened 2018-11-20
Turn on warnings as errors by default, and fix some warnings

davidbeckingsale:


davidbeckingsale:
Couple of problems with the ROCm backend, but I'm going to skip over those for now, since they were warnings in AMD's code.

PR #24, opened 2018-11-29
CMake bug workaround when using clang on OSX

gzagaris:
This commit fixes an issue wherein setting CMAKE_CXX_STANDARD to 11
was not adding the "-std=c++11" flag to the command line when
using Clang on OS X. This is a known CMake issue reported in
https://gitlab.kitware.com/cmake/cmake/issues/15943. The suggested
workaround is to set the CMake policy CMP0025.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@
 # Please also see the LICENSE file for MIT license.
 ##############################################################################
 
+cmake_policy(SET CMP0025 NEW)
 cmake_policy(SET CMP0048 NEW)
 cmake_policy(SET CMP0057 NEW)
 


gzagaris:
@davidbeckingsale, alternatively moving the calls to `cmake_minimum_required` above the `project()` call, would also fix this.  It is actually recommended practice to call the `cmake_minimum_required` before invoking the `project()` command so that all policy settings are established before making any other cmake calls (see the note here: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html). This also explains why I haven't come across this issue in other projects e.g., RAJA and Axom. However, setting the policy seems to be more aligned with the current setup in place.

PR #32, opened 2018-12-06
Gaurd replay abi usage against ROCM compilation

mcfadden8:
This first step simply guards against replay usage on AMD platform.

### DIFF
--- a/tools/replay.cpp
+++ b/tools/replay.cpp
@@ -277,9 +277,11 @@ int main(int ac, char** av)
   if ( ac != 2 )
     Replay::usage_and_exit( "Incorrect number of program arguments" );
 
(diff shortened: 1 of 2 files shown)


tools/replay.cpp
```diff
@@ -277,9 +277,11 @@ int main(int ac, char** av)
   if ( ac != 2 )
     Replay::usage_and_exit( "Incorrect number of program arguments" );
 
+#if !defined(UMPIRE_ENABLE_ROCM)
```
davidbeckingsale: Is this bit necessary right now?

mcfadden8: Yes, otherwise the replay tests will fail on ARM.

davidbeckingsale: Makes sense 👍

PR #33, opened 2018-12-08
Moved Replay from util library to umpire library since it uses objects

mcfadden8:
that are in the umpire library (Allocator).

### DIFF
--- a/src/umpire/Allocator.cpp
+++ b/src/umpire/Allocator.cpp
@@ -16,6 +16,7 @@
 
 #include "umpire/ResourceManager.hpp"
 #include "umpire/util/Macros.hpp"
+#include "umpire/Replay.hpp"
 
 #if defined(UMPIRE_ENABLE_STATISTICS)
 #include "umpire/util/StatisticsDatabase.hpp"
--- a/src/umpire/CMakeLists.txt
+++ b/src/umpire/CMakeLists.txt
@@ -14,6 +14,7 @@
 ##############################################################################
 set (umpire_headers
   Allocator.hpp
+  Replay.hpp
   ResourceManager.hpp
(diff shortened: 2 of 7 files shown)


mcfadden8:
@davidbeckingsale, we need to discuss this.  I had to move Replay from the util library into the umpire library level since Replay uses the Allocator object.  But, this doesn't seem like to correct level to me.

For some reason (that I don't understand), the ARM linker was reporting a linker issue that was not being seen on other platforms.  The problem was that the >> override function in Replay.cpp was calling umpire::Allocator.getName() and this was undefined within the umpire_util.a library.  This library was later included by allocation map tests and was reporting that it was undefined since the allocation map tests did not previously have a need to have linkage with the umpire.a library.

So, for this PR, I have moved Replay so that we may discuss further.

mcfadden8:
@davidbeckingsale - ping?

davidbeckingsale:
This is an acceptable solution for now, but we have some weird mixed up dependencies between different components of Umpire (my fault). We should discuss how to address this going forward.

mcfadden8:
Do you know why the build is failing?  I have not seen this particular failure before.

davidbeckingsale:
That was me (I think) - by adding `override` to one of the methods in that class, I actually need it for all of them. I can fix it.

mcfadden8:
@davidbeckingsale, it appears that I am not authorized to merge the pull request.  If you are satisfied with the fix, would you either perform the merge or give me the permission to do so?

Thanks,

davidbeckingsale:
Are you sure?

mcfadden8:
@davidbeckingsale, it worked now.  Thanks

PR #37, opened 2018-12-11
[WIP] Added heuristic function to cause coalesce operation

mcfadden8:
@davidbeckingsale, this is what I am thinking for the coalesce heuristic.  Thoughts??


davidbeckingsale:
I like this. There are two things I would consider adding:
- a few general "utility" heuristics that people could use
- a way to get more detailed stats (over and above getCurrentSize and getActualSize) that the heuristic function could use to make a decision.

davidbeckingsale:
Something you can fix too: https://travis-ci.com/LLNL/Umpire/jobs/164456036#L945

mcfadden8:
Will do along with other cleanup.  I just wanted to send out a “WIP” PR to give you an idea of the direction I was heading.

davidbeckingsale:
Of course! It's looking good 😄

mcfadden8:
I've updated the interface to what I believe it should be for the heuristic.  What type of general "utility heuristic functions" did you have in mind?

davidbeckingsale:
For the "utility" functions, something like ones that return true if the pool has 50%, 75% and 90% used/total_block_size ratio. Does that make sense?

Basically, write some examples that people can use, rather than have multiple users have the same simple code that calculates that the pool has X% usage.

davidbeckingsale:
The signature for the heuristic looks good to me.

src/umpire/strategy/DynamicPool.hpp
```diff
@@ -50,13 +65,15 @@ class DynamicPool :
      * \param min_initial_alloc_size The minimum size of the first allocation
      *                               the pool will make.
      * \param min_alloc_size The minimum size of all future allocations.
+     * \param h_fun Heuristic callback function.
      */
     DynamicPool(
         const std::string& name,
-        int id,
+        int _id,
```
davidbeckingsale: This name change isn't required.

```suggestion
        int id,
```

mcfadden8:
Hi @davidbeckingsale, 

I knew that this would fail during the build process (at least for gcc).  But I haven't been able to figure out (yet) how to fix it.  Do you happen to know what would cause [this error](https://travis-ci.com/LLNL/Umpire/jobs/170355774#L1026)?

Replay does not seem to like the addition of a function pointer to the variadic template.

Marty

mcfadden8:
@trws, do you know why [this compilation warning](https://travis-ci.com/LLNL/Umpire/jobs/170355774#L1026) is happening?  I'm not sure I understand where the NULL comparison is even coming from.

davidbeckingsale:
Can you split the the line to look like:
```
ss << ",";
ss << firstArg << ...
```

@trws figured out that it's trying to left-shift the function pointer!

mcfadden8:
Hmm..  I changed it, but I don't recall seeing anything in the error message indicating a bitwise operator.  The error seemed to be talking about comparison.  I'm not sure I understand the error to be honest.

trws:
It’s a bad error message, clang’s makes it more clear:

test.cpp:6:25: warning: address of function 'foo' will always evaluate to 'true' [-Wpointer-bool-conversion]
  (std::cout << ",") << foo << std::endl;
                     ~~ ^~~
test.cpp:6:25: note: prefix with the address-of operator to silence this warning
  (std::cout << ",") << foo << std::endl;
                        ^
                        &


It’s not that it’s a comparison, it’s that the pointer is being implicitly converted to bool, and that conversion will always evaluate to true since the value is a function literal.

On 15 Jan 2019, at 14:21, Marty McFadden wrote:

Hmm.. I changed it, but I don't recall seeing anything in the error message indicating a bitwise operator. The error seemed to be talking about comparison. I'm not sure I understand the error to be honest.

mcfadden8:
Interestingly (annoyingly) enough, I plan to drop this particular argument as a pointer to a heuristic function cannot be duplicated for allocation replay purposes.  Perhaps I only reference firstArg if its type is not this bool function pointer.

mcfadden8:
Tom, would you attach your test source to this?  I am still having difficulty getting this to work even with the suggested placement of the address-of operator.

mcfadden8:
I will take Tom’s advice and change the interface to use a functor as the callback instead of a function pointer.

tests/integration/strategy_tests.cpp
```diff
@@ -77,6 +77,47 @@ TEST_P(StrategyTest, Allocate) {
...
 }
 
+TEST_P(StrategyTest, ReleaseableSizes) {
+  auto strategy = allocator->getAllocationStrategy();
+
+  auto dynamic_pool = std::dynamic_pointer_cast<umpire::strategy::DynamicPool>(strategy);
+
+  if (!dynamic_pool) {
+    std::cerr << allocator->getName() << " is not a DynamicPool, cannot run this test!\n";
+    return;
```
mcfadden8: @davidbeckingsale , I have thus far not been able to get a non-null value for dynamic_pool.  Can you take a look and help me understand what I might be doing wrong here?

src/umpire/strategy/DynamicPoolHeuristic.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "umpire/strategy/DynamicPoolHeuristic.hpp"
+
+#include "umpire/ResourceManager.hpp"
+
+#include "umpire/util/Macros.hpp"
+
+namespace umpire {
+namespace strategy {
+
+bool heuristicAllAllocationsAreReleaseable( const strategy::DynamicPool& dynamic_pool )
```
davidbeckingsale: These should be `snake_case`

src/umpire/strategy/AllocationStrategy.hpp
```diff
@@ -74,15 +74,30 @@ class AllocationStrategy :
...
+     * against them.  If the size of the set is greater than one, then
+     * the pool will have a number of bytes that may be released back to
+     * the resource or coalesced into a larger block.
+     *
+     * Note: This method will always return 0 for strategies that are not
+     *       memory pools.
+     *
+     * \return The total number of bytes that are releaseable
+     */
+    virtual long getReleaseableSize() const noexcept;
```
davidbeckingsale: I thought this was only going to be added to the DynamicPool, since it's not applicable to multiple strategies?

mcfadden8: You're right.  Fixed.

src/umpire/strategy/DynamicPoolHeuristic.cpp
mcfadden8: I've addressed this.

src/umpire/Replay.hpp
```diff
@@ -18,29 +18,16 @@
...
-}
+#include "umpire/Allocator.hpp"
+#include "umpire/strategy/AllocationStrategy.hpp"
+#include "umpire/strategy/DynamicPool.hpp"
 
 namespace umpire {
 namespace replay {
-std::ostream& operator<< (std::ostream& out, umpire::Allocator& alloc);
+std::ostream& operator<< (std::ostream& out, umpire::Allocator& );
+std::ostream& operator<< (std::ostream& out,
```
davidbeckingsale: Do we still need this, given that you are trying to use `typeid` below to skip printing the heuristic?

src/umpire/strategy/AllocationTracker.hpp
```diff
@@ -37,12 +37,12 @@ class AllocationTracker :
     void* allocate(size_t bytes);
 
     void deallocate(void* ptr);
-
+
```
davidbeckingsale: There's blank spaces on this line.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -23,13 +23,16 @@ namespace strategy {
 
 DynamicPool::DynamicPool(
     const std::string& name,
-    int id,
+    int _id,
```
davidbeckingsale: This doesn't need the leading underscore anymore does it?

src/umpire/strategy/DynamicPool.hpp
```diff
@@ -50,32 +60,50 @@ class DynamicPool :
      * \param min_initial_alloc_size The minimum size of the first allocation
      *                               the pool will make.
      * \param min_alloc_size The minimum size of all future allocations.
+     * \param h_fun Heuristic callback function.
```
davidbeckingsale: h_fun is probably an unnecessarily short argument name, can it be slightly more descriptive?

src/umpire/strategy/DynamicPool.hpp
```diff
@@ -50,32 +60,50 @@ class DynamicPool :
...
+     * against them.  If the size of the set is greater than one, then
+     * the pool will have a number of bytes that may be released back to
+     * the resource or coalesced into a larger block.
+     *
+     * Note: This method will always return 0 for strategies that are not
+     *       memory pools.
+     *
+     * \return The total number of bytes that are releaseable
+     */
+    long getReleaseableSize() const noexcept;
```
davidbeckingsale: Nice!

src/umpire/strategy/DynamicPool.hpp
```diff
@@ -50,32 +60,50 @@ class DynamicPool :
...
 
+    /*!
+     * \brief Get the number of bytes that may be released back to resource
+     *
+     * A memory pool has a set of blocks that have no allocations
+     * against them.  If the size of the set is greater than one, then
+     * the pool will have a number of bytes that may be released back to
+     * the resource or coalesced into a larger block.
+     *
+     * Note: This method will always return 0 for strategies that are not
```
davidbeckingsale: This note is not needed anymore.

src/umpire/Replay.hpp
mcfadden8: Yes, even though I am using the typeid, the code will still exist within the if block and the compiler will complain that the operator does not know how to handle that type.  So, the code is protected by the if block and will never be called, but the code itself still exists.

davidbeckingsale: Gotcha.

src/umpire/strategy/AllocationTracker.hpp
davidbeckingsale:
```suggestion

```

davidbeckingsale: @mcfadden8

tests/integration/replay/replay_tests.bash
```diff
@@ -7,7 +7,7 @@ replayprogram=$1/replay
 # may be replayed.
 #
 UMPIRE_REPLAY="On" $testprogram >& replay_test1.csv
-if [ $? -nq 0 ]; then
+if [ $? -ne 0 ]; then
```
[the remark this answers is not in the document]
mcfadden8: Oh yeah.  This was intentional  I discovered that the script was not previously reporting failure conditions to bamboo and tests were succeeding when they should not have been.

PR #40, opened 2019-01-09
Switch to regular tracked code (not submodule) for Simpool

davidbeckingsale:


mcfadden8:
Great!  Will the history be retained?

jdahm:
Using an orphaned branch could work...

PR #41, opened 2019-01-16
[Deprecated] NUMA memory resources

jdahm:
This PR adds memory resources for NUMA nodes. It is larger than normally adding a resource because it is the first instance in which where the number of memory resources introduced at initialization is a function of the system hardware.

The major changes are:
1. Added a traits parameter to pure virtual `MemoryResource::isValidMemoryResourceFor()`.
2. Added a new container to `ResourceManager` to hold the variable size list of NUMA memory resources. Since each system has a different number of nodes, an enumerated list or string-based keys for unordered_maps doesn't work. The new `m_resource_list` holds all the memory resources, and `r.getAllocatorFor(traits)` simply loops through looking for a "good enough" match.

To do before merging:
- [x] Additional unit tests
- [ ] Add stream tests to verify this works


jdahm:
I see Travis, even though it's running on Linux apparently doesn't have the `numa.h` header 😄 . Should we make ENABLE_NUMA an option at the CMake level? Currently it is auto-detecting the system type and enabling it.

davidbeckingsale:
@jdahm - yup, that makes sense. Our Travis builds actually use Docker containers, which could be updated to add libnuma...

CMakeLists.txt
```diff
@@ -17,12 +17,13 @@ cmake_policy(SET CMP0025 NEW)
...
 cmake_policy(SET CMP0057 NEW)
 
-project(Umpire 
+project(Umpire
   LANGUAGES CXX C
   VERSION 0.3.0)
 
 set(ENABLE_OPENMP On CACHE Bool "")
 set(ENABLE_CUDA On CACHE Bool "")
+set(ENABLE_NUMA Off CACHE Bool "")
```
davidbeckingsale: @jdahm as discussed, we should consider disabling CUDA resources when ENABLE_NUMA is on. This might be best done with another option.

Is there a way to only consider CPU attached NUMA nodes?

src/umpire/CMakeLists.txt
```diff
@@ -86,6 +87,12 @@ else ()
   set(UMPIRE_ENABLE_UM Off)
 endif()
 
+# Disable NUMA if on Mac OS or a non-UNIX platform
+if (ENABLE_NUMA AND (APPLE OR NOT UNIX))
+  message(WARNING "NUMA is not available. Setting UMPIRE_ENABLE_NUMA=Off.")
```
davidbeckingsale: I think this should be a fatal message, and tell the user how to re-run CMake.

src/umpire/ResourceManager.cpp
```diff
@@ -32,6 +32,10 @@
 #include "umpire/resource/RocmPinnedMemoryResourceFactory.hpp"
 #endif
 
+#if defined(UMPIRE_ENABLE_NUMA)
+#include "umpire/resource/NUMAMemoryResourceFactory.hpp"
```
davidbeckingsale: Can you rename the file NumaMemoryResourceFactory. It's a bit nit-picky but I think it looks better.

src/umpire/ResourceManager.cpp
```diff
@@ -203,6 +227,21 @@ ResourceManager::getAllocator(int id)
   return Allocator(m_allocators_by_id[id]);
 }
 
+Allocator
+ResourceManager::getAllocatorFor(const resource::MemoryResourceTraits traits)
+{
+  UMPIRE_LOG(Debug, "(Looking up allocator by traits)");
+
+  for (auto r : m_resource_list) {
```
davidbeckingsale: I'd like to find a way to tie this to the "matcher" function of the resource factories, but for now, can you update it to take into account the default values of the traits?

src/umpire/resource/NUMAMemoryResourceFactory.cpp
```diff
@@ -0,0 +1,66 @@
...
+#endif
+
+namespace umpire {
+namespace resource {
+
+NUMAMemoryResourceFactory::NUMAMemoryResourceFactory(const int numa_node_)
+  : numa_node(numa_node_) {}
+
+std::size_t
+NUMAMemoryResourceFactory::getNodeCount() {
```
davidbeckingsale: Can this be something like getNumberOfNumaNodes(), I think that getNodeCount is too ambiguous.

src/umpire/resource/NUMAMemoryResourceFactory.cpp
```diff
@@ -0,0 +1,66 @@
...
+std::shared_ptr<MemoryResource>
+NUMAMemoryResourceFactory::create(const std::string& UMPIRE_UNUSED_ARG(name), int id)
+{
+  MemoryResourceTraits traits;
+
+  traits.unified = false;
+  traits.size = 0;
+  traits.numa_node = numa_node;
+
+  traits.vendor = MemoryResourceTraits::vendor_type::IBM;
```
davidbeckingsale: Is there a good way to figure out the CPU vendor?

src/umpire/resource/NUMAMemoryResourceFactory.cpp
```diff
@@ -0,0 +1,66 @@
...
+  traits.kind = MemoryResourceTraits::memory_type::DDR;
```
davidbeckingsale: This is not always the case, especially if the NUMA node is on a GPU....

jdahm: I did this because I copied the HostResourceFactory method, which uses IBM. I'll go back and change both of these.

CMakeLists.txt
davidbeckingsale: I still think this is a good idea, we should have 3 possible configurations:
1. NUMA everywhere (CPU & GPU)
2. NUM CPU, CUDA for GPU
3. No NUMA

CMakeLists.txt
```diff
@@ -74,7 +81,7 @@ if (NOT BLT_LOADED)
   endif ()
 
   include(${BLT_SOURCE_DIR}/SetupBLT.cmake)
-endif()
+endif ()
```
davidbeckingsale: I don't think these space changes are necessary, and I think most of the CMakeLists are formatted with spacing before the `()`

src/umpire/resource/DetectVendor.cpp
```diff
@@ -0,0 +1,30 @@
...
+//
+// For details, see https://github.com/LLNL/Umpire
+// Please also see the LICENSE file for MIT license.
+//////////////////////////////////////////////////////////////////////////////
+
+#include "umpire/resource/DetectVendor.hpp"
+
+namespace umpire {
+
+resource::MemoryResourceTraits::vendor_type CpuVendorType() noexcept {
```
davidbeckingsale: Can you give this function a `snake_case` name?

src/umpire/resource/NumaMemoryResourceFactory.cpp
```diff
@@ -0,0 +1,123 @@
...
+extern "C" {
+#include <numa.h>
+}
+
+namespace umpire {
+namespace resource {
+
+namespace numa {
+
+std::size_t nodeCount() {
```
davidbeckingsale: `snake_case` please 😄

src/umpire/resource/NumaMemoryResourceFactory.cpp
```diff
@@ -0,0 +1,123 @@
...
+  if (numa_available() < 0) UMPIRE_ERROR("libnuma is unusable.");
+  return numa_max_possible_node() + 1;
+}
+
+std::vector<std::size_t> getHostNodes() {
```
davidbeckingsale: snakes....

src/umpire/resource/NumaMemoryResourceFactory.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  numa_free_nodemask(node_mask);
+  numa_free_cpumask(node_mask);
+
+  return host_nodes;
+}
+
+std::vector<std::size_t> getDeviceNodes() {
```
davidbeckingsale: 🐍

CMakeLists.txt
jdahm: I was trying to make it consistent. If you look throughout this file there are endifs both with and without the space. Let me know if you don't want that in this PR and I'll take them out.

jdahm: No worries. It might have been better to do in a separate PR, but I thought I'd just put it here for simplicity :)

src/umpire/ResourceManager.cpp
```diff
@@ -95,6 +109,16 @@ ResourceManager::ResourceManager() :
     std::make_shared<resource::RocmPinnedMemoryResourceFactory>());
 #endif
 
+#if defined(UMPIRE_ENABLE_DEVICE) and defined(UMPIRE_ENABLE_NUMA)
```
davidbeckingsale: Does `and` work here rather than `&&`?

jdahm: I think this works, but for uniformity I'll change it to &&

CMakeLists.txt
```diff
@@ -17,12 +17,14 @@ cmake_policy(SET CMP0025 NEW)
...
 cmake_policy(SET CMP0057 NEW)
 
-project(Umpire 
+project(Umpire
   LANGUAGES CXX C
   VERSION 0.3.1)
 
 set(ENABLE_OPENMP On CACHE Bool "")
 set(ENABLE_CUDA On CACHE Bool "")
+option(ENABLE_NUMA_HOST "Add allocators for host NUMA nodes" Off)
```
davidbeckingsale: I think that the options should be:

- `ENABLE_NUMA`: host NUMA
- `ENABLE_NUMA_DEVICES`: GPU numa nodes

Sorry for flip-flopping a bit here.

examples/cookbook/recipe_numa_allocator.cpp
```diff
@@ -0,0 +1,55 @@
...
+int main(int, char**) {
+  auto& rm = umpire::ResourceManager::getInstance();
+
+  /*
+   * Get the first NUMA node on the system and set this as the
+   * location for the process.
+   */
+  auto host_nodes = umpire::numa::get_host_nodes();
+  const int node = host_nodes[0];
+  umpire::numa::run_on_node(node);
```
davidbeckingsale: Do you have to set this to allocate on a specific NUMA node?

src/umpire/ResourceManager.cpp
```diff
@@ -118,52 +145,105 @@ ResourceManager::initialize()
...
+    for (std::size_t n : host_nodes) {
+      resource::MemoryResourceTraits traits;
+      traits.numa_node = n;
+      const std::string name = base_name + std::to_string(n);
+      m_memory_resources[name] = registry.makeMemoryResource(name, getNextId(), traits);
+    }
+  }
+#endif
 
 #if defined(UMPIRE_ENABLE_DEVICE)
```
davidbeckingsale: I don't think we want to create these if ENABLE_NUMA is on. For each physical resource, we should only support one way of accessing it.

src/umpire/ResourceManager.cpp
```diff
@@ -118,52 +145,105 @@ ResourceManager::initialize()
...
-  m_allocators_by_id[device_allocator->getId()] = device_allocator;
+  {
+    const std::string name = resource::type_to_string(resource::Device);
+    auto device_allocator = m_memory_resources[name];
+    m_allocators_by_name[name] = device_allocator;
+    m_allocators_by_id[device_allocator->getId()] = device_allocator;
+  }
 #endif
 
 #if defined(UMPIRE_ENABLE_PINNED)
```
davidbeckingsale: Can pinned memory be created any other way that cudaMallocHost?

CMakeLists.txt
jdahm: No worries, I was the one who made this initial structure. I'll go ahead and make the change.

examples/cookbook/recipe_numa_allocator.cpp
jdahm: No, I thought it would clarify the NUMA allocator, but I can switch it to get the preferred node and allocate there instead of changing the run node.

src/umpire/ResourceManager.cpp
jdahm: If we remove this and keep everything else untouched, then allocators on NUMA nodes, other than the preferred node one replacing the malloc allocator for resource::Host, will not be found. `getAllocatorFor()` loops through `m_memory_resources` here:
```c++
Allocator
ResourceManager::getAllocatorFor(const resource::MemoryResourceTraits traits)
{
  UMPIRE_LOG(Debug, "(Looking up allocator by traits)");

  const resource::MemoryResourceTraits d_traits;
  for (auto r : m_memory_resources) {
```

src/umpire/ResourceManager.cpp
jdahm: Yes, using `cudaHostRegister()`. However, there isn't a pure Linux solution without the CUDA driver involved. A quick investigation of cudaMallocHost shows that it uses mmap with MAP_FIXED to keep the virtual to physical mapping constant. This we can do, but it also maps this into the page table of the GPU, which we don't have direct access to.

src/umpire/ResourceManager.cpp
davidbeckingsale: With the new option naming, my question is: what happens if ENABLE_NUMA_DEVICES is on? We don't want to also make the cudaMalloc device?

src/umpire/ResourceManager.cpp
davidbeckingsale: No problem 😄

davidbeckingsale:
@jdahm - can you also resolve the conflict?

CMakeLists.txt
```diff
@@ -17,12 +17,14 @@ cmake_policy(SET CMP0025 NEW)
...
 
-project(Umpire 
+project(Umpire
   LANGUAGES CXX C
   VERSION 0.3.1)
 
 set(ENABLE_OPENMP On CACHE Bool "")
 set(ENABLE_CUDA On CACHE Bool "")
+set(ENABLE_NUMA Off CACHE Bool "Add allocators for host NUMA nodes")
+set(ENABLE_NUMA_DEVICES Off CACHE Bool "Add allocators for all NUMA nodes (including devices)")
```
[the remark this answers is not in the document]
jdahm: Thanks! Added them there too.

PR #43, opened 2019-01-23
Add cookbook example for growing and shrinking a pool

davidbeckingsale:


examples/cookbook/recipe_one_shrink.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <iostream>
+
+int main(int, char**) {
+  auto& rm = umpire::ResourceManager::getInstance();
+
+  auto allocator = rm.getAllocator("DEVICE");
+
+  /* 
+   * Create a 4 Gb pool and reserve one byte.
```
mcfadden8: You probably already know this, but the actual implementation will allocate more than just one byte for alignment purposes.

davidbeckingsale: Yes, I ought to change it to be explicitly a word?

davidbeckingsale:
@mcfadden8 - this should be ready to roll.

davidbeckingsale:
@mcfadden8 - can I get a re-approval, the build was failing with an "unused variable" warning.

PR #47, opened 2019-01-29
Added helper function for generating DynamicPool heuristic function

mcfadden8:


src/umpire/strategy/DynamicPoolHeuristic.hpp
```diff
@@ -31,7 +31,19 @@ class DynamicPool;
...
+   * \brief Return true if 90% of pool is releaseable
+   *
+   * When 90% has been deallocated back to the pool, this heuristic
+   * will return true.
+   *
+   * \param dynamic_pool The dynamic pool object in question.
+   *
+   * \return True if 90% memory in pool is releaseable.
+   */
+  bool heuristic_90_percent_releaseable( const strategy::DynamicPool& dynamic_pool );
```
mcfadden8: @davidbeckingsale, can you think of a way that I could make this function more generic?  I'd like to parameterize the percentage to use rather than creating a separate callback for each percentage.

davidbeckingsale: You can have it be a function that generates a lambda. e.g. it takes the percentage, and returns a function that takes a DynamicPool. Does that make sense?

davidbeckingsale:
```suggestion
  std::function<bool(const strategy::DynamicPool&> heuristic_percent_releaseable( int percentage );
```

src/umpire/strategy/DynamicPoolHeuristic.cpp
```diff
@@ -22,8 +22,18 @@
 namespace umpire {
 namespace strategy {
 
-bool heuristic_all_allocations_are_releaseable( const strategy::DynamicPool& dynamic_pool )
+bool heuristic_90_percent_releaseable( const strategy::DynamicPool& dynamic_pool )
 {
+  const float percentage = 0.90;
+  const long threshold = percentage * dynamic_pool.getActualSize();
+
+  return (dynamic_pool.getReleaseableSize() > threshold);
```
davidbeckingsale:
```suggestion
  return [=] (const strategy::DynamicPool& pool) {
      const threshold = percentage*pool.getActualSize();
      return (pool.getReleasableSize() > threshold);
   };
```

src/umpire/strategy/DynamicPoolHeuristic.cpp
```diff
@@ -22,8 +22,18 @@
 namespace umpire {
 namespace strategy {
 
-bool heuristic_all_allocations_are_releaseable( const strategy::DynamicPool& dynamic_pool )
+bool heuristic_90_percent_releaseable( const strategy::DynamicPool& dynamic_pool )
 {
+  const float percentage = 0.90;
```
davidbeckingsale:
```suggestion

```

src/umpire/strategy/DynamicPoolHeuristic.cpp
```diff
@@ -22,8 +22,18 @@
...
+  const long threshold = percentage * dynamic_pool.getActualSize();
```
davidbeckingsale:
```suggestion

```

src/umpire/strategy/DynamicPoolHeuristic.cpp
```diff
@@ -22,8 +22,18 @@
 namespace umpire {
 namespace strategy {
 
-bool heuristic_all_allocations_are_releaseable( const strategy::DynamicPool& dynamic_pool )
+bool heuristic_90_percent_releaseable( const strategy::DynamicPool& dynamic_pool )
```
davidbeckingsale:
```suggestion
std::function<bool(const strategy::DynamicPool&)> heuristic_percent_releaseable( int percentage )
```

src/umpire/strategy/DynamicPoolHeuristic.cpp
davidbeckingsale: I'm not 100% sure this is correct, but it should be close.

src/umpire/strategy/DynamicPoolHeuristic.hpp
mcfadden8: Great suggestion!  Done.

mcfadden8:
@davidbeckingsale, I think this is ready.

docs/sphinx/cookbook/dynamic_pool_heuristics.rst
```diff
@@ -0,0 +1,43 @@
+.. _dynamic_pool_heuristics:
+
+=======================================
+Creating a Coalesce Heurisitic Function
```
davidbeckingsale: Maybe change the title to make it specific it's about a DynamicPool:

```suggestion
Improving DynamicPool Performance with a Coalesce Heuristic
```

docs/sphinx/cookbook/dynamic_pool_heuristics.rst
```diff
@@ -0,0 +1,43 @@
...
+=======================================
+
+By default, a DynamicPool object will not coalesce blocks that it allocates
```
davidbeckingsale: Could define what you mean by coalesce

docs/sphinx/cookbook/dynamic_pool_heuristics.rst
```diff
@@ -0,0 +1,43 @@
...
+for the pool.  As needed, the DynamicPool will continue to allocate blocks to
```
davidbeckingsale: This can be a link to the actual DynamicPool stuff:

```suggestion
for the pool.  As needed, the :class:`umpire::strategy::DynamicPool` will continue to allocate blocks to
```

docs/sphinx/cookbook/dynamic_pool_heuristics.rst
```diff
@@ -0,0 +1,43 @@
...
+lead to fragmentation within the pool which could cause the pool to grow too
+large.  For example, a problematic allocation pattern is when an application
+makes several allocations of incrementing size where each allocation is larger
+than the previous block size allocated.
+
+To help address this, applications may offer a heuristic function to the
+DynamicPool object during instantiation that will return true whenever a pool
+reaches a specific threshold of releasable bytes (represented by completely
+free blocks) to the total size of the pool.  The DynamicPool will call this
+heuristic function just before it returns from its deallocate() method and when
```
davidbeckingsale: You can link the functions too:

```suggestion
heuristic function just before it returns from its :func:`umpire::strategy::DynamicPool::deallocate` method and when
```

examples/cookbook/recipe_dynamic_pool_heuristic.cpp
```diff
@@ -0,0 +1,75 @@
+//////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2018, Lawrence Livermore National Security, LLC.
```
davidbeckingsale:
```suggestion
// Copyright (c) 2018-2019, Lawrence Livermore National Security, LLC.
```

src/umpire/strategy/DynamicPool.hpp
```diff
@@ -68,7 +68,7 @@ class DynamicPool :
         Allocator allocator,
         const std::size_t min_initial_alloc_size = (512 * 1024 * 1024),
         const std::size_t min_alloc_size = (1 * 1024 *1024),
-        Coalesce_Heuristic coalesce_heuristic = &heuristic_all_allocations_are_releaseable) noexcept;
+        Coalesce_Heuristic coalesce_heuristic = heuristic_percent_releasable(0)) noexcept;
```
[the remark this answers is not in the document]
mcfadden8: When we discussed this previously, I thought you expressed an interest in making the default be that the heuristic was disabled which is what this effectively does.  I was concerned about making this the default in 0.3.1 because I didn't believe the 0.3.1 heuristic framework offered an adequate interface to provide other options and disabling the heuristic would be different behavior from the previous release.

My preference is to make the default be heuristic_percent_releasable(100) which is what 0.3.0 and 0.3.1 provide.  If you are okay with that, I will switch to that.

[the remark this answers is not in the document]
davidbeckingsale: Yup, I remember 😄 we can make the default have the same behavior as 0.3.0 and 0.3.1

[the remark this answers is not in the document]
davidbeckingsale: Please mark this resolved once you make the switch.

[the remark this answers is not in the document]
mcfadden8: Will do.  Thanks

docs/sphinx/cookbook/dynamic_pool_heuristics.rst
mcfadden8: FYI - working on this now...

mcfadden8: Okay, @davidbeckingsale, please take another look.

PR #54, opened 2019-01-31
Add recipe for moving data to unified memory

davidbeckingsale:


davidbeckingsale:
@Ram3sh - you asked about this a while back, now we have the "cookbook", here is the recipe!

docs/sphinx/cookbook/move_to_managed.rst
```diff
@@ -0,0 +1,18 @@
...
+
+When using a system with NVIDIA GPUs, you may realize that some host data
+should be moved to unified memory in order to make it accessible by the GPU.
+You can do this with the :func:`umpire::ResourceManager::move` operation:
+
+.. literalinclude:: ../../../examples/cookbook/recipe_move_to_managed.cpp
+                    :lines: 34
+
+The move operation will copy the data from host memory to unified memory,
+allocated using the provided ``um_allocator``. The complete example is included
```
mcfadden8: I believe that this operation will also deallocate the memory used from the source of the data move.  I noticed that this is stated in the comment of the cookbook source, but perhaps it also merits being stated here?

davidbeckingsale: Good idea, I have addressed this

davidbeckingsale:
@mcfadden8 - this should be ready

PR #57, opened 2019-02-04
[WIP] Store allocation records in a pool for improved performance

davidbeckingsale:


mcfadden8:
@davidbeckingsale, what speedup is this providing?

davidbeckingsale:
TBD 😄

benchmarks/allocator_burn_test.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <chrono>
+#include <functional>
+#include <iostream>
+#include <random>
+
+#include "umpire/config.hpp"
+
+#include "umpire/Allocator.hpp"
+#include "umpire/ResourceManager.hpp"
+
```
mcfadden8: I'm not sure I understand the significance of this burn test within the context of a benchmark?  Is it being timed somewhere?  Perhaps a comment would be helpful to describe how it is used and what the expected outputs are?

benchmarks/inspector_benchmarks.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  double* data = new double[256];
+
+  void* ptr = static_cast<void*>(data);
+  size_t length = 256*sizeof(double);
+  std::shared_ptr<umpire::strategy::AllocationStrategy> strat(nullptr);
+
+  while (state.KeepRunning()) {
+    inspector.registerAllocation(
+        ptr,
```
mcfadden8: What happens when the same pointer is given to registerAllocation again and again?  Does it cause new records to be created for each registration?

src/umpire/op/HostReallocateOperation.cpp
```diff
@@ -31,9 +33,15 @@ void HostReallocateOperation::transform(
 {
   auto allocator = dst_allocation->m_strategy;
 
-  ResourceManager::getInstance().deregisterAllocation(src_ptr);
-
   *dst_ptr = ::realloc(src_ptr, length);
+  
+  if (dst_ptr == src_ptr) {
```
mcfadden8: Was this a bug that was found while testing this?  Does it relate to using pool storage of an allocation record?

davidbeckingsale: It's not quite finished.

davidbeckingsale:
This is all still a WIP...

src/umpire/strategy/AllocationAdvisor.cpp
```diff
@@ -66,11 +66,10 @@ AllocationAdvisor::AllocationAdvisor(
 void* AllocationAdvisor::allocate(size_t bytes)
 {
   void* ptr = m_allocator->allocate(bytes);
-  auto alloc_record = new util::AllocationRecord{ptr, bytes, this->shared_from_this()};
 
   m_advice_operation->apply(
       ptr,
-      alloc_record,
+      nullptr,
```
mcfadden8: I don't understand what is happening here.  What was this change for?

davidbeckingsale: It used to create a new AllocationRecord that was never used, so now I skip that step.

davidbeckingsale:
Thanks @jdahm - still a bit of work to do on this, I'll ping you guys when it's ready.

benchmarks/inspector_benchmarks.cpp
davidbeckingsale: A new record will be created (from the pool), and it will be inserted at the end of the vector in the AllocationMap

mcfadden8: Ah yes, I believe I remember this now.  It carries with it an implied assumption that deallocations will be done in reverse order from allocations and it is necessary because this is what will allow pools to work where they may have one allocation for the block and a different one for a sub-allocation of the block.

benchmarks/allocator_burn_test.cpp
mcfadden8: @jdahm, can you answer my question?  What is the intention of a "burn test" and how does it relate to a benchmark other than being in the benchmarks directory?

src/umpire/ResourceManager.cpp
```diff
@@ -291,14 +291,14 @@ ResourceManager::hasAllocator(void* ptr)
   return m_allocations.contains(ptr);
 }
 
-void ResourceManager::registerAllocation(void* ptr, util::AllocationRecord* record)
+void ResourceManager::registerAllocation(void* ptr, util::AllocationRecord record)
```
mcfadden8: @jdahm, @davidbeckingsale, - just checking for babies in bathwater here...  What was the original intention behind previously passing around a pointer to an AllocationRecord instead of by value?

src/umpire/Umpire.cpp
```diff
@@ -27,21 +27,23 @@ void print_allocator_records(Allocator allocator, std::ostream& os) {
...
 
   std::vector<const util::AllocationRecord*> recs;
-  std::copy_if(rm.m_allocations.begin(), rm.m_allocations.end(),
-               std::back_inserter(recs), [strategy] (const util::AllocationRecord* rec) {
-                 return rec->m_strategy == strategy;
-               });
+  // std::copy_if(rm.m_allocations.begin(), rm.m_allocations.end(),
+  //              std::back_inserter(recs), [strategy] (const util::AllocationRecord& rec) {
+  //                return rec.m_strategy == strategy;
+  //              });
```
mcfadden8: I don't understand any of the commented out code above and the 'TODO TBD' is nebulous.  Please describe and/or remove.

src/umpire/op/HostReallocateOperation.cpp
```diff
@@ -31,20 +33,24 @@ void HostReallocateOperation::transform(
 {
   auto allocator = dst_allocation->m_strategy;
 
-  delete ResourceManager::getInstance().deregisterAllocation(src_ptr);
+  ResourceManager::getInstance().deregisterAllocation(src_ptr);
 
   *dst_ptr = ::realloc(src_ptr, length);
 
+  if (*dst_ptr == src_ptr) {
+    dst_allocation->m_size = length;
```
mcfadden8: Does this also update the registered allocation of the Inspector with the new length?

src/umpire/strategy/FixedPool.cpp
```diff
@@ -105,7 +105,6 @@ FixedPool::allocInPool(Pool& p) noexcept
     }
   }
 
-  UMPIRE_ASSERT("FixedPool: Logic error in allocate");
```
mcfadden8: Why does this assertion no longer need to exist?

src/umpire/util/AllocationRecordMap.cpp
```diff
@@ -0,0 +1,387 @@
...
+};
+
+static umpire::util::FixedMallocPool block_pool(sizeof(ListBlock<AllocationRecord>));
+
+// Forward declare const iterator
+class RecordListConstIterator;
+
+// TODO Profile this to determine if we should add a small string
+// optimization on top of the block pooling.
+class RecordList
```
mcfadden8: Where is thread safety insured for this object?

src/umpire/util/AllocationRecordMap.cpp
```diff
@@ -0,0 +1,387 @@
...
+  T rec;
+  ListBlock* prev;
+};
+
+static umpire::util::FixedMallocPool block_pool(sizeof(ListBlock<AllocationRecord>));
+
+// Forward declare const iterator
+class RecordListConstIterator;
+
+// TODO Profile this to determine if we should add a small string
```
mcfadden8: Should this TODO be a jira ticket?

src/umpire/util/AllocationRecordMap.cpp
mcfadden8: After reading further down into the public section, it appears that mutual exclusion is being set up there.  Let me know if this is not right, otherwise, you should be able to resolve this comment.

src/umpire/util/FixedMallocPool.cpp
```diff
@@ -0,0 +1,159 @@
...
+        return static_cast<void*>(p.data + m_obj_bytes * index);
+      }
+    }
+  }
+
+  return nullptr;
+}
+
+void*
+FixedMallocPool::allocate(size_t bytes)
```
mcfadden8: So was the heap previously used for these types of allocations?

src/umpire/ResourceManager.cpp
davidbeckingsale: The need for speed....

benchmarks/allocator_burn_test.cpp
jdahm: This benchmark will be removed before merging.

src/umpire/Umpire.cpp
jdahm: This method returns a vector of allocation records created by a specific `Allocator`. I created it a while back because it was useful for inspecting mixed pools. When I changed the `ResourceManager` API, I quickly commented out this method, but yes, it will be updated before merging.

src/umpire/strategy/FixedPool.cpp
jdahm: Good catch. The UMPIRE_ASSERT placed there was malformed/dumb, it always failed because there isn't a condition. I tried switching it to an UMPIRE_ERROR, but the compiler spits out an error telling me the code will always fail at that point. It's not taking into account that it should really never execute that line. I was going to try and fix the compiler error later before merging.

src/umpire/util/AllocationRecordMap.cpp
jdahm: Yes, it could be. I dislike having to be VPNed to access the JIRA, though. Let's see how the performance is after merging. My thinking is that having a small array of RecordLists will be no faster (or slower) than what is now implemented here with the PR, since we have to add an additional integer and keep track of how many elements of the RecordListArray are set/used.

src/umpire/util/AllocationRecordMap.cpp
jdahm: Indeed. This is only used inside `AllocationRecordMap`, which uses a lock.

src/umpire/util/FixedMallocPool.cpp
jdahm: The heap is used in all Umpire pools. `FixedMallocPool` is a pool in `util` that hard-codes `std::malloc() / std::free()`, which (1) avoids the overhead of introspection and (2) avoids potential infinite recursion if you were to track allocations in a pool used in the tracking process!

benchmarks/allocator_burn_test.cpp
mcfadden8: Thanks, I will leave this comment unresolved so that we are reminded about it.

src/umpire/ResourceManager.cpp
mcfadden8: Thanks @davidbeckingsale, but I don't understand.  Are you saying that we needed to pass the pointers around for speed which would imply that copying values around might be slower?

src/umpire/strategy/FixedPool.cpp
mcfadden8: It sounds like you are asserting that this line should never be executed, which, in turn, sounds like a greate place for an UMPIRE_ASSERT

src/umpire/util/AllocationRecordMap.cpp
mcfadden8: Are we merging before running performance tests?  My vote is that we run performance tests (and possibly make updates) before we merge.

src/umpire/ResourceManager.cpp
jdahm: I think the intent was just that. Copying a pointer is 8 bytes, but AllocationRecord has 3 8 byte members. @davidbeckingsale and I were deciding between speed and complexity of the memory management. Storing and passing around pointers means that the method creating the allocation records outside the map "owned" the allocation/deallocation of the structs. With increased optimizations elsewhere, I figured we could back off and switch to pass by value, simplifying the ownership model. Also, there may be certain optimizations performed for passing by value.

src/umpire/strategy/FixedPool.cpp
jdahm: `assert()` has a condition, which this doesn't. It should be an error-type macro.

src/umpire/util/AllocationRecordMap.cpp
jdahm: In this case we can probably quickly assess that before merging, yes. In general my vote is to keep PRs small and keep merging. Merge this, have another PR that expands the benchmark suite, check if there's room for improvement, and iterate.

jdahm:
I'm moving this PR to a new draft.

PR #58, opened 2019-02-05
Add a MixedPool that combines Fixed and Dynamic pools in an attempt to improve performance

davidbeckingsale:
TODO
- [x] Switch to use non-templated fixed pool
- [x] Add strategy test
- [x] Add a cookbook recipe


mcfadden8:
When I tried to build this branch on rzansel, I get a compilation error:
``` script

CMake Error at blt/cmake/BLTMacros.cmake:609 (add_executable):
  Cannot find source file:

    mixed_pool.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx
Call Stack (most recent call first):
  examples/CMakeLists.txt:55 (blt_add_executable)
```

src/umpire/strategy/MixedPool.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "umpire/Allocator.hpp"
+
+namespace umpire {
+namespace strategy {
+
+class MixedPool :
+  public AllocationStrategy
+{
+  public:
+    MixedPool(
```
mcfadden8: @davidbeckingsale, shouldn't this take the same default arguments as DynamicPool so that it knows how to construct the underlying DynamicPool object for larger allocations? ([min_initial_alloc_size, min_alloc_size, and heuristic_fun](https://github.com/LLNL/Umpire/blob/feature/UM-281-faster-pool/src/umpire/strategy/DynamicPool.hpp#L69))

davidbeckingsale: Good suggestion. I will add more comprehensive Doxygen to the class too.

jdahm:
I had been testing out an idea of storing "user data" in the allocation record, in order to avoid having another map stored in the mixed pool and an additional map lookup for every deallocate call. However, it seems that `AllocationRecords` are created at the `MemoryResource` level, not at `AllocationStrategy`, so allowing a strategy to pass "user data" down resource to resource requires extensive changes.

It appears that `AllocationRecords` are being used inside `ResourceManager`, among other uses, as a way to get the size of an allocation, and find the correct `Allocator` from which to deallocate. Could it make sense, in the long run, to move this tracking to the `Allocator` level (have `Allocator` inherit from `mixins::Inspector`)?

jdahm:
I didn't actually edit the CHANGELOG for this feature, but the test is passing. Let's use something like `git diff develop...feature`

davidbeckingsale:
Looks like the CHANGELOG check is failing to me. It does what you suggest: https://github.com/LLNL/Umpire/blob/develop/.github/actions/bin/diff-check/entrypoint.sh#L5

jdahm:
Strange - it was passing last night on `57b4d10`.

src/umpire/Umpire.hpp
```diff
@@ -72,13 +74,25 @@ int get_patch_version()
...
+/*!
+ * \brief Filters and returns a vector of AllocationRecords.
+ *
+ * Note that this returns a vector of allocations because unlike
+ * general allocation tracking, no more than one address can be
+ * associated with an allocator.
+ *
+ * \param allocator source Allocator.
+ */
+std::vector<util::AllocationRecord> get_allocator_records(Allocator allocator);
```
davidbeckingsale: This should be a vector of const pointers I think.

benchmarks/allocator_stress_test.cpp
```diff
@@ -56,6 +59,18 @@ void benchmark_allocator(std::string name) {
 int main(int, char**) {
   benchmark_allocator("HOST");
 
+  auto& rm = umpire::ResourceManager::getInstance();
```
mcfadden8: Reminder, please also update replay to be able to replay the new makeAllocator<umpire::strategy::MixedPool> operation.

src/umpire/strategy/MixedPool.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+    long getCurrentSize() const noexcept override;
+    long getActualSize() const noexcept override;
+    long getHighWatermark() const noexcept override;
+
+    Platform getPlatform() noexcept override;
+
+  private:
+    enum { NUM_FIXED_POOLS = LastFixed - FirstFixed + 1 };
+    using StrategyPtr = std::shared_ptr<umpire::strategy::AllocationStrategy>;
```
mcfadden8: I thought we weren't using std::shared_ptr anymore?

src/umpire/strategy/MixedPool.inl
```diff
@@ -0,0 +1,157 @@
...
+
+template<int FirstFixed, int Current, int LastFixed, int Increment>
+struct make_fixed_pool_array {
+  template<typename Array>
+  static void eval(Array& fixed_pool, Allocator& allocator) {
+    const int index = Current - FirstFixed;
+    const std::size_t size = 1 << Current;
+    std::stringstream ss{"internal_fixed_"};
+    ss << size;
+    fixed_pool[index] = std::make_shared<FixedPool<unsigned char[size]> >(ss.str(), -1, allocator);
```
mcfadden8: Same comment about shared pointers.  I'm okay with using them, but I thought that there was a recent effort to remove them?

src/umpire/strategy/MixedPool.inl
```diff
@@ -0,0 +1,157 @@
...
+    const std::size_t size = 1 << LastFixed;
+    std::stringstream ss{"internal_fixed_"};
+    ss << size;
+    fixed_pool[index] = std::make_shared<FixedPool<unsigned char[size]> >(ss.str(), -1, allocator);
+  }
+};
+
+} // anonymous namespace
+
+inline static size_t nextPower2(unsigned int n) {
```
mcfadden8: I suggest that either a comment is added or the name is updated to reflect that this will only work for a 32 bit number

jdahm: Good idea!

src/umpire/strategy/MixedPool.inl
jdahm: Removed!

src/umpire/strategy/MixedPool.hpp
jdahm: 👍

davidbeckingsale:
This is looking good. The comments are helpful too!

davidbeckingsale:
@jdahm - this branch has some conflicts, but I think it's good to go. Also, note that I can't approve it since I opened the PR! 😄

tools/replay.cpp
```diff
@@ -553,21 +556,18 @@ class Replay {
...
+      }
+      else if (m_row[3] == "umpire::strategy::MixedPool" ) {
         const std::string& allocName = m_row[6];
-        std::size_t PoolSize = hmm...
-
-        if ( introspection )  m_rm.makeAllocator<umpire::strategy::FixedPool<PoolSize>, true>(name, m_rm.getAllocator(allocName));
-        else                  m_rm.makeAllocator<umpire::strategy::FixedPool<PoolSize>, false>(name, m_rm.getAllocator(allocName));
-#endif
+        if ( introspection ) m_rm.makeAllocator<umpire::strategy::MixedPool, true>(name, m_rm.getAllocator(allocName));
+        else                 m_rm.makeAllocator<umpire::strategy::MixedPool, false>(name, m_rm.getAllocator(allocName));
```
mcfadden8: Thanks for adding this handler in the replay tool.  Now, we need to please update replay_test to generate one of these pools and then update the output file that the CI uses so that it becomes part of our regular checks.

jdahm: iirc the credit for adding the handler to the replay tool should go to @davidbeckingsale!

jdahm:
This is waiting for a non-templated `FixedPool` to be available.

mcfadden8:
> This is waiting for a non-templated `FixedPool` to be available.

@jdahm and @davidbeckingsale,

Why?  If possible, I would rather see us decouple the two projects since the non-templated `FixedPool` will require significant testing and delay the availability of `MixedPool`.

Can we just change the `MixedPool` implementation to internally use a hard-coded set of `FixedPool` sizes and increments?

Then we can update `MixedPool` and `FixedPool` as a subsequent effort.

src/umpire/strategy/MixedPool.hpp
```diff
@@ -0,0 +1,80 @@
...
+namespace umpire {
+namespace strategy {
+
+/**
+ * \brief A faster pool that pulls from a series of pools
+ *
+ * Pool implementation using a series of FixedPools for small sizes,
+ * and a DynamicPool for sizes larger than (1 << LastFixed) bytes.
+ */
+template<int FirstFixed = 8, int Increment = 1, int LastFixed = 22>
```
mcfadden8: @jdahm, I would like to see a recipe for usage of the MixedPoolImpl that uses both the default values for FirstFixed, Increment, and LastFixed and values set by the caller.

Having a comment in the recipe above stating what setting those values will do will be helpful and will illustrate the flexibility of this pool.

jdahm: Even when we change to non-templated fixed pools we'll probably stick to powers of 2 for the sizes. This class won't be templated then, but will probably have a constructor that takes a `vector<int>` of those powers. I'll write one for this version, and we can change it later.

src/umpire/util/allocation_statistics.cpp
```diff
@@ -0,0 +1,40 @@
+//////////////////////////////////////////////////////////////////////////////
```
davidbeckingsale: @jdahm - please create a JIRA ticket about finding a better home for this method.

src/umpire/strategy/MixedPool.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * \brief A faster pool that pulls from a series of pools
+ *
+ * Pool implementation using a series of FixedPools for small sizes,
+ * and a DynamicPool for sizes larger than (1 << LastFixed) bytes.
+ */
+class MixedPool :
+  public AllocationStrategy
+{
+  public:
+    MixedPool(const std::string& name, int id,
```
davidbeckingsale: I think we need a way to also specify options for the internal DynamicPool.

jdahm: Yes, that's a good idea. New commit adds those.

davidbeckingsale:
@jdahm - can you merge in `develop`, then let's get this merged!

benchmarks/allocator_stress_test.cpp
mcfadden8: I will make this work after this has been merged into develop.

tools/replay.cpp
mcfadden8: I'll deal with this after merge.

PR #61, opened 2019-02-12
Feature/probinso/misc fixes

robinson96:
These are some miscellaneous build tweaks and privacy tweaks to help CHAI out.


src/umpire/ResourceManager.hpp
```diff
@@ -132,6 +132,9 @@ class ResourceManager {
     bool hasAllocator(void* ptr);
 
     void registerAllocation(void* ptr, util::AllocationRecord* record);
+    util::AllocationRecord* findAllocationRecord(void *ptr) {
```
davidbeckingsale: Can we make this const? Don't want people messing with records.

src/umpire/tpl/judy/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ blt_add_library(
   HEADERS ${judy_headers}
   HEADERS_OUTPUT_SUBDIR umpire/tpl/judy)
 
-if (CMAKE_CXX_COMPILER_ID MATCHES Clang)
+if (CMAKE_CXX_COMPILER_ID MATCHES Clang AND NOT CMAKE_BUILD_TYPE MATCHES Debug)
```
davidbeckingsale: Nice! 👍

src/umpire/util/AllocationMap.cpp
```diff
@@ -116,6 +116,9 @@ AllocationMap::findRecord(void* ptr) const
             << " with size " << alloc_record->m_size);
 
       }
+      else {
+         alloc_record = nullptr;
```
davidbeckingsale: 💯

mcfadden8:
Just a comment to let you know that I merged in most of these changes and have been running with them in my branch to debug these judy problems.  I had to uncomment some lines dealing with CONST memory (not sure why they were commented out, but it seems these commented lines were not part of this PR)

src/umpire/ResourceManager.hpp
robinson96: CHAI is using the non-constness to update an allocation record when a user declares a raw pointer capture to be a certain length.

davidbeckingsale: I'm confused why it would be re-registered - the only reason CHAI would have to update the size is if you want to use `ResourceManager::copy` without a size argument. However, CHAI knows the size it needs to copy, so I would prefer to do it that way.

src/umpire/ResourceManager.cpp
```diff
@@ -130,7 +130,7 @@ ResourceManager::initialize()
 #endif
 
 #if defined(UMPIRE_ENABLE_CUDA)
-  m_memory_resources[resource::Constant] = registry.makeMemoryResource("DEVICE_CONST", getNextId());
+  //m_memory_resources[resource::Constant] = registry.makeMemoryResource("DEVICE_CONST", getNextId());
```
davidbeckingsale: Please maintain this change locally. As you know, we haven't been able to reproduce the issue.

davidbeckingsale:
I think nearly all these changes are either superseded by others, or are unnecessary, closing.

PR #62, opened 2019-02-13
Numa interface

jdahm:
Add support for setting a NUMA policy via AllocationStrategy where the memory is provided by another allocator, by default the PosixMemalignAllocator. The restriction is that the memory provided is page-aligned for `mbind()` to succeed.

To do:
- [x] Detect vendor
- [x] Document option
- [x] Integration tests
- [x] Pull over cookbook example from deprecated branch
- [x] Add a CPU<->GPU memory movement example


src/umpire/alloc/PosixMemalignAllocator.hpp
```diff
@@ -0,0 +1,72 @@
...
+   * \brief Allocate bytes of memory using posix_memalign.
+   *
+   * \param bytes Number of bytes to allocate. Does not have to be a multiple of the system page size.
+   * \return Pointer to start of the allocation.
+   *
+   * \throws umpire::util::Exception if memory cannot be allocated.
+   */
+  void* allocate(size_t bytes)
+  {
+    const long pagesize = sysconf(_SC_PAGESIZE);
```
davidbeckingsale: Can we cache this? I don't know if this call is expensive

src/umpire/resource/HostResourceFactory.cpp
```diff
@@ -46,11 +56,11 @@ HostResourceFactory::create(const std::string& UMPIRE_UNUSED_ARG(name), int id)
   traits.unified = false;
   traits.size = 0;
 
-  traits.vendor = MemoryResourceTraits::vendor_type::IBM;
-  traits.kind = MemoryResourceTraits::memory_type::GDDR;
+  traits.vendor = MemoryResourceTraits::vendor_type::UNKNOWN;
```
davidbeckingsale: It would be good to use the vendor detection from #41

src/umpire/strategy/NumaPolicyStrategy.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+NumaPolicyStrategy::NumaPolicyStrategy(
+    const std::string& name,
+    int id,
+    int numa_node,
+    Allocator allocator) :
+  AllocationStrategy(name, id),
+  m_node(numa_node),
+  m_mask(numa_bitmask_alloc(numa_max_node() + 1)),
+  m_allocator(allocator.getAllocationStrategy())
```
davidbeckingsale: We might want to check that the `Platform` of the passed allocator is `Platform::cpu`, since this won't work for others.

src/umpire/strategy/NumaPolicyStrategy.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "umpire/Allocator.hpp"
+
+// Forward declaration of bitmask for NUMA
+struct bitmask;
+
+namespace umpire {
+
+namespace strategy {
+
+class NumaPolicyStrategy :
```
davidbeckingsale: I think it's better to drop the `Strategy` from the name, less verbose.

src/umpire/resource/HostResourceFactory.cpp
jdahm: Agreed. Moving that over!

src/umpire/alloc/PosixMemalignAllocator.hpp
jdahm: Sure. I didn't do that initially because there's no clear location for it, since this is in a header file. I'll add some NUMA wrapper methods in `util/Numa.{c,h}pp` and drop it in there 👍

src/umpire/strategy/NumaPolicyStrategy.hpp
jdahm: 👍

src/umpire/strategy/NumaPolicyStrategy.cpp
jdahm: good catch! I added a check.

src/umpire/alloc/PosixMemalignAllocator.hpp
davidbeckingsale: You could just make it a `static`, but either way is fine.

docs/sphinx/advanced_configuration.rst
```diff
@@ -35,6 +36,11 @@ These arguments are explained in more detail below:
   This option enables support for NVIDIA GPUs. If Umpire is built without CUDA
   or ROCM support, then only the ``HOST`` allocator is available for use.
 
+* ``ENABLE_NUMA``
+  This option enables support for NUMA. The `strategy::NumaPolicy` is available
```
davidbeckingsale: if you add Doxygen, you will be able to use `` :class:`umpire::strategy::NumaPolicy` `` to link to it from here!

src/umpire/ResourceManager.cpp
```diff
@@ -399,6 +405,40 @@ ResourceManager::move(void* ptr, Allocator allocator)
...
+      auto src_alloc_record = m_allocations.find(ptr);
+
+      const size_t size = src_alloc_record->m_size;
+      util::AllocationRecord dst_alloc_record;
+      dst_alloc_record.m_size = src_alloc_record->m_size;
+      dst_alloc_record.m_strategy = numa_alloc;
+
+      void *ret = nullptr;
+      if (size > 0) {
+        auto op = op_registry.find("RELOCATE",
```
davidbeckingsale: Can you register this as "MOVE" - I kept reading this as "REALLOCATE" and was thinking, "Why is it called reallocate?!

src/umpire/ResourceManager.cpp
```diff
@@ -399,6 +405,40 @@ ResourceManager::move(void* ptr, Allocator allocator)
     return ptr;
   }
 
+#if defined(UMPIRE_ENABLE_NUMA)
+  {
+    // short circuit if allocator is a NUMA allocator
+    auto numa_alloc = std::dynamic_pointer_cast<strategy::NumaPolicy>(allocator.getAllocationStrategy());
```
davidbeckingsale: The `Allocator` might be wrapped in an `AllocationTracker`, so you will have to check twice

src/umpire/op/NumaRelocateOperation.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "umpire/util/Macros.hpp"
+#include "umpire/util/Numa.hpp"
+
+#include "umpire/strategy/NumaPolicy.hpp"
+
+namespace umpire {
+namespace op {
+
+void NumaRelocateOperation::transform(
```
davidbeckingsale: Rename `NumaMoveOperation`

src/umpire/resource/DetectVendor.cpp
```diff
@@ -0,0 +1,32 @@
...
+// For details, see https://github.com/LLNL/Umpire
+// Please also see the LICENSE file for MIT license.
+//////////////////////////////////////////////////////////////////////////////
+
+#include "umpire/resource/DetectVendor.hpp"
+
+namespace umpire {
+namespace resource {
+
+resource::MemoryResourceTraits::vendor_type cpu_vendor_type() noexcept {
```
davidbeckingsale: 👍

src/umpire/resource/DetectVendor.hpp
```diff
@@ -0,0 +1,28 @@
+//////////////////////////////////////////////////////////////////////////////
```
davidbeckingsale: Can you make this file snake case? `detect_vendor.hpp` - I've been trying to follow that if the file doesn't contain a class.

src/umpire/util/Numa.cpp
```diff
@@ -0,0 +1,52 @@
+//////////////////////////////////////////////////////////////////////////////
```
davidbeckingsale: Rename `numa.cpp`

docs/sphinx/advanced_configuration.rst
jdahm: Very cool! 👍

docs/sphinx/advanced_configuration.rst
```diff
@@ -35,6 +36,11 @@ These arguments are explained in more detail below:
   This option enables support for NVIDIA GPUs. If Umpire is built without CUDA
   or ROCM support, then only the ``HOST`` allocator is available for use.
 
+* ``ENABLE_NUMA``
+  This option enables support for NUMA. The
+  :class:`umpire::strategy::NumaPolicy` is available when built with this
+  option, which may be used to locate the allocation to a specific node.
+
 * ``ENABLE_ROCM``
```
mcfadden8: 👍

jdahm:
Definitely. My plan is to add an example of basic NUMA usage, and a cookbook example that shows how to move memory between nodes.

jdahm:
Good news: this PR is done!
Bad news: So far, this fails:
```
    // Can use CUDA API with this pointer
    if (cudaMemset(dst_ptr, 0, alloc_size) != cudaSuccess) {
      UMPIRE_ERROR("Cannot use address with CUDA");
    }
```
I'm talking with the Linux folks to figure this out.

jdahm:
The CUDA memset call fails in the cookbook example. We're waiting to see if it will get fixed. In the meantime, I suppose we can comment out the CUDA memset bit and merge. Is that acceptable?

davidbeckingsale:
I would be okay with that, but I'm interested to know what @mcfadden8 thinks...

mcfadden8:
Overall, this is looking good.  Instead of disabling the memset feature, can we instead make a slower implementation?  Perhaps an implementation that does a series of mem-copies from a local buffer initialized with values to cuda memory?

I'm interested to find out what is going on here.

jdahm:
My most recent commit returns the correct `umpire::Platform` for the NUMA node.

jdahm:
The NUMA cookbook now uses the host memset operation instead of the device version, so it executes without throwing an error. It also executes both the host and device versions of the move operation.

PR #63, opened 2019-02-13
Add findAllocationRecord(ptr)

jdahm:
This feature is needed for integration with MFEM.


davidbeckingsale:
I think this could be a useful PR, but it might be better to expose a const version of the AllocationMap, so that users can perform there own introspection?

@robinson96 has something similar in #61

jdahm:
Should be ready to merge.

davidbeckingsale:
Looking good, thanks @jdahm

PR #64, opened 2019-02-14
Bugfix/um 238 judy allocation not found

mcfadden8:
The fix for this is in judyL2Array.h.  The other changes were a merge changes made by @robinson96


src/umpire/util/AllocationMap.cpp
```diff
@@ -116,6 +116,9 @@ AllocationMap::findRecord(void* ptr) const
             << " with size " << alloc_record->m_size);
 
       }
+      else {
+         alloc_record = nullptr;
```
davidbeckingsale: Should keep this.

mcfadden8: 👍

mcfadden8:
> Can you rebase this branch and remove the miscellaneous changes that are also included in #61

This should be ready.  I also checked in the test tool for replaying allocation map operations from the replay.csv file since I use that program to confirm the fix.  I'll add documentation for that program under a separate PR.

robinson96:
hey - I think this will cause issues with my PR, and now it's merged, so nvm I guess. This will pull many of my commits into develop and then the last commit undoes them, which will cause fun issues when I merge this in.

PR #66, opened 2019-02-14
Deregister allocation record before resource deallocation

mcfadden8:


mcfadden8:
This one needs a little discussion.  The replay logs I was given indicated a duplicate deallocation was being performed.  I have confirmed with this new test that Umpire is throwing an exception when this happens.  So I am not sure why the application that generated the replay log did not see one unless it was being caught/ignored

davidbeckingsale:
I'm guessing one was being caught...

src/umpire/resource/DefaultMemoryResource.inl
```diff
@@ -55,8 +55,8 @@ void DefaultMemoryResource<_allocator>::deallocate(void* ptr)
 
   UMPIRE_RECORD_STATISTIC(getName(), "ptr", reinterpret_cast<uintptr_t>(ptr), "size", 0x0, "event", "deallocate");
 
-  m_allocator.deallocate(ptr);
   deregisterAllocation(ptr);
+  m_allocator.deallocate(ptr);
```
mcfadden8: I think that reordering this is the correct thing to do here.  Without the reordering, the device will be given the reallocation before we even look to see it is something we have registered.

PR #67, opened 2019-02-14
Add registerExternalAllocation and make registerAllocation private

davidbeckingsale:


src/umpire/ResourceManager.cpp
```diff
@@ -278,6 +278,27 @@ util::AllocationRecord* ResourceManager::deregisterAllocation(void* ptr)
...
 }
 
+void ResourceManager::registerExternalAllocation(
+    void* ptr,
+    size_t size,
+    Platform platform)
+{
+  UMPIRE_LOG(Debug, "(ptr=" << ptr << ", size=" << size << ") with " << this );
+
+  auto record = m_allocations.find(ptr);
```
mcfadden8: This check only checks for the base pointer.  The entire range should be checked.

src/umpire/ResourceManager.hpp
```diff
@@ -239,6 +248,10 @@ class ResourceManager {
     std::shared_ptr<strategy::AllocationStrategy>& findAllocatorForId(int id);
     std::shared_ptr<strategy::AllocationStrategy>& getAllocationStrategy(const std::string& name);
 
+
+    void registerAllocation(void* ptr, util::AllocationRecord* record);
```
mcfadden8: I believe that the implementation of ```registerAllocation()``` needs to confirm that its allocations are also not overlapping with external ones.

src/umpire/ResourceManager.cpp
davidbeckingsale: Agree!

src/umpire/ResourceManager.hpp
davidbeckingsale: I think that is a hopefully unlikely case, unless a user registers a pointer erroneously. I'd prefer to leave it unchecked, since it would add overhead to the registerAllocation call.

src/umpire/ResourceManager.cpp
davidbeckingsale: Actually, this will find any that falls inside the range `[record->m_ptr, record->m_ptr + record->m_size]`. Does that address your concern?

davidbeckingsale:
@mcfadden8 - can you please re-review this?

src/umpire/ResourceManager.cpp
mcfadden8: I'm not sure about that.  For example, I think it will completely miss an allocation that is in the range ```[record->m_ptr+N, ...]``` where ```N<=record->m_size```.  If I understand correctly, this would yield an overlapping set of records.

src/umpire/ResourceManager.hpp
mcfadden8: Could we do something similar to what the DEBUG version of ```malloc()``` does?  Could we add the checks there, incur the overhead, and find the bug?

src/umpire/ResourceManager.cpp
davidbeckingsale: It would find it, since when `N <= record->m_size`, it would count as being an offset of the base pointer.

src/umpire/ResourceManager.hpp
davidbeckingsale: Possibly. Do we want to establish a precedent for that? It would be hard to find overlapping externals without searching the whole AllocationMap.

mcfadden8: I think it could be done with one additional search.  If you looked for the ```ptr+size``` and examined the previous record to see if it found something in the range of ```[ptr, ptr+size]```, I think this would do it right?

davidbeckingsale: No, because there's nothing stopping a user registering an allocations in this situation (where the entries in this list are ordered by address, and each entry is the size):

Before: [ 2, 4, 2 ]
User registers addr with size 11: [_11_, 2, 4, 2]

The user could register an external allocation with any size.

davidbeckingsale:
@mcfadden8 - closing this for now, it will need to be rewritten if we decide to add this support.

PR #68, opened 2019-02-15
Consistent output directories

jdahm:
This PR puts the tests and benchmarks in a directory structure that mimics the source.


CMakeLists.txt
```diff
@@ -93,11 +93,11 @@ install(EXPORT umpire-targets DESTINATION share/umpire/cmake)
 add_subdirectory(src)
 
 if (ENABLE_TESTS)
-  set (UMPIRE_TEST_OUTPUT_DIR ${CMAKE_BINARY_DIR}/test)
+  set (UMPIRE_TEST_OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY})
```
davidbeckingsale: Is this a CMake variable?

jdahm: It seems to be something defined by BLT.
```
~/Umpire/blt $ git grep TEST_OUTPUT_DIRECTORY
SetupBLT.cmake:    set(TEST_OUTPUT_DIRECTORY
```

davidbeckingsale: Perfect!

davidbeckingsale:
@jdahm - let's wait for @mcfadden8 to review this prior to merging.

jdahm:
These are put under the root directory, but they could be put under `bin/`, so the output directories would be `bin/{examples,tools,...}`. What do you think?

davidbeckingsale:
I think nesting the directories like that is a good idea.

CMakeLists.txt
```diff
@@ -93,19 +93,21 @@ install(EXPORT umpire-targets DESTINATION share/umpire/cmake)
 add_subdirectory(src)
 
 if (ENABLE_TESTS)
-  set (UMPIRE_TEST_OUTPUT_DIR ${CMAKE_BINARY_DIR}/test)
+  set(UMPIRE_TESTS_OUTPUT_DIR ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/tests)
```
davidbeckingsale: I'm assuming `CMAKE_RUNTIME_OUTPUT_DIRECTORY` will be `bin`?

benchmarks/CMakeLists.txt
```diff
@@ -12,29 +12,31 @@
 # For details, see https://github.com/LLNL/Umpire
 # Please also see the LICENSE file for MIT license.
 ##############################################################################
+set(benchmarks_dir ${UMPIRE_BENCHMARKS_OUTPUT_DIR})
```
davidbeckingsale: I don't think renaming the variable here helps much? Why not use it directly.

CMakeLists.txt
jdahm: Yeah, it's another variable defined by BLT

benchmarks/CMakeLists.txt
jdahm: 👍

davidbeckingsale:
@mcfadden8 - can you take a look at this?

tests/integration/replay/.gitignore
```diff
@@ -0,0 +1 @@
+replay_test?.csv
```
mcfadden8: Missing a newline at end of file

jdahm: Good catch. Thanks!

davidbeckingsale:
Merge at will :)

PR #69, opened 2019-02-22
Build one library!

davidbeckingsale:


mcfadden8:
This sounds like a great idea @davidbeckingsale!  The builds are currently not passing so I will wait until after they are fixed before reviewing.

davidbeckingsale:
@mcfadden8 - builds are passing 😁

src/umpire/op/CMakeLists.txt
```diff
@@ -28,8 +28,7 @@ set (umpire_op_sources
   MemoryOperation.cpp
   MemoryOperationRegistry.cpp)
 
-set (umpire_op_depends
-  umpire_util)
+set (umpire_op_depends)
```
mcfadden8: So is this just setting umpire_op_depends to nothing?  Should this and the reference line below be removed?

davidbeckingsale: No, because if CUDA or ROCm are used, they are added to the dependency list.

PR #70, opened 2019-02-22
Switch from ENABLE_X_API to just ENABLE_X for C/FORTRAN

davidbeckingsale:


CMakeLists.txt
```diff
@@ -48,9 +46,8 @@ endif ()
 
 message(STATUS "Using CMake version ${CMAKE_VERSION}")
 
-if (ENABLE_FORTRAN_API)
-  set(ENABLE_FORTRAN On)
-  set(ENABLE_C_API On)
+if (ENABLE_FORTRAN)
+  set(ENABLE_C On)
```
mcfadden8: Where does ```ENABLE_FORTRAN``` and ```ENABLE_C``` get set and what is its default?  In other words, I don't see a line similar to the ```set(ENABLE_DOCS Off...)``` line above for ```ENABLE_FORTRAN``` and ```ENABLE_C```

davidbeckingsale: They are BLT options.

mcfadden8: Oh nice!

PR #71, opened 2019-02-22
Add native allocator interface to Fortran API

davidbeckingsale:
This PR will add extra functions to the Allocator wrapper in Fortran, allowing arrays to be allocated in a more native way (based on type and dimensionality).

Documentation can be found here: https://umpire.readthedocs.io/en/feature-um-262-native-fortran/index.html


tests/integration/interface/umpire_fortran_interface.f
```diff
@@ -29,6 +29,17 @@ subroutine allocate_sixty_four
...
 
+  subroutine allocate_through_generic
+    integer(C_INT), pointer :: array(:)
+    type(UmpireAllocator) allocator
+    type(UmpireResourceManager) rm
+
+    rm = rm%get_instance()
+    allocator = rm%get_allocator_by_id(0)
+    call allocator%allocate_array( [ 10 ] , array)
+
```
aaroncblack: This looks pretty good.  I would flip the order of the parameters, if it's trivial.

```suggestion
allocator%allocate_array( array, [10] )
```
The FORTRAN devs are used to:
allocate(var(dim1,dim2,etc))
so when moving to your interface, I think having the pointer var first will feel more natural.

aaroncblack:
Can you call both allocate versions 'allocate' instead of one allocate_array?
ptr = allocate(size)
and
allocate(ptr, [shape])

FORTRAN can't overload on the presence of a return type or not... but it could do it based number of arguments.
The first case has one arg, the second case has two args.

docs/sphinx/tutorial.rst
```diff
@@ -50,3 +50,15 @@ classes that Umpire provides.
    tutorial/c/allocators.rst
    tutorial/c/resources.rst
    tutorial/c/pools.rst
+
+Finally, we have a tutorial for Umpire's FORTRAN API. These examples will be
+compiled whet configuring with ``-DENABLE_FORTRAN=On``. The FORTRAN tutorial
```
mcfadden8:
```suggestion
compiled when configuring with ``-DENABLE_FORTRAN=On``. The FORTRAN tutorial
```

PR #76, opened 2019-03-18
Add ability to print allocations from a specific allocator and visualize

jdahm:
- Add `AllocationMap::print` that takes a predicate and output stream, and test it
- Add `umpire::print_allocator_records` to output records from a specific allocator (using method above)
- Add python script to plot these log files of allocation records
- Add cookbook example `recipe_plot_allocations.cpp` showing an advanced usage with a pool


docs/sphinx/cookbook/plot_allocations.rst
```diff
@@ -0,0 +1,26 @@
...
+name. In this cookbook we use this feature to visualize a pooled allocator.
+
+The cookbook generates two files, `allocator.log` and
+`pooled_allocator.log`, that contain the allocation records from the
+underlying allocator and the pool. These can then be plotted using a
+command similar to the following:
+
+```
+scripts/plot_allocations.py allocator.log gray 0.2 pooled_allocator.log purple 0.8
+```
```
davidbeckingsale: Could you generate an image and include it here?

scripts/plot_allocations.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python
```
davidbeckingsale: Even though this is a script, it would be better in the `tools` sub-directory.

davidbeckingsale: Also, once you move it, please update the CMakeLists.txt to make sure the script is installed.

src/umpire/ResourceManager.cpp
```diff
@@ -569,4 +570,15 @@ ResourceManager::getNextId() noexcept
   return m_id++;
 }
 
+void print_allocator_records(Allocator alloc, std::ostream& os) {
```
davidbeckingsale: Can you move this to Umpire.hpp/cpp, rather than the ResourceManager implementation file. It's not a class method, so shouldn't be grouped in here.

docs/sphinx/cookbook/plot_allocations.rst
jdahm: Could we check if that's working before merging?

scripts/plot_allocations.py
jdahm: 👍

src/umpire/ResourceManager.cpp
jdahm: Done!

examples/cookbook/recipe_plot_allocations.cpp
```diff
@@ -0,0 +1,57 @@
...
+  {
+    std::ofstream out("pooled_allocator.log");
+    umpire::print_allocator_records(pooled_allocator, out);
+    out.close();
+  }
+
+  for (int i = 0; i < 4; ++i) pooled_allocator.deallocate(a[i]);
+
+  // Visualize this using the python script. Example usage:
+  // tools/plot_allocations allocator.log gray 0.2 pooled_allocator.log purple 0.8
```
jdahm: This is a hold-over from calling the method `print_allocation_map`. Now that this is called `print_allocator_records`, should the cookbook be called `plot_allocator_records`?

docs/sphinx/cookbook/plot_allocations.rst
jdahm: It works for now. Fixing `make docs` will be addressed in a separate PR.

davidbeckingsale:
@jdahm - this needs some changes to get it working after the merge with `develop`. Switch from `std::shared_ptr` needs to be made.

jdahm:
This should be ready now.

PR #78, opened 2019-03-19
Remove coalesce from ResourceManager

davidbeckingsale:


benchmarks/allocator_benchmarks.cpp
```diff
@@ -50,11 +49,6 @@ class allocatorBenchmark : public ::benchmark::Fixture {
     }
   }
 
-  void coalesceBench(benchmark::State &st) {
```
mcfadden8: I think I am okay with the removal of this benchmark?  What was the reason?

davidbeckingsale: Coalesce is no longer accessible.

mcfadden8: Not directly

PR #79, opened 2019-03-20
Print list of Allocators when one isn't found

davidbeckingsale:
Example error:

```
[ERROR][/Users/beckingsale1/Projects/umpire/code/umpire/src/umpire/ResourceManager.cpp:195]: getAllocationStrategy Allocator "BANANA" not found. Available allocators: HOST
```

### DIFF
--- a/src/umpire/ResourceManager.cpp
+++ b/src/umpire/ResourceManager.cpp
@@ -183,7 +183,7 @@ ResourceManager::getAllocationStrategy(const std::string& name)
   UMPIRE_LOG(Debug, "(\"" << name << "\")");
   auto allocator = m_allocators_by_name.find(name);
   if (allocator == m_allocators_by_name.end()) {
-    auto available_names = getAvailableAllocators();
+    auto available_names = getAllocatorNames();
 
     const char* const delim = " ";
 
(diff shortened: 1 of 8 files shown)


davidbeckingsale:
Should we do a similar thing for Allocator ids?

src/umpire/ResourceManager.cpp
```diff
@@ -214,7 +225,14 @@ ResourceManager::getAllocator(int id)
 
   auto allocator = m_allocators_by_id.find(id);
   if (allocator == m_allocators_by_id.end()) {
-    UMPIRE_ERROR("Allocator \"" << id << "\" not found.");
+    auto available_ids = getAllocatorIds();
+
+    const char* const delim = " ";
+
+    std::ostringstream id_list;
```
mcfadden8: I think it would perhaps be more useful to display both the ID and the Name.  Perhaps a helper function that both errors could use that will display ID=XXX, Name=SSS?

davidbeckingsale: You think? I was thinking that if someone was searching by id they wouldn't be concerned about the names...

mcfadden8: I was thinking that getting a list of IDs without names might not be helpful.  If they get just a list of IDs, how will they know which one is the right one?

At any rate, I don't think the additional information would hurt.

davidbeckingsale: I have a proposal... we could implement `<<` for Allocator, and have that print name and id?

davidbeckingsale: This is resolved in latest changes.

davidbeckingsale:
@mcfadden8 this is ready to go!

PR #82, opened 2019-03-21
Fix `make docs`

jdahm:
Umpire's `make docs` was designed for use with read the docs, so building it through CMake didn't quite work. Specifically, read the docs did an in-place build, so separating the binary and source directories caused issues. This PR:

* Fixes syntax in the NUMA move cookbook
* Updates BLT for a feature that was recently added
* Ignores the `api/` directory created in source
* Adds a Sphinx `conf.py.in` for use with CMake.


jdahm:
Travis is hiding an issue. This build fails for me using CMake 3.13.3 with the following error:
```
CMake Error: install(EXPORT "umpire-targets" ...) includes target "umpire" which requires target "umpire_resource" that is not in the export set.
CMake Error: install(EXPORT "umpire-targets" ...) includes target "umpire" which requires target "umpire_strategy" that is not in the export set.
CMake Error: install(EXPORT "umpire-targets" ...) includes target "umpire" which requires target "umpire_op" that is not in the export set.
CMake Error: install(EXPORT "umpire-targets" ...) includes target "umpire" which requires target "umpire_util" that is not in the export set.
CMake Error: install(EXPORT "umpire-targets" ...) includes target "umpire" which requires target "umpire_tpl_judy" that is not in the export set.
```
Once this is addressed we can de-draft and prep for merge.

jdahm:
After checking that this hasn't broken RTD builds, this should be ready to go. I'm not promising that the Sphinx output is identical to that on RTD, but it generates nice enough looking HTML to get a sense of what cookbooks and API docs will look like online.

jdahm:
This can also be reviewed now.

PR #81, opened 2019-03-20
Remove shared_ptr

davidbeckingsale:
Unnecessary because we don't have shared ownership of objects, and `shared_ptr` adds overhead.

PR #83, opened 2019-03-21
Replay can output Allocator operations by name and sequence ID

mcfadden8:
Replay can produce an output file containing all of the allocation operations using the string name of the Allocator and a sequence ID of the allocation.

### DIFF
--- a/tests/integration/replay/replay_tests.cpp
+++ b/tests/integration/replay/replay_tests.cpp
@@ -16,18 +16,16 @@
 #include <vector>
 #include <string>
 
-#include "umpire/config.hpp"
-#include "umpire/ResourceManager.hpp"
-
-#include "umpire/strategy/SlotPool.hpp"
-#include "umpire/strategy/MonotonicAllocationStrategy.hpp"
-#include "umpire/strategy/DynamicPool.hpp"
-#include "umpire/strategy/AllocationStrategy.hpp"
 #include "umpire/Allocator.hpp"
+#include "umpire/ResourceManager.hpp"
 #include "umpire/op/MemoryOperation.hpp"
+#include "umpire/strategy/AllocationStrategy.hpp"
 #include "umpire/strategy/AllocationAdvisor.hpp"
-#include "umpire/strategy/ThreadSafeAllocator.hpp"
+#include "umpire/strategy/DynamicPool.hpp"
 #include "umpire/strategy/FixedPool.hpp"
+#include "umpire/strategy/MonotonicAllocationStrategy.hpp"
+#include "umpire/strategy/SlotPool.hpp"
+#include "umpire/strategy/ThreadSafeAllocator.hpp"
 
 class replayTest {
 public:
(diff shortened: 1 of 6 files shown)


mcfadden8:
@davidbeckingsale, this is ready for review.  Replay has been modify to take an optional second command line argument.  If the second command line argument is given, it will be used as the name of the file where it will store the allocator name, operation name, and unique operation sequence number for each operation.  This generated file is then used for the tests to verify that the replay is operating correctly.

tests/integration/replay/replay_tests.cpp
```diff
@@ -72,11 +57,13 @@ class replayTest {
       "thread_safe_allocator", rm.getAllocator("HOST"));
     allocatorNames.push_back("thread_safe_allocator");
 
+#if 0
```
davidbeckingsale: Why is this `#if 0`?

tests/integration/replay/test_output.good
```diff
@@ -0,0 +1,46 @@
+makeMemoryResource HOST
+makeAllocator <umpire::strategy::DynamicPool>(host_simpool_defaults, getAllocator(HOST))
+makeAllocator <umpire::strategy::DynamicPool>(host_simpool_spec1, getAllocator(HOST), 9876, 1234)
+makeAllocator <umpire::strategy::DynamicPool>(host_simpool_spec2, getAllocator(HOST), 9876, 1234)
+makeAllocator <umpire::strategy::MonotonicAllocationStrategy>(MONOTONIC 1024, 1024, getAllocator(HOST))
+makeAllocator <umpire::strategy::SlotPool>(host_slot_pool, 64, getAllocator(HOST))
+makeAllocator <umpire::strategy::ThreadSafeAllocator>(thread_safe_allocator, getAllocator(HOST))
+allocate (17) host_simpool_defaults --> 1
```
davidbeckingsale: How do we get the names of Allocators?

davidbeckingsale: Or were we already recording it?

tests/integration/replay/replay_tests.cpp
mcfadden8: It has been there ever since we realized that replay cannot support FixedBlock allocations.  This will become problematic with the upcoming changes for the pool of fixed pool strategy.

davidbeckingsale: Got it 👍

mcfadden8: I can remove this entirely if you like.  Or, I can ifdef it some other way and add a comment speaking to the limitation (probably the best idea)

davidbeckingsale: Explaining the limitation would be a good idea.

tests/integration/replay/test_output.good
mcfadden8: The names were recorded at makeAllocator time.  Replay established a mapping of the name to the object reference.  It later looks for the object reference of other allocations and then performs the allocation by getting the allocator for the mapped name.

tests/integration/replay/replay_tests.cpp
mcfadden8: Done, will you resolve this?

mcfadden8:
@davidbeckingsale, I will wait until after you have merged in #81, then I will merge those changes and will get your final review.

davidbeckingsale:
#81 is merged!

mcfadden8:
@davidbeckingsale, everything merged in. Things build.  One more review please?

PR #84, opened 2019-03-21
Bugfix/corbett/leak

corbett5:
Fixes multiple memory leaks. Now  `allocator_integration_tests`, `strategy_tests` and `operation_tests` all pass `memcheck` with no reported errors. To try for yourself run `memcheck` with the provided suppression file `scripts/valgrind.supp` via 
`memcheck --suppressions=../scripts/valgrind.supp ./bin/tests/integration/test_to_run`
I've tested this on both quartz and ray with cuda enabled.


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "blt"]
 	path = blt
-	url = https://github.com/LLNL/blt.git
+	url = ../blt.git
```
davidbeckingsale: We don't want this change.

.gitmodules
```diff
@@ -1,6 +1,6 @@
...
 [submodule "src/umpire/tpl/conduit"]
 	path = src/umpire/tpl/conduit
-	url = https://github.com/LLNL/conduit.git
+	url = ../conduit.git
```
davidbeckingsale: Or this one.

src/umpire/ResourceManager.cpp
```diff
@@ -250,6 +267,26 @@ ResourceManager::registerAllocator(const std::string& name, Allocator allocator)
   m_allocators_by_name[name] = allocator.getAllocationStrategy();
 }
 
+void
+ResourceManager::removeAllocator(Allocator allocator)
+{
```
davidbeckingsale: I think we need to discuss how this would work. I'm not sure ripping out all the Allocators sharing a strategy is a good idea.

src/umpire/tpl/judy/judyL2Array.h
```diff
@@ -217,7 +217,7 @@ class judyL2Array {
         bool removeEntry( JudyKey key ) {
             if( 0 != ( _lastSlot = ( vector ** ) judy_slot( _judyarray, ( const unsigned char * ) &key, _depth * JUDY_key_size ) ) ) {
                 // _lastSlot->~vector(); //for use with placement new
-                // delete _lastSlot;
+                delete *_lastSlot;
```
corbett5: @rrsettgast this leak in particular is what caused the issue in geos.

src/umpire/strategy/AllocationAdvisor.cpp
```diff
@@ -66,11 +66,11 @@ AllocationAdvisor::AllocationAdvisor(
 void* AllocationAdvisor::allocate(size_t bytes)
 {
   void* ptr = m_allocator->allocate(bytes);
-  auto alloc_record = new util::AllocationRecord{ptr, bytes, this->shared_from_this()};
+  util::AllocationRecord alloc_record {ptr, bytes, this->shared_from_this()};
```
davidbeckingsale: Nice fix.

src/umpire/tpl/judy/judyL2Array.h
davidbeckingsale: Are you 100% sure this is right?

tests/integration/strategy_tests.cpp
```diff
@@ -599,3 +614,15 @@ TEST(NumaPolicyTest, Location) {
...
 #endif // defined(UMPIRE_ENABLE_NUMA)
+
+
+int main( int argc, char* argv[] )
+{
+  int result = 0;
+  testing::InitGoogleTest( &argc, argv );
+  result = RUN_ALL_TESTS();
+
+  umpire::ResourceManager::getInstance().finalize();
```
davidbeckingsale: This is kind of annoying to do for every test.

tests/integration/strategy_tests.cpp
```diff
@@ -328,10 +343,10 @@ TEST(SizeLimiter, Host)
 
   EXPECT_NO_THROW(data = alloc.allocate(64));
 
-  EXPECT_THROW({
-    void* tmp_data = alloc.allocate(1024);
-    UMPIRE_USE_VAR(tmp_data);
-  }, umpire::util::Exception);
+  // EXPECT_THROW({
```
davidbeckingsale: Please don't comment out tests.

corbett5: This is what's causing the test to fail on ray. For whatever reason the exception is being thrown (you can see it gets printed to the command line) but it isn't caught on blueos. I couldn't figure out how to fix this so I commented it out. It's your call but personally I'd rather just not test this than have a test that fails on a target platform.

tests/integration/strategy_tests.cpp
corbett5: You only have to do it for tests that you want to run memcheck on. If we were to change
``` c++
ResourceManager::getInstance()
{
  if (!s_resource_manager_instance) {
    s_resource_manager_instance = new ResourceManager();
  }

  UMPIRE_LOG(Debug, "() returning " << s_resource_manager_instance);
  return *s_resource_manager_instance;
}
```
 to
``` c++
ResourceManager::getInstance()
{
  static ResourceManager rm;
  s_resource_manager_instance = &rm;
  return rm;
}
```
and then deleted the `ResourceManager` copy/move constructors and the copy/move assignment operators then we could use `~ResourceManager()` to do this. However you already had `finalize` methods elsewhere in the code and I wanted to limit my changes.

src/umpire/tpl/judy/judyL2Array.h
corbett5: No. All I can say is that it looks right, it prevents a serious memory leak, and the tests pass.

.gitmodules
corbett5: I'll remove the commit.

src/umpire/ResourceManager.cpp
corbett5: You're right, I should compare by ID instead of by strategy.

corbett5: Wait, `m_allocators_by_name` only has strategies not allocators, same with `m_allocators_by_id`. If they instead held allocators then I could could compare the allocators by ID.

src/umpire/tpl/judy/judyL2Array.h
mcfadden8: @corbett5, thanks for finding this!  What was the 'issue in geos'?

rrsettgast: if ` _lastSlot` is of type `vector **`, then wouldn't `delete _lastSlot` only delete the `vector**`, leaving the `vector*` hanging out? I assume that was the "leak" we were seeing. However, I can't tell where is the `new` that corresponds with this `delete`. @corbett5 I would dig around a bit to confirm what is the "correct" thing to do....on the other hand, you don't have any more leaks....soooooo....yeah.

src/umpire/ResourceManager.cpp
davidbeckingsale: Allocator ID == strategy ID. What I mean is, we need two different methods: one that "deregisters" a named AllocationStrategy/Allocator, the second that destroys it if and only if it's not registered under another name.

tests/integration/strategy_tests.cpp
davidbeckingsale: I don't think we need to fix this, instead we can suppress it.

tests/integration/strategy_tests.cpp
davidbeckingsale: Can you please try and track it down? It's testing something that should be tested.

I don't think commenting out failing tests is an appropriate solution.

src/umpire/tpl/judy/judyL2Array.h
corbett5: @mcfadden8 A developer created a `chai::ManagedArray` in a tight loop and then ran out of memory on his laptop. When he changed the `chai::ManagedArray` to a vector the memory usage went back to normal. A simple reproducer is to use Umpire to allocate and deallocate memory in an infinite loop then look at the memory usage of the process over time.

@rrsettgast  You'll notice that the call to `delete _lastSlot` was previously commented out. However the line above that is equivalent to the line I added when using placement new.
The corresponding `new`s are here
https://github.com/LLNL/Umpire/blob/725284aad29641c6754f7ca81f4c94d4e5e4cb2a/src/umpire/tpl/judy/judyL2Array.h#L105
and here
https://github.com/LLNL/Umpire/blob/725284aad29641c6754f7ca81f4c94d4e5e4cb2a/src/umpire/tpl/judy/judyL2Array.h#L130

You can see that `delete *_lastSlot` is called in `clear`
https://github.com/LLNL/Umpire/blob/725284aad29641c6754f7ca81f4c94d4e5e4cb2a/src/umpire/tpl/judy/judyL2Array.h#L68

Which with my change makes `removeEntry` basically the inner loop of `clear` which certainly makes sense to me.

src/umpire/ResourceManager.cpp
corbett5: That makes sense to me. This isn't actually integral to this PR, I didn't wind up using it anywhere seeing as the Allocators/AllocationStrategies are now all "destroyed" upon calling `ResourceManager::finalize()`. However I certainly think this functionality is useful for destroying AllocationStrategies that themselves hold a significant amount of data.

tests/integration/strategy_tests.cpp
corbett5: Was your comment intended for the conversation below (The tests case I commented out)? If not I'm not sure I understand what you mean by suppressing it.

davidbeckingsale: I meant, don't have the custom main for the tests, and just suppress the warnings they generate from not calling `ResourceManager::finalize()`

davidbeckingsale: Although the approach you suggest using `static ResourceMangager rm` might be a good solution. Do you know of any major pros/cons of these two different approaches to implement a singleton?

tests/integration/strategy_tests.cpp
corbett5: Will do.

davidbeckingsale: Don't spend long tracking it down - I don't want you to waste time. 

As long as the test is left in, that's fine, even if it's still failing for you on Ray.

tests/integration/strategy_tests.cpp
corbett5: It's thread safe (unlike the current implementation although you could just put a lock around the check). It doesn't involve an `if` check every time and the destructor will automatically be called. The one downside is that with a pointer `ResourceManager::getInstance()' should be valid throughout program termination so any other static objects can access it in their destructor. However if we use `static ResourceMangager rm` then it would depend on the initialization order. @rrsettgast knows more about this than I do. 

https://stackoverflow.com/questions/1008019/c-singleton-design-pattern/1008289#1008289
https://stackoverflow.com/questions/1661529/is-meyers-implementation-of-the-singleton-pattern-thread-safe

tests/integration/strategy_tests.cpp
corbett5: I think it has to do with the level of virtual indirection (2) because if I throw in `AllocationTracker::allocate` it will catch it and if I change the allocator to not use `AllocationTracker` it succeeds.

corbett5:
So now that #81 has been merged it's less clear to me what owns each `AllocationStrategy` object. It looks to me like the obvious candidate is `ResourceManager::m_allocators_by_id`. For example in `ResourceManager::finalize` you wouldn't want to delete every `AllocationStrategy` in `ResourceManager::m_allocators_by_name` because a user could have registered an `Allocator` they created themselves. However  `ResourceManager::m_allocators_by_id` is only populated with allocators that the `ResourceManager` itself created.

If this is the case would it make sense to make `ResourceManager::m_allocators_by_id` hold a `std::unique_ptr` so that the ownership is clear?

src/umpire/ResourceManager.cpp
corbett5: Done.

tests/integration/strategy_tests.cpp
corbett5: This works, is it acceptable to get rid of the `AllocationTracker` for this test?
https://github.com/LLNL/Umpire/blob/d1eac57416a0d873e7b06bac6b322b86e62f0b90/tests/integration/strategy_tests.cpp#L341

corbett5:
Things are working again and leak free after removal of the shared pointers.

src/umpire/tpl/judy/judyL2Array.h
rrsettgast: looks correct.

tests/integration/strategy_tests.cpp
rrsettgast: Do you have other static objects that require access to the ResourceManager? If not, use the Myers singleton. Here is another discussion.

https://stackoverflow.com/questions/13047526/difference-between-singleton-implemention-using-pointer-and-using-static-object

CMakeLists.txt
```diff
@@ -21,20 +21,20 @@ project(Umpire
   LANGUAGES CXX C
   VERSION 0.3.2)
 
-set(ENABLE_OPENMP On CACHE Bool "")
```
davidbeckingsale: What's this change for?

corbett5: `BOOL` is the proper CMake variable type.
https://cmake.org/cmake/help/v3.3/command/set.html

davidbeckingsale:
Closing, but let's keep this branch around. Thanks @corbett5!

PR #96, opened 2019-04-05
Various small fixes

white238:
* Turns on and updates example that was missed (examples/allocator.cxx)
* Fixes various problems with the BGQ Clang 4.0.0 host-config minus an actual test failure in Allocations/StrategyTest but that is probably best left up to someone more knowledgable
* Update SetupBLT message to be a bit more helpful
* Deleted host-config that unless I'm missing something is not going to work anymore due to the compiler not supporting C++11


CMakeLists.txt
```diff
@@ -77,11 +77,12 @@ if (NOT BLT_LOADED)
     set (BLT_SOURCE_DIR ${PROJECT_SOURCE_DIR}/blt CACHE PATH "")
 
     if (NOT EXISTS ${BLT_SOURCE_DIR}/SetupBLT.cmake)
-      message(FATAL_ERROR "\
-      The BLT submodule is not present. \
-      If in git repository run the following two commands:\n \
-      git submodule init\n \
-      git submodule update")
+      message(FATAL_ERROR
+          "The BLT is not present. "
```
```suggestion
          "BLT is not present. "
```

white238: Yea... pushing a change and fixing that in Axom as well...

PR #97, opened 2019-04-08
REPLAY now emits exactly one line per recorded operation.

mcfadden8:


mcfadden8:
@davidbeckingsale, this should be ready for your review now.  Thanks

PR #98, opened 2019-04-08
Um 324 add help option to replay

mcfadden8:


tools/replay.cpp
```diff
@@ -102,8 +103,19 @@ class Replay {
     static void usage_and_exit( const std::string& errorMessage ) {
       std::cerr << errorMessage
       << std::endl
-      << "Usage: replay <replayfile.csv>"
-      << std::endl;
+      << "Usage: replay <options>\n"
+      << "Common Options:\n"
+      << "    -inputfile filename          CSV file created by Umpire library\n"
```
davidbeckingsale: I would prefer using a more robust parser, and having both long and short versions, e.g.: `-i` or `--input`

mcfadden8: That's fair.  I have used GNU getopt() in the past, but I wasn't sure if it is portable enough.  If you're not suggestion Boost, then do you have another one in mind?

tools/replay.cpp
```diff
@@ -482,20 +479,58 @@ class Replay {
...
-  if ( ac < 2 || ac > 3 )
-    Replay::usage_and_exit( "Incorrect number of program arguments" );
+  try
+  {
+    cxxopts::Options options(argv[0], "Replay an umpire session from a file");
+
+    options
+      .add_options()
+      (  "h, help", "Print help")
+      (  "i, infile"
```
davidbeckingsale: Is the second argument the long-form option?

mcfadden8: Yep.  Here is what the help output looks like:

``` script
$ ./bin/tools/replay -h
Replay an umpire session from a file
Usage:
  ./bin/tools/replay [OPTION...]

  -h, --help           Print help
  -i, --infile FILE    Input file created by Umpire library with
                       UMPIRE_REPLAY=On
  -t, --testfile FILE  Test file created used for testing replay only
```

davidbeckingsale: Can you fix this message to make more sense: `Test file created used for testing replay only`, or see if you can "hide" that option?

mcfadden8: @davidbeckingsale, The -t option now has english text and has been hidden since it is used for testing purposes.

tools/replay.cpp
```diff
@@ -482,20 +479,61 @@ class Replay {
...
+       , "Input file created by Umpire library with UMPIRE_REPLAY=On"
+       , cxxopts::value<std::string>(), "FILE"
+      )
+    ;
+
+    options.add_options("Group")
```
davidbeckingsale: Does this hide the -t option?

mcfadden8: Yes it does.   You tell help which group names you want to display and we do not say that it should show "Group".   I will rename "Group" to be something better.  Just a sec...

mcfadden8: Take a look now, this should be better.  Resolve this conversation if you agree..

davidbeckingsale:
💯 - looking good!

mcfadden8:
@davidbeckingsale, can you approve this?

PR #99, opened 2019-04-09
Um 313 make replay output mpi aware

mcfadden8:
With this change, UMPIRE_REPLAY will now include the PID of the calling process.

The replay tool has also been updated to only replay one of the PIDs.  By default, replay will replay the all of the operations from the first PID it finds.  If it finds other PIDs in the file, it will send a single error message per PID to std::cerr saying that it is skipping replays for that particular PID.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -37,10 +37,10 @@ and CUDA
 
 - GitHub action to run Clang static analysis.
 
-- Replay now includes the process ID of the logging process to help
+- Replay now includes unique replay ID of the logging process to help
 distinguish processes in an multi-process run.
 
-- Umpire replay now takes a "-help" option and displays usage information.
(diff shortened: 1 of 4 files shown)


mcfadden8:
@davidbeckingsale, should we consider adding a new option to replay to tell it which PID to replay?  Something like -p or --pid PID#

mcfadden8:
This should be ready for review now.  The command line options look like:
```
$ ./bin/tools/replay -h
Replay an umpire session from a file
Usage:
  ./bin/tools/replay [OPTION...]

  -h, --help         Print help
  -i, --infile FILE  Input file created by Umpire library with
                     UMPIRE_REPLAY=On
  -u, --uid UID      The format of a REPLAY line begins with REPLAY,UID,...
                     This instructs replay to only replay items for the
                     specified UID.
```

PR #105, opened 2019-04-15
Outputting replay in JSON format

mcfadden8:


mcfadden8:
@davidbeckingsale, how do I build with hcc from docker?  I am getting the following error when I try:

```script
$ make help
usage: make [variable] [target]

Build Umpire using Docker!

target:
    gcc                            build with GCC 8
    clang                          build with Clang 6
    nvcc                           build with CUDA 9

variable:
    DEBUG                          display all output if set to 1

For example

    make DEBUG=1 nvcc

builds with the nvcc Docker image and displays all output.
mcfadden8@odinn[LOGIN(umpire) - /Users/mcfadden8/projects/src/umpire/umpire] [ UM-336-replay-to-use-json ]
$ make DEBUG=1 hcc
make: *** No rule to make target `hcc'.  Stop.
```

mcfadden8:
@davidbeckingsale, this is now ready to be review/checkin

tests/integration/replay/test_output.good
```diff
@@ -1,45 +1,94 @@
+makeAllocator <umpire::strategy::MixedPool>(host_mixedpool_default, getAllocator(HOST))
```
davidbeckingsale: Should the "good" output also be JSON?

mcfadden8: hmm..  I suppose it could even though it is not strictly required since it is only a normalized/sequenced view of operations which is only used by the test program (diff).

Should I spend the time to change that too?

mcfadden8: As discussed, leaving this in its current human readable form since it is easier to decipher diff failures if/when they fail.

mcfadden8:
@davidbeckingsale, adding a .readthedocs.yml at the root of our source directory to state we should
use python 3.7 seems to have fixed the document builds on read.the.docs.  You can see/review my replay document here: [read-the-docs](https://umpire.readthedocs.io/en/um-336-replay-to-use-json/tutorial/replay.html)

mcfadden8:
@davidbeckingsale, do you have any time to look at this?

src/umpire/ResourceManager.cpp
```diff
@@ -115,9 +115,10 @@ ResourceManager::initialize()
       UMPIRE_VERSION_MINOR << "." <<
       UMPIRE_VERSION_PATCH);
 
-  UMPIRE_REPLAY( "Umpire v" << UMPIRE_VERSION_MAJOR 
-      << "." << UMPIRE_VERSION_MINOR
-      << "." << UMPIRE_VERSION_PATCH);
+  UMPIRE_REPLAY( "\"event\": \"version\", \"result\": { \"major\":" << UMPIRE_VERSION_MAJOR
```
davidbeckingsale: Should this be "payload" rather than "result"?

mcfadden8: Yeah, I guess I could see that.  I'll change it.

mcfadden8: This is now done.  @davidbeckingsale, please review.

PR #107, opened 2019-04-15
Remove m_ from AllocationRecord struct members

jdahm:
`AllocationRecord`s are simple structs with three public members, so I don't think using `m_` is beneficial for those members.


davidbeckingsale:
I like this in general. Needs an updated CHANGELOG though 😄

davidbeckingsale:
I like this, but the conflicts need resolving.

PR #108, opened 2019-04-15
Error when deallocating with different Allocator than performed the allocation

davidbeckingsale:


src/umpire/resource/DefaultMemoryResource.inl
```diff
@@ -55,8 +56,15 @@ void DefaultMemoryResource<_allocator>::deallocate(void* ptr)
 
   UMPIRE_RECORD_STATISTIC(getName(), "ptr", reinterpret_cast<uintptr_t>(ptr), "size", 0x0, "event", "deallocate");
 
-  deregisterAllocation(ptr);
+  auto record = deregisterAllocation(ptr);
+
+  if (record->m_strategy != this) {
+    UMPIRE_ERROR(ptr << " was not allocated by " << getName());
+  }
+
```
mcfadden8: I really like this!  Since this pattern repeats, I wonder if it would be better in a single helper function?

src/umpire/strategy/mixins/Inspector.cpp
```diff
@@ -47,13 +47,13 @@ Inspector::registerAllocation(
...
 
-void 
+util::AllocationRecord* 
 Inspector::deregisterAllocation(void* ptr)
 {
   auto record = ResourceManager::getInstance().deregisterAllocation(ptr);
 
   m_current_size -= record->m_size;
-  delete record;
+  return record;
```
mcfadden8: Won't this cause a leak?

src/umpire/strategy/AllocationTracker.cpp
```diff
@@ -40,7 +42,12 @@ AllocationTracker::allocate(size_t bytes)
 void
 AllocationTracker::deallocate(void* ptr)
 {
-  deregisterAllocation(ptr);
+  auto record = deregisterAllocation(ptr);
+
+  if (record->m_strategy != this) {
+    UMPIRE_ERROR(ptr << " was not allocated by " << m_allocator->getName());
+  }
+
```
mcfadden8: delete record missing

src/umpire/strategy/mixins/Inspector.cpp
davidbeckingsale: No way around it really, we have to break the abstraction of the Inspector to allow this information to leak out 😞

src/umpire/resource/DefaultMemoryResource.inl
davidbeckingsale: Almost - the difference is due to the error messages. In the AllocationTracker case, since that's not really exposed to the user I pull the underlying allocator name to generate a better error message.

src/umpire/strategy/mixins/Inspector.cpp
mcfadden8: Okay, this is all the more reason to place all of this into a single helper function that checks m_strategy, produces UMPIRE_ERROR, and does a delete

davidbeckingsale: See comment above.

src/umpire/resource/DefaultMemoryResource.inl
mcfadden8: Then may I suggest that you pass in the string of the error message as an argument or throw an exception that the callers may catch?

davidbeckingsale: 😄

src/umpire/util/Macros.hpp
```diff
@@ -126,4 +126,10 @@
 #define UMPIRE_UNLOCK \
   m_mutex->unlock();
 
+#define UMPIRE_CHECK_ALLOCATOR(record, name) \
+  if (record->m_strategy != this) { \
+    UMPIRE_ERROR(ptr << " was not allocated by " << m_allocator->getName()); \
```
mcfadden8: I think you want this to be:
```suggestion
    UMPIRE_ERROR(ptr << " was not allocated by " << name); \
```

davidbeckingsale: I was rushing, thanks for the correction.

src/umpire/resource/CudaConstantMemoryResource.cu
```diff
@@ -67,9 +67,13 @@ void CudaConstantMemoryResource::deallocate(void* ptr)
 {
   UMPIRE_LOG(Debug, "(ptr=" << ptr << ")");
 
-  util::AllocationRecord* record = ResourceManager::getInstance().deregisterAllocation(ptr);
+  auto record = ResourceManager::getInstance().deregisterAllocation(ptr);
```
mcfadden8: Question: can't deregisterAllocation return a copy of the object instead of a pointer?

davidbeckingsale: Yes, it could.

mcfadden8: I don't think it would be too much of a performance hit and that way, the callers would own the copied object.  And since it is an object instead of a object pointer, object deletion would be automatically deleted via scope

PR #110, opened 2019-04-16
Remove ENABLE_ASSERTS and disable OpenMP by default

jdahm:


CHANGELOG.md
```diff
@@ -59,12 +59,16 @@ and the smart pointers added unecessary overhead.
 
 - Moved CHANGELOG to CHANGELOG.md.
 
+- OpenMP changed to off by default.
```
davidbeckingsale: Please put these in the "Unreleased" section of the CHANGELOG

CMakeLists.txt
```diff
@@ -33,8 +33,11 @@ set(ENABLE_WARNINGS_AS_ERRORS On CACHE BOOL "")
 option(ENABLE_EXAMPLES "Build Umpire examples" On)
 option(ENABLE_LOGGING "Build Umpire with Logging enabled" On)
 option(ENABLE_SLIC "Build Umpire with SLIC logging" Off)
-option(ENABLE_ASSERTS "Build Umpire with assert() enabled" On)
-set(ENABLE_GTEST_DEATH_TESTS ${ENABLE_ASSERTS} CACHE BOOL "")
+if (CMAKE_BUILD_TYPE MATCHES "Debug")
```
davidbeckingsale: I don't think option is right here, since this is actually created in BLT. I don't think we should try and set it automatically here, let's just use BLT's default option setting, and allow users to override it with `-D` on the CMake command line.

In short, just get rid of all these lines.

src/umpire/util/Macros.hpp
```diff
@@ -25,13 +25,13 @@
 #include <sstream>
 #include <iostream>
 #include <mutex>
-
-#ifdef UMPIRE_ENABLE_ASSERTS
 #include <cassert>
+
+#ifndef NDEBUG
```
davidbeckingsale: I think this bit is redundant, since http://www.cplusplus.com/reference/cassert/assert/ is a no-op when `NDEBUG` is defined, we can just always call it?

CHANGELOG.md
```diff
@@ -36,13 +38,15 @@ and a dynamic pool for those that are larger.
 
 ### Removed
 
+- `ENABLE_ASSERTS` was removed in factor of checking the `NDEBUG` macro.
```
davidbeckingsale:
```suggestion
- `ENABLE_ASSERTS` was removed in favour of checking the `NDEBUG` macro.
```

PR #121, opened 2019-04-25
Fix ENABLE_NUMA=On

calccrypto:
Fixed mixing of raw pointer and `std::shared_ptr` by removing `std::shared_ptr`
Only call `umpire::numa::preferred_node` when CUDA is enabled, since `numa_node` is only used in a CUDA block

### DIFF
--- a/src/umpire/ResourceManager.cpp
+++ b/src/umpire/ResourceManager.cpp
@@ -430,8 +430,8 @@ ResourceManager::reallocate(void* src_ptr, size_t size, Allocator allocator)
 }
 
 #if defined(UMPIRE_ENABLE_NUMA)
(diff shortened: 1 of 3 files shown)


src/umpire/ResourceManager.cpp
```diff
@@ -430,8 +430,8 @@ ResourceManager::reallocate(void* src_ptr, size_t size, Allocator allocator)
 }
 
 #if defined(UMPIRE_ENABLE_NUMA)
-static std::shared_ptr<strategy::NumaPolicy> cast_as_numa_policy(Allocator& allocator) {
-  std::shared_ptr<strategy::NumaPolicy> numa_alloc;
+static strategy::NumaPolicy* cast_as_numa_policy(Allocator& allocator) {
```
jdahm: Thanks for the fix!

PR #125, opened 2019-05-29
AMD HIP Support in Umpire

noelchalmers:
This PR adds support for AMD's HIP API in Umpire. There are a couple outstanding issues that I'm hoping I could get some help/input on:

1. ~The replay_tests currently fails. I'm unsure how to triage this test.~ Resolved
2. ~The `OBJECT TRUE` setting when passed to `blt_add_library` causes errors in BLT. I suspect this is a mistake in how HIP support in BLT should handle that flag, but I'm unclear on what the correct behavior should be. Could someone with more familiarity with BLT help to diagnose this?~ Resolved


davidbeckingsale:
@noelchalmers - I have tested this locally and it's all working! 🎉 

Could you fix the conflict, and also update the CHANGELOG and we can get this merged.

Also, the branch seems a bit weird still. I think you want to take the commits: `f9bab28` -> `9069c40` and get them so that they are based off develop.

It would be awesome to merge this this week!

.travis.yml
```diff
@@ -62,6 +62,13 @@ matrix:
       - CMAKE_EXTRA_FLAGS="-C ../host-configs/rocm.cmake -DCMAKE_CXX_COMPILER=/opt/rocm/bin/hcc -DENABLE_HCC=On -DENABLE_OPENMP=Off -DENABLE_CUDA=Off -DENABLE_BENCHMARKS=Off -DENABLE_WARNINGS_AS_ERRORS=Off"
       - DO_TEST=no
 
+   - compiler: hip
+     env:
+       - COMPILER=g++
+       - IMD=rocm
+       - CMAKE_EXTRA_FLAGS="-C ../host-configs/hip.cmake"
+       - DO_TEST=no
```
mcfadden8: What is the plan for running CI for this?

davidbeckingsale: Travis, but I'm removing it for now since the Travis build didn't seem to be running. I will add it back in a task branch once this gets merged.

noelchalmers:
Thanks for doing that rebase @davidbeckingsale. I had tried that earlier, but I think git carried over some extra commits from the master branch into this PR.

davidbeckingsale:
No worries @noelchalmers , I think we got it tidied up now.

cmake/SetupCompilerFlags.cmake
```diff
@@ -17,3 +17,7 @@ if (ENABLE_COVERAGE)
   set(CMAKE_CXX_FLAGS "-coverage ${CMAKE_CXX_FLAGS}")
   set(CMAKE_EXE_LINKER_FLAGS "-coverage ${CMAKE_EXE_LINKER_FLAGS}")
 endif ()
+
+if (ENABLE_HIP)
+	set(HIP_HIPCC_FLAGS "${HIP_HIPCC_FLAGS} -Wno-inconsistent-missing-override")
```
mcfadden8: I take it that running the compiler without this flag generates a lot of missing override warnings?  I'm curious, should we consider fixing these?

davidbeckingsale: Good question. I'll defer that one to @noelchalmers

noelchalmers: I'm fairly sure this is a bug on AMD's end. The use of ```override``` seems correct in umpire. I'll look into what's causing this with the hipcc compiler team.

davidbeckingsale: Awesome, we'll keep this flag for now then.

mcfadden8: May we add a comment here or write a Jira ticket so that we don't forget about it?

host-configs/hip.cmake
```diff
@@ -0,0 +1,21 @@
...
+# For details, see https://github.com/LLNL/Umpire
+# Please also see the LICENSE file for MIT license.
+##############################################################################
+
+set(ENABLE_HIP ON CACHE BOOL "")
+set(ENABLE_OPENMP OFF CACHE BOOL "")
+set(ENABLE_CUDA Off CACHE BOOL "")
+
+set(HIP_HIPCC_FLAGS "--amdgpu-target=gfx900" CACHE STRING "")
+set(HIP_ROOT_DIR "/opt/rocm/hip" CACHE PATH "HIP ROOT directory path")
```
mcfadden8: Should the root directory be hard-coded like this?  Should a warning be generated if the root directory cannot be found?

jdahm: It's an option that can be overridden, and something like that is already in the develop branch in `rocm.cmake`. Once there are machines, they can be switched to correct paths.

cmake/SetupCompilerFlags.cmake
noelchalmers: I'll file a Jira ticket today for this.

host-configs/hip.cmake
davidbeckingsale: Yeah this a config that works at LLNL. If you don't supply this file to cmake (using the `-C` option), then the FindHIP module will look in the default locations to try and find everything.

src/umpire/op/HipCopyOperation.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+/*!
+ * \brief Copy operation to move data between two GPU addresses.
+ */
+class HipCopyOperation : public MemoryOperation {
+ public:
+   /*!
+    * @copybrief MemoryOperation::transform
+    *
+    * Uses cudaMemcpy to move data when both src_ptr  and dst_ptr are on AMD
```
mcfadden8:
```suggestion
    * Uses hipMemcpy to move data when both src_ptr  and dst_ptr are on AMD
```

cmake/SetupCompilerFlags.cmake
davidbeckingsale: I think Marty meant a JIRA ticket for us 😄, but if you can get this sorted on your end that would be excellent!

src/umpire/resource/HipConstantMemoryResource.hpp
```diff
@@ -0,0 +1,58 @@
...
+#define UMPIRE_HipConstantMemoryResource_HPP
+
+#include "umpire/resource/MemoryResource.hpp"
+
+#include "umpire/util/AllocationRecord.hpp"
+#include "umpire/util/Platform.hpp"
+
+#include <hip/hip_runtime.h>
+
+__constant__ char umpire_internal_device_constant_memory[64*1024];
```
mcfadden8: Is this configurable?  Or is this fixed to 64*1024?

cmake/SetupCompilerFlags.cmake
noelchalmers: Ah, I didn't catch that :smiley: . Yeah since I think this issue is in our compiler, I'll make a ticket here so that someone on our compiler team can confirm/deny (and fix?) this.

src/umpire/tpl/json/json.hpp
```diff
@@ -2410,7 +2410,7 @@ class input_buffer_adapter : public input_adapter_protocol
     input_buffer_adapter& operator=(input_buffer_adapter&) = delete;
     input_buffer_adapter(input_buffer_adapter&&) = delete;
     input_buffer_adapter& operator=(input_buffer_adapter&&) = delete;
-    ~input_buffer_adapter() override = default;
+    ~input_buffer_adapter() override {};
```
mcfadden8: What was the problem here?

noelchalmers: This was generating a "undefined reference to `vtable for nlohmann::detail::input_buffer_adapter'" error at link time for the replay test. Some Google'ing suggested that some compilers don't like overriding a pure-virtual destructor with a defaulted function. Strange issue, but this change seemed to resolve it.

src/umpire/resource/HipConstantMemoryResource.hpp
noelchalmers: I hadn't thought to make this configurable. Let me check and see whether the constant memory amounts change significantly on different AMD GPUs.

src/umpire/tpl/json/json.hpp
jdahm: Very odd! The `;` is not needed.

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ project(Umpire
   VERSION 0.3.5)
 
 set(ENABLE_CUDA On CACHE BOOL "")
+set(ENABLE_HIP Off CACHE BOOL "")
```
jdahm: Why is CUDA on by default but not HIP? It would be nice to have it consistent. Both options will probably not be a widely used configuration mode, so maybe default both to off? What are your thoughts @davidbeckingsale and @mcfadden8? Probably best done in a separate PR.

src/umpire/op/HipMemsetOperation.hpp
```diff
@@ -0,0 +1,45 @@
...
+    */
+  void apply(
+      void* src_ptr,
+      util::AllocationRecord* ptr,
+      int value,
+      size_t length);
+};
+
+} // end of naemspace op
+} //end of namespace umpire
```
jdahm:
```suggestion
} // end of namespace umpire
```

CMakeLists.txt
davidbeckingsale: @jdahm - they should both be off. Can you create a JIRA ticket?

src/umpire/op/HipCopyFromOperation.hpp
```diff
@@ -0,0 +1,48 @@
...
+  void transform(
+      void* src_ptr,
+      void** dst_ptr,
+      util::AllocationRecord *src_allocation,
+      util::AllocationRecord *dst_allocation,
+      size_t length);
+};
+
+} // end of namespace op
+} //end of namespace umpire
```
jdahm:
```suggestion
} // end of namespace umpire
```

tools/CMakeLists.txt
```diff
@@ -12,10 +12,17 @@
 # For details, see https://github.com/LLNL/Umpire
 # Please also see the LICENSE file for MIT license.
 ##############################################################################
+
+set (tools_depends umpire umpire_tpl_cxxopts umpire_tpl_json)
+
+if (ENABLE_HIP)
+	set (tools_depends ${tools_depends} hip_runtime)
```
jdahm:
```suggestion
  set (tools_depends ${tools_depends} hip_runtime)
```

noelchalmers:
@jdahm Formatting changes are always welcome :smiley:

src/umpire/resource/HipConstantMemoryResource.hpp
mcfadden8: This is probably fine for now.  I was just checking.

noelchalmers: So here's a more lengthy explanation of the "constant" memory cache in AMD hardware: Despite the legacy name, this is actually a read+write cache. It uses scalar memory instructions to efficiently 'broadcast' a memory load to all the threads in a compute unit (CU). On the CUs, there is a dedicated cache for this memory and if a memory request misses in this cache it will spill to L2 and then to global memory. So you can allocate GBs of "constant" memory if you'd like. However, requests may then spill. I think for Vega architectures this hardware cache is 16KB.

src/umpire/tpl/json/json.hpp
jdahm: Did this get removed?

PR #126, opened 2019-05-30
Fix Intel Compile errors with cxxopts and json.

mcfadden8:
Intel was finicky and required adjustments to these two files.


mcfadden8:
@davidbeckingsale, @jdahm, this is ready for review.

.gitmodules
```diff
@@ -6,4 +6,4 @@
 	url = https://github.com/LLNL/conduit.git
 [submodule "src/umpire/tpl/cxxopts"]
 	path = src/umpire/tpl/cxxopts
-	url = https://github.com/jarro2783/cxxopts.git
+  url = https://github.com/LLNL/cxxopts.git
```
jdahm: Tabbing issue

src/umpire/tpl/json/json.hpp
```diff
@@ -12610,6 +12610,14 @@ class serializer
         }
     }
 
+    template<typename NumberType, detail::enable_if_t<
```
jdahm: `std::enable_if_t` was added in C++14. Good, it looks like this is defined by the code on line 620 based on C++11 constructs.

.gitmodules
mcfadden8: Nice catch.  I believe I have addressed this now.

src/umpire/tpl/json/json.hpp
mcfadden8: Is there something further I need to change here?  I'm not sure I follow the "Good" portion of the comment

jdahm: Sorry for the confusion. No, I was just commenting that it's nice that they defined `enable_if_t`!

mcfadden8:
Thanks @jdahm,

@davidbeckingsale ? :-)

PR #127, opened 2019-05-31
Pool allocation records for improved performance

jdahm:
This PR adds `util::FixedMallocPool` and uses this to pool list blocks used for the multimap `util::AllocationMap`. The performance gain is quite significant, roughly 2x for the system allocator. Here are some quick benchmarks done on my laptop:

| alloc size (bytes) | develop time (ns) | PR time (ns) |
|------|--------------|---------|
|    4 |          324 |     163 |
|    8 |          320 |     165 |
|   64 |          344 |     174 |
|  512 |          400 |     189 |
| 1024 |          405 |     231 |

| dealloc size (bytes) | develop time (ns) | PR time(ns)  |
|------|--------------|---------|
|    4 |          418 |     266 |
|    8 |          416 |     267 |
|   64 |          416 |     258 |
|  512 |          501 |     397 |
| 1024 |          464 |     240 |


benchmarks/allocator_benchmarks.cpp
```diff
@@ -262,11 +264,31 @@ class FixedPoolDevice : public ::FixedPool {
 BENCHMARK_DEFINE_F(FixedPoolDevice, allocate)(benchmark::State &st) { allocation(st); }
 BENCHMARK_DEFINE_F(FixedPoolDevice, deallocate)(benchmark::State &st)   { deallocation(st); }
 
+class FixedMallocPool : public ::allocatorBenchmark {
```
jdahm: Do we want these here?

src/umpire/ResourceManager.hpp
```diff
@@ -135,14 +135,14 @@ class ResourceManager {
     /*!
      * \brief register an allocation with the manager.
      */
-    void registerAllocation(void* ptr, util::AllocationRecord* record);
```
jdahm: Add a CHANGELOG.md entry for this change.

src/umpire/strategy/FixedPool.cpp
```diff
@@ -105,7 +105,6 @@ FixedPool::allocInPool(Pool& p) noexcept
     }
   }
 
-  UMPIRE_ASSERT("FixedPool: Logic error in allocate");
```
jdahm: Add an UMPIRE_ERROR here.

benchmarks/allocator_benchmarks.cpp
davidbeckingsale: I think these are fine here for now.

src/umpire/util/AllocationMap.cpp
```diff
@@ -14,249 +14,481 @@
...
 
-AllocationMap::ConstIterator& AllocationMap::ConstIterator::operator++() {
-  if (!end) {
-    ++(data->it);
+// Forward declare const iterator
+class RecordListConstIterator;
 
-    // Use find() here rather than using the last lookup because
-    // judyL2Array begin()/end() cache the state internally.
+// TODO Profile this to determine if we should add a small string
```
davidbeckingsale: Please create a JIRA ticket and remove this TODO.

src/umpire/util/AllocationMap.cpp
```diff
@@ -14,249 +14,481 @@
...
+// RecordListConstIterator
+RecordListConstIterator::RecordListConstIterator(const RecordList* list,
+                                                 bool end)
+  : m_list(list), m_curr(end ? nullptr : m_list->m_tail)
+{
+}
+
+AllocationRecord& RecordListConstIterator::operator*() const
+{
+  // WARNING: Don't call this if at end() -- dereferences a nullptr
```
davidbeckingsale: Can you please enforce this with a check?

src/umpire/util/AllocationMap.cpp
```diff
@@ -14,249 +14,481 @@
...
+
+  // Delete the mutex
   delete m_mutex;
 }
 
-void
-AllocationMap::insert(void* ptr, AllocationRecord* alloc_record)
+void AllocationMap::insert(void* ptr, AllocationRecord record)
 {
   try {
```
davidbeckingsale: Will any of the code in here throw? We might be able to move the try-catch block...

src/umpire/util/AllocationMap.cpp
```diff
@@ -14,249 +14,481 @@
...
+  return (findRecord(ptr) != nullptr);
 }
 
-AllocationMap::ConstIterator AllocationMap::end() const {
-  return ConstIterator(m_records, true);
-}
+void AllocationMap::clear()
+{
+  uintptr_t key = 0;
+  // TODO Why is max = 0 here?
```
davidbeckingsale: Either figure this out and update the comment, or remove the comment and create a JIRA ticket.

src/umpire/util/AllocationMap.hpp
```diff
@@ -12,72 +12,106 @@
 // For details, see https://github.com/LLNL/Umpire
 // Please also see the LICENSE file for MIT license.
 //////////////////////////////////////////////////////////////////////////////
-#ifndef UMPIRE_AllocationMap_HPP
-#define UMPIRE_AllocationMap_HPP
+#ifndef UMPIRE_MemoryMap_HPP
```
davidbeckingsale: Needs to be updated to reflect the filename.

src/umpire/util/AllocationMap.hpp
```diff
@@ -12,72 +12,106 @@
...
 #include <mutex>
 #include <iostream>
-#include <functional>
+#include <iterator>
+
+#include "umpire/util/AllocationRecord.hpp"
 
-template< typename JudyKey, typename JudyValue >
-class judyL2Array;
+// TODO Forward declare Judy and JudySlot
```
davidbeckingsale: Please get rid of the TODO comments.

src/umpire/util/FixedMallocPool.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace util {
+
+/*!
+ * \brief Pool for fixed size allocations using malloc()
+ *
+ * Another version of this class exists in umpire::strategy, but this
+ * version does not rely on Allocator and all the memory tracking
+ * statistics, so it is useful for building objects in umpire::util.
+ */
+class FixedMallocPool
```
davidbeckingsale: Might want to reference where the algorithm comes from?

tests/unit/util/allocation_map_performance_tests.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include <chrono>
 
+// TODO Can this be removed now that there is an inspector test?
```
davidbeckingsale: I think so. You decide 😄

tools/replay_allocation_map.cpp
```diff
@@ -66,7 +66,7 @@ std::istream& operator>>(std::istream& str, CSVRow& data)
 {
```
mcfadden8: This should be removed.  It is not being maintained and is not needed (UM-358)

jdahm: I removed the TODO below and we discussed putting this in a separate PR. Resolving for now.

src/umpire/util/AllocationMap.hpp
```diff
@@ -12,72 +12,106 @@
...
+  AllocationMapConstIterator begin() const;
+  AllocationMapConstIterator end() const;
 
-    std::mutex* m_mutex;
+private:
+  Judy* m_array;
+  mutable JudySlot* m_last; // last found value in Judy
+  unsigned int m_max_levels, m_depth;
+  size_t m_size;
+  std::mutex* m_mutex;
```
jdahm: Can we remove the indirection from mutex locks `UMPIRE_LOCK / UMPIRE_UNLOCK`?

src/umpire/util/AllocationMap.hpp
```diff
@@ -12,72 +12,106 @@
...
-class AllocationMap
+class AllocationMap;
+class RecordList;
+class RecordListConstIterator;
+
+// Note that this iterator is
+class AllocationMapConstIterator : public std::iterator<std::forward_iterator_tag, AllocationRecord>
 {
-  public:
+public:
```
jdahm: This indentation is inconsistent with the rest of Umpire.

src/umpire/util/AllocationMap.cpp
```diff
@@ -14,249 +14,481 @@
...
+{
+  return m_iter->operator->();
+}
+
+AllocationMapConstIterator& AllocationMapConstIterator::operator++()
+{
+  (*m_iter)++;
+  if (*m_iter == m_list->end()) {
+    // Move to a new pointer
+    JudySlot* new_slot = judy_strt(m_array, reinterpret_cast<const unsigned char*>(&m_ptr), 0);
```
jdahm: Is this the correct approach to iterate over the whole array?

src/umpire/strategy/FixedPool.cpp
mcfadden8: Or change to:
```
UMPIRE_ASSERT("FixedPool: Logic error in allocate" && 0);
```

jdahm: Ah, upon a deeper check `allocInPool` should be `noexcept`, because only `allocate` is called and it has an assert check at the bottom before returning. Is that sufficient, @mcfadden8?

src/umpire/util/AllocationMap.cpp
jdahm: Done!

src/umpire/util/AllocationMap.cpp
jdahm: Good catch! I removed the try/catch blocks from here and findRecord.

src/umpire/strategy/FixedPool.cpp
mcfadden8: I'm not sure I understand how `noexcept` is related to an `assert`.  I find it useful to have assertions happen closest to where things first started going wrong.

src/umpire/Umpire.hpp
```diff
@@ -87,7 +87,7 @@ void print_allocator_records(Allocator allocator, std::ostream& os = std::cout);
  *
  * \param allocator source Allocator.
  */
-std::vector<const util::AllocationRecord*> get_allocator_records(Allocator allocator);
+std::vector<util::AllocationRecord> get_allocator_records(Allocator allocator);
```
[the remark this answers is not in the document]
jdahm: I don’t mind using pointers, but the problem came about with the `ResourceManager::deregisterAllocation()`, which had originally returned a pointer. If the `AllocationMap` internally manages its own data (a switch this PR makes), then the pointer isn't valid after the de-registration. `ResourceManager::findAllocationRecord()`, and `AllocationMap::find*()` still return pointers.

If you'd rather this method return a vector of pointers, I think that can work since the allocations are still registered by the `AllocationMap` and that class doesn't move around memory internally

benchmarks/inspector_benchmarks.cpp
```diff
@@ -0,0 +1,105 @@
...
+      st.PauseTiming();
+      register_all(inspector, ptr);
+      i = 0;
+      st.ResumeTiming();
+    }
+    inspector.deregisterAllocation(reinterpret_cast<void*>(ptr[i++]));
+  }
+}
+
+BENCHMARK_MAIN();
```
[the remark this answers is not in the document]
jdahm: Good point. I haven't checked this PR for memory leaks yet...

src/umpire/Umpire.hpp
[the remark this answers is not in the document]
mcfadden8: Okay, this makes sense.  I can't remember what the justification was for using pointers in the first place (@davidbeckingsale?).  Since this PR improves performance by so much, am fine with these changes and we can pick up the pointer thread later if/when we want to deal with improvements to the right of the decimal point.

src/umpire/strategy/FixedPool.cpp
jdahm: I added back an error there.

src/umpire/util/AllocationMap.hpp
mcfadden8: I'm no sure.  Why do you wish to do so?  @davidbeckingsale, @jdahm?

src/umpire/strategy/FixedPool.cpp
```diff
@@ -105,7 +105,9 @@ FixedPool::allocInPool(Pool& p) noexcept
...
+  // The method should either return nullptr at the top, or within the
+  // loop. This should never be reached.
+  UMPIRE_ERROR("A logic error occured");
```
mcfadden8: This can be caught which is not what I think we want.  What can the user do with this?

I really think this should be an assertion and not an error as it points to a bug in the Umpire implementation.

src/umpire/util/AllocationMap.hpp
jdahm: Fight the indirection

src/umpire/op/HostReallocateOperation.cpp
```diff
@@ -36,9 +36,11 @@ void HostReallocateOperation::transform(
   ResourceManager::getInstance().deregisterAllocation(src_ptr);
   *dst_ptr = ::realloc(src_ptr, length);
 
+  // Ensure the realloc was successful
+  UMPIRE_ASSERT(*dst_ptr);
+
```
mcfadden8: This one should actually be an UMPIRE_ERROR since we want for the allocation failure to get back to the user so they can recover.

src/umpire/util/AllocationMap.cpp
jdahm: Information on judy indicates this is not always the case. It only iterates over level 0 trees in the array. Fixed in a new version

tests/unit/util/allocation_map_tests.cpp
```diff
@@ -135,43 +154,39 @@ TEST_F(AllocationMapTest, RegisterMultiple)
...
   map.insert(data, record);
   map.insert(data, next_record);
   map.insert(extra_data, extra_record);
 
   map.printAll();
 
-  map.print([this](const umpire::util::AllocationRecord* r) {
-    return r->m_ptr == data;
+  map.print([this](const umpire::util::AllocationRecord& r) {
+    return r.m_ptr == data;
```
jdahm: Remove `m_` from `AllocationRecord` members.

benchmarks/inspector_benchmarks.cpp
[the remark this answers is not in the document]
jdahm: Checked!

src/umpire/resource/DefaultMemoryResource.inl
```diff
@@ -56,10 +56,11 @@ void DefaultMemoryResource<_allocator>::deallocate(void* ptr)
 
   UMPIRE_RECORD_STATISTIC(getName(), "ptr", reinterpret_cast<uintptr_t>(ptr), "size", 0x0, "event", "deallocate");
 
-  auto record = deregisterAllocation(ptr);
-
+  // TODO Would it be better if this uses AllocationTracker?
```
davidbeckingsale: No TODOs!!!!

src/umpire/strategy/AllocationTracker.cpp
```diff
@@ -42,10 +42,11 @@ AllocationTracker::allocate(size_t bytes)
 void
 AllocationTracker::deallocate(void* ptr)
 {
-  auto record = deregisterAllocation(ptr);
-
+  // TODO Is there a way to only search the array once?
```
davidbeckingsale: @jdahm ...

src/umpire/strategy/FixedPool.cpp
```diff
@@ -105,13 +105,14 @@ FixedPool::allocInPool(Pool& p) noexcept
     }
   }
 
-  UMPIRE_ASSERT("FixedPool: Logic error in allocate");
+  UMPIRE_ASSERT("A logic error occured" && 0);
```
davidbeckingsale: Could this be a more descriptive error?

src/umpire/strategy/FixedPool.cpp
```diff
@@ -128,7 +129,9 @@ FixedPool::allocate(size_t bytes)
     ptr = allocate(bytes);
   }
 
-  UMPIRE_ASSERT(ptr);
+  if (!ptr) {
+    UMPIRE_ERROR("FixedPool::allocate(size=" << m_obj_bytes << "): Could not allocate");
```
davidbeckingsale: 👍

jdahm:
The benchmarks from the initial commit are still roughly equivalent to those on the latest commit. Allocations have slowed down about 10ms, but deallocations are sped up by about 10ms.

src/umpire/strategy/FixedPool.cpp
```diff
@@ -105,13 +105,14 @@ FixedPool::allocInPool(Pool& p) noexcept
...
 
-  UMPIRE_ASSERT("FixedPool: Logic error in allocate");
+  UMPIRE_ASSERT("FixedPool::allocInPool(): num_avail > 0, but no available slots" && 0);
   return nullptr;
 }
 
 void*
 FixedPool::allocate(size_t bytes)
 {
+  UMPIRE_ASSERT(!bytes || bytes == m_obj_bytes);
```
mcfadden8: Could you add a comment stating what this is asserting?  It would be helpful to understand the semantics for `bytes`.  What does it mean when `bytes == 0` and what does it mean when `bytes == m_obj_bytes`?

src/umpire/strategy/AllocationTracker.cpp
```diff
@@ -42,10 +42,7 @@ AllocationTracker::allocate(size_t bytes)
 void
 AllocationTracker::deallocate(void* ptr)
 {
-  auto record = deregisterAllocation(ptr);
-
-  UMPIRE_CHECK_ALLOCATOR(record, m_allocator->getName());
```
mcfadden8: Is there a similar check now being done in `deregisterAllocation`?  I do not see this macro being called there.

src/umpire/util/AllocationMap.hpp
```diff
@@ -15,66 +15,104 @@
...
-    void insert(void* ptr, AllocationRecord* record);
+    // Would require a deep copy of the Judy data
+    AllocationMap(const AllocationMap&) = delete;
 
-    AllocationRecord* remove(void* ptr);
+    // Insert a new record -- copies record
+    void insert(void* ptr, AllocationRecord record);
 
-    AllocationRecord* find(void* ptr) const;
+    // Find a record -- throws an exception of the record is not found.
```
mcfadden8:
```suggestion
    // Find a record -- throws an exception if the record is not found.
```

src/umpire/util/AllocationMap.hpp
```diff
@@ -15,66 +15,104 @@
...
+    // the resulting address of a find(ptr) call can be stored
+    // externally until deregistered. Note also that this class
+    // deallocates the AllocationRecord when removed(), so the pointer
+    // will become invalid at that point.
+    const AllocationRecord* find(void* ptr) const;
+    AllocationRecord* find(void* ptr);
 
-    ConstIterator begin() const;
+    // This version of find never throws an exception
+    const AllocationRecord* findRecord(void* ptr) const;
```
mcfadden8:
```suggestion
    const AllocationRecord* findRecord(void* ptr) const noexcept;
```

jdahm:
The only leaks I found and fixed were in the tests, nothing in the core of the library leaked, at least in so far as it was used in `make test`.

src/umpire/strategy/AllocationTracker.cpp
jdahm: Yes. I checked for leaks and found the bug: deallocating after attempting to deregister with the wrong allocator fails because it was already removed from the allocation map. I fixed it adding `findAllocator` to `Inspector`, but that slowed the code down to almost previous performance. To recover the performance I do the check inside `Inspector::deregisterAllocation(void*)`.

src/umpire/util/AllocationMap.hpp
jdahm: Good idea!

src/umpire/util/AllocationMap.hpp
jdahm: Fixed!

PR #130, opened 2019-06-03
Cookbook for enabling Umpire logging

mcfadden8:


davidbeckingsale:
I think this is okay for now, but in general, I don't think features like this should be presented as recipes. It's a fundamental option that applies to the whole library, and recipes were designed to show how to compose Umpire objects for complex use cases.

mcfadden8:
I agree that recipes don't seem lik the right place to put this.  I'm happy to place it somewhere else.  Do you have any suggestions?

mcfadden8:
What do you think of this @davidbeckingsale?

docs/sphinx/tutorial/replay.rst
```diff
@@ -4,21 +4,6 @@
 Replay
 ======
 
-Umpire provides a lightweight replay capability that can be used to investigate
```
davidbeckingsale: This paragraph can stay here.

davidbeckingsale: But you don't want the bit about the list of events.

PR #135, opened 2019-06-05
DynamicPool refactor

jdahm:
- [X] Splits `AllocationMap` into `MemoryMap` (an L2-like array using `FixedMallocPool`) and defines iterators for all. The idea being that `MemoryMap` could be used inside other objects.
- [ ] Refactors `DynamcPool`


jdahm:
I will make another PR with the changes as they are right now, to fix a bug @mcfadden8 has run into when testing.

PR #136, opened 2019-06-06
Fixed bug finding 0-byte allocation

mcfadden8:


src/umpire/util/AllocationMap.cpp
```diff
@@ -114,7 +114,7 @@ AllocationMap::findRecord(void* ptr) const
 
       if (alloc_record &&
           ((static_cast<char*>(parent_ptr) + alloc_record->m_size)
-             > static_cast<char*>(ptr))) {
+             >= static_cast<char*>(ptr))) {
```
jdahm: 👍 good catch.

tests/integration/operation_tests.cpp
```diff
@@ -68,6 +68,23 @@ class OperationTest :
     umpire::Allocator* dest_allocator;
 };
 
+class ZeroCopyTest :
+  public OperationTest
+{
+};
+
+TEST_P(ZeroCopyTest, Zero) {
```
jdahm: Why does this need to be `TEST_P`? Can't we just use `TEST`?

mcfadden8: Sorry @jdahm, I missed this comment when I closed this particular PR.  I have no answer for this.  I merely copy/pasted from what was previously there.  Perhaps @davidbeckingsale knows?

PR #139, opened 2019-06-06
Add debug build for GCC 5 and Clang 6

davidbeckingsale:


.travis.yml
```diff
@@ -12,7 +12,7 @@ matrix:
     env:
     - COMPILER=g++
     - IMG=gcc-5
-    - CMAKE_EXTRA_FLAGS="-DENABLE_CUDA=Off -DENABLE_COVERAGE=On -DENABLE_C=On"
+    - CMAKE_EXTRA_FLAGS="-DENABLE_CUDA=Off -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=On -DENABLE_C=On"
```
mcfadden8: Can we get Debug builds for gcc8 as well?  We also might want to run with ReleaseWithDebug?

davidbeckingsale: Good suggestion. I don't want to blow up the build matrix too much though.

mcfadden8: Okay.  I'll take whatever I can get..

davidbeckingsale:
This should be ready to go @mcfadden8 @jdahm

PR #141, opened 2019-06-09
Fix compilation and test issues

jdahm:
- Splits `AllocationMap` into `MemoryMap` (an L2-like array using `FixedMallocPool`) and defines iterators for all. `MemoryMap` could be used inside other objects
- Fixes issue with uniform initialization using GCC 4.9.3
- Plugs a memory leak that was causing tests to fail with specific compiler versions
- Corrects a corner case when looking for keys in `AllocationMap`


src/umpire/util/AllocationMap.hpp
```diff
@@ -32,87 +32,136 @@
 namespace umpire {
 namespace util {
 
-class AllocationMap;
-class RecordList;
-class RecordListConstIterator;
-
-class AllocationMapConstIterator : public std::iterator<std::forward_iterator_tag, AllocationRecord>
+class RecordList
```
jdahm: This needed to be exposed because AllocationMap uses `MemoryMap<RecordList>`. The iterator also needed to be exposed because it's used inside `AllocationMap::ConstIterator`. If `m_inner_iter` were not exposed by value (e.g. a pointer) then this could be forward declared and put in `AllocationMap.cpp`. That is, however, an uglier solution. Fight the indirection!

src/umpire/util/AllocationMap.hpp
```diff
@@ -32,87 +32,136 @@
...
-    ~AllocationMapConstIterator();
-
-    const AllocationRecord& operator*() const;
-    const AllocationRecord* operator->() const;
-    AllocationMapConstIterator& operator++();
-    AllocationMapConstIterator operator++(int);
-
-    bool operator==(const AllocationMapConstIterator& other);
-    bool operator!=(const AllocationMapConstIterator& other);
+    ConstIterator(const RecordList* list, iterator_begin);
```
jdahm: Added tag dispatching

src/umpire/util/MemoryMap.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include <cstdint>
+#include <iterator>
+#include <utility>
+#include <type_traits>
+
+namespace umpire {
+namespace util {
+
+struct iterator_begin {};
```
jdahm: These could be moved somewhere else. Thoughts?

src/umpire/util/MemoryMap.hpp
```diff
@@ -0,0 +1,141 @@
...
+template <typename V>
+class MemoryMap
+{
+public:
+  using Key = void*;
+  using Value = V;
+  using KeyValuePair = std::pair<Key, Value*>;
+
+  template <bool Const = false>
+  class Iterator_ : public std::iterator<std::forward_iterator_tag, Value> {
```
jdahm: `MemoryMap` includes both const and non-const iterator implementations. `RecordList` and `AllocationMap` only have const iterators for simplicity (and because they aren't needed at the moment).

tests/integration/free_functions_integration_tests.cpp
```diff
@@ -72,4 +72,4 @@ const std::string allocators[] = {
 INSTANTIATE_TEST_CASE_P(
   FreeFunctions,
   FreeFunctionsTest,
-  ::testing::ValuesIn(allocators));
+  ::testing::ValuesIn(allocators),);
```
jdahm: These commas were added to fix pedantic warnings.

src/umpire/util/MemoryMap.hpp
```diff
@@ -0,0 +1,141 @@
...
+namespace util {
+
+struct iterator_begin {};
+struct iterator_end {};
+
+// MemoryMap maps addresses to a templated type Value, using a
+// FixedMallocPool underneath for speed. It is not threadsafe.
+
+template <typename V>
+class MemoryMap
```
jdahm: This class is differs from `judyL2Array` because it knows the full type of the values, and therefore can construct them using placement new using the efficient `FixedMallocPool`.

src/umpire/util/AllocationMap.cpp
```diff
@@ -119,122 +55,115 @@ void RecordList::push_back(AllocationRecord rec)
...
-  auto plist = reinterpret_cast<RecordList**>(m_last);
-
-  if (!*plist) {
-    // if there is no list there, create one and emplace the record
-    (*plist) = new (list_pool.allocate()) RecordList{record};
-  }
-  else {
-    // else, push onto that list
-    (*plist)->push_back(record);
+  std::tie(iter, found) = m_map.get(ptr, record);
```
jdahm: This crazy logic is necessary to avoid an additional call to `judy_key` for efficiency.

src/umpire/util/MemoryMap.inl
```diff
@@ -0,0 +1,346 @@
...
+    UMPIRE_ERROR("Trying to insert at" << ptr << "but already exists");
+  }
+
+  // Create it
+  (*pval) = new (m_pool.allocate()) Value{val};
+
+  // Increment size
+  ++m_size;
+
+  return Iterator{this, ptr};
```
jdahm: Passing `ptr` here again avoids an additional `judy_key` call.

src/umpire/util/MemoryMap.inl
```diff
@@ -0,0 +1,346 @@
...
+{
+  // Find the ptr and update m_oper
+  m_last = judy_slot(m_array, reinterpret_cast<unsigned char*>(&ptr), judy_max);
+  m_oper = reinterpret_cast<uintptr_t>(this);
+  return m_last ? ConstIterator{this, ptr} : ConstIterator{this, iterator_end{}};
+}
+
+template <typename V>
+void
+MemoryMap<V>::remove(void* ptr)
```
jdahm: Remove newline

src/umpire/util/AllocationMap.cpp
```diff
@@ -254,39 +183,24 @@ AllocationRecord* AllocationMap::find(void* ptr)
...
-
-  // If a list was found, return its tail
-  if (list) {
-    alloc_record = list->back();
-
-    if ((reinterpret_cast<unsigned char*>(parent_ptr) + alloc_record->size)
-         >= reinterpret_cast<unsigned char*>(ptr)) {
-      UMPIRE_LOG(Debug, "Found " << ptr << " at " << parent_ptr
-                 << " with size " << alloc_record->size);
+    if ((static_cast<char*>(candidate->ptr) + candidate->size) >= static_cast<char*>(ptr)) {
```
jdahm: Add another check for size > 0 here

CHANGELOG.md
```diff
@@ -82,14 +82,6 @@ and a dynamic pool for those that are larger.
 - Fix CodeCov reporting by explicitly downloading older version of upload
   script.
 
-## [0.3.5] - 2019-06-11
```
mcfadden8: This 0.3.5 block needs to be put back.

README.md
```diff
@@ -1,4 +1,4 @@
-# <img src="https://cdn.rawgit.com/LLNL/Umpire/develop/share/umpire/logo/umpire-logo.png" width="128" valign="middle" alt="Umpire"/>  Umpire v0.3.5
```
mcfadden8: This needs to still be 0.3.5

docs/sphinx/conf.py
```diff
@@ -106,7 +106,7 @@
 # The short X.Y version.
 version = u'0.3'
 # The full version, including alpha/beta/rc tags.
-release = u'0.3.5'
```
mcfadden8: This needs to be 0.3.5

scripts/make_release_tarball.sh
```diff
@@ -15,7 +15,7 @@
 ##############################################################################
 
 TAR_CMD=gtar
-VERSION=0.3.5
```
mcfadden8: Same - Looks like things did not merge correctly from develop -> your branch

src/umpire/interface/c_fortran/genumpiresplicer.py
```diff
@@ -1,17 +1,3 @@
...
-# Created by David Beckingsale, david@llnl.gov
-# LLNL-CODE-747640
-#
-# All rights reserved.
-#
-# This file is part of Umpire.
-#
-# For details, see https://github.com/LLNL/Umpire
-# Please also see the LICENSE file for MIT license.
-##############################################################################
```
mcfadden8: 0.3.5 merge -> your branch

CHANGELOG.md
jdahm: good catch. I used `--strategy=ours`. Will fix now.

PR #134, opened 2019-06-05
Update Travis script to fail if leaks detected

davidbeckingsale:

PR #142, opened 2019-06-11
ResourceManager Instance Changed to Automatic Variable

calccrypto:
Changed ResourceManager::getInstance to return reference to static variable instead of pointer to remove memory leak

### DIFF
--- a/src/umpire/ResourceManager.cpp
+++ b/src/umpire/ResourceManager.cpp
@@ -51,17 +51,13 @@
 
 namespace umpire {
 
-ResourceManager* ResourceManager::s_resource_manager_instance = nullptr;
-
 ResourceManager&
 ResourceManager::getInstance()
 {
-  if (!s_resource_manager_instance) {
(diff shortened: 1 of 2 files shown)


src/umpire/ResourceManager.cpp
```diff
@@ -51,17 +51,13 @@
...
 ResourceManager::getInstance()
 {
-  if (!s_resource_manager_instance) {
-    s_resource_manager_instance = new ResourceManager();
-  }
+  static ResourceManager s_resource_manager_instance;
 
-  UMPIRE_LOG(Debug, "() returning " << s_resource_manager_instance);
-  return *s_resource_manager_instance;
+  UMPIRE_LOG(Debug, "() returning " << &s_resource_manager_instance);
```
jdahm: Should this log message be removed? The method is returning a reference, not a pointer.

davidbeckingsale:
@calccrypto - I'm currently doing the same thing in #134 😄

PR #147, opened 2019-06-13
Helpful changes in MemoryMap for use in CHAI

jdahm:
Makes `MemoryMap` methods more like `std::map`, and  adds `std::lock_guard<T>`.

- [X] `remove` -> `erase`
- [x] Make threadsafe
- [x] Run all tests
- [x] Check performance against develop
- [X] Add a CHANGELOG.md entry


src/umpire/util/MemoryMap.inl
```diff
@@ -59,7 +59,7 @@ MemoryMap<V>::~MemoryMap()
 template <typename V>
 template <typename... Args>
 std::pair<typename MemoryMap<V>::Iterator, bool>
-MemoryMap<V>::get(void* ptr, Args&&... args) noexcept
+MemoryMap<V>::get(Key ptr, Args&&... args) noexcept
```
davidbeckingsale: Was the `void*` just laziness before? Or does this introduce any real changes?

jdahm: It's the same thing. `using Key = void*;`

jdahm:
I checked and the performance is equivalent to that on `develop`

src/umpire/util/AllocationMap.cpp
```diff
@@ -265,20 +268,28 @@ bool AllocationMap::contains(void* ptr) const
...
+AllocationMap::size() const
+{
+  return m_size;
+}
 
 void
 AllocationMap::print(const std::function<bool (const AllocationRecord&)>&& pred,
                      std::ostream& os) const
 {
+  // Aquire lock
```
davidbeckingsale: This is pedantic, but I think these comments above every `lock_guard` are redundant.

jdahm: ok! I'll remove them

davidbeckingsale: thanks :)

jdahm:
This is ready to be reviewed!

PR #149, opened 2019-06-14
Enable pedantic warnings

mcfadden8:


cmake/SetupCompilerFlags.cmake
```diff
@@ -21,3 +21,10 @@ endif ()
 if (ENABLE_HIP)
 	set(HIP_HIPCC_FLAGS "${HIP_HIPCC_FLAGS} -Wno-inconsistent-missing-override")
 endif()
+
+blt_append_custom_compiler_flag(
+  FLAGS_VAR UMPIRE_PEDANTIC_FLAG
+  DEFAULT  "-Wpedantic"
```
jdahm: Do all compilers use this flag? Should an option for pedantic warnings be added to BLT directly?

mcfadden8: Possibly not, that is why I started this "draft" PR.  I wanted to run it through our CI.  In parallel I am also setting up to run with our other compilers on our HPC systems.

mcfadden8: I cannot answer the BLT question.  Perhaps @davidbeckingsale has an opinion on that?

davidbeckingsale: I think we should guard this with a CMake option for now. E.g. `ENABLE_PEDANTIC_WARNINGS`

src/umpire/strategy/DynamicPoolHeuristic.cpp
```diff
@@ -43,7 +43,7 @@ std::function<bool(const strategy::DynamicPool&)> heuristic_percent_releasable(
   float f = (float)((float)percentage / (float)100.0);
 
   return [=] (const strategy::DynamicPool& pool) {
-      const long threshold = f * pool.getActualSize();
+      const std::size_t threshold = (std::size_t)(f * pool.getActualSize());
```
davidbeckingsale: Should use C++ style casts...

cmake/SetupCompilerFlags.cmake
mcfadden8: Done.

davidbeckingsale: Ace, thanks!

CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ set(ENABLE_TESTS On CACHE BOOL "")
 set(ENABLE_BENCHMARKS On CACHE BOOL "")
 set(ENABLE_GMOCK On CACHE BOOL "")
 set(ENABLE_DOCS Off CACHE BOOL "")
+set(ENABLE_PEDANTIC_WARNINGS On CACHE BOOL "")
```
davidbeckingsale: Since this is our own option, it can be an `option`, rather than a `set`.

src/umpire/strategy/DynamicPoolHeuristic.cpp
```diff
@@ -43,7 +43,7 @@ std::function<bool(const strategy::DynamicPool&)> heuristic_percent_releasable(
   float f = (float)((float)percentage / (float)100.0);
 
   return [=] (const strategy::DynamicPool& pool) {
-      const long threshold = f * pool.getActualSize();
+      const std::size_t threshold = static_cast<std::size_t>(f * pool.getActualSize());
```
davidbeckingsale: Can you a comment like "Calculate threshold in bytes from the percentage"? I had to read this twice to make sure the casting was correct.

CMakeLists.txt
mcfadden8: Done.

CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ option(ENABLE_SLIC "Build Umpire with SLIC logging" Off)
 option(ENABLE_STATISTICS "Track statistics for allocations and operations" Off)
 option(ENABLE_COVERAGE "Enable code coverage (with GCC)" Off)
 option(ENABLE_TOOLS "Enable Umpire Development Tools" On)
+option(ENABLE_PEDANTIC_WARNINGS "Enable Pedantic Compiler Warnings" On)
```
davidbeckingsale: 👍

davidbeckingsale: Why is the description in title case? 

```suggestion
option(ENABLE_PEDANTIC_WARNINGS "Enable pedantic compiler warnings" On)
```

src/umpire/strategy/DynamicPoolHeuristic.cpp
mcfadden8: don

CMakeLists.txt
mcfadden8: I don't know.  I just copied from the previous line

davidbeckingsale: I think you wrote that line too 😄

CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ option(ENABLE_SLIC "Build Umpire with SLIC logging" Off)
 option(ENABLE_STATISTICS "Track statistics for allocations and operations" Off)
 option(ENABLE_COVERAGE "Enable code coverage (with GCC)" Off)
 option(ENABLE_TOOLS "Enable Umpire Development Tools" On)
```
```suggestion
option(ENABLE_TOOLS "Enable Umpire development tools" On)
```

davidbeckingsale: `ENABLE_PENDANTIC_WARNINGS` has also enabled pedantic reviewing 😆

mcfadden8: Huh.  I Guess So.

jdahm:
Good finds 👍

mcfadden8:
This is now ready for review.  I have tested also tested this with IBM and Intel compilers.

cmake/SetupCompilerFlags.cmake
```diff
@@ -21,3 +21,18 @@ endif ()
...
+
+if (ENABLE_PEDANTIC_WARNINGS)
+  blt_append_custom_compiler_flag(
+    FLAGS_VAR UMPIRE_PEDANTIC_FLAG
+    DEFAULT  "-Wpedantic"
+    MSVC "/Wall /WX"
+    INTEL "-Wall -Wcheck -wd2259"
+  )
+
+  set(CMAKE_CXX_FLAGS "${UMPIRE_PEDANTIC_FLAG} ${CMAKE_CXX_FLAGS}")
```
davidbeckingsale: Do you still have to "add" the flag like this? I think you can use the blt macro to update `CMAKE_CXX_FLAGS` accordingly.

mcfadden8: Which blt macro do I use to add to `CMAKE_CXX_FLAGS`?  I looked in the documentation, but could not find anything (or I may have missed it).  Can you send a pointer to an example?

davidbeckingsale: I was mistaken. Modifying the variable directly is the correct way.

PR #151, opened 2019-06-18
New DynamicPool

jdahm:
Adds a new and much improved dynamic pool.

- [x] Formal benchmarking results
- [x] Add back coalesce heuristics


jdahm:
This brings the benchmarks down to (malloc/dealloc is ~60ns):
```
PoolHost/allocate/4               306 ns        305 ns    2104320
PoolHost/allocate/8               307 ns        306 ns    2231118
PoolHost/allocate/64              315 ns        314 ns    2193938
PoolHost/allocate/512             303 ns        302 ns    2340354
PoolHost/allocate/1024            350 ns        350 ns    1971670
PoolHost/deallocate/4             399 ns        399 ns    1729065
PoolHost/deallocate/8             397 ns        396 ns    1766597
PoolHost/deallocate/64            249 ns        249 ns    2798299
PoolHost/deallocate/512           247 ns        247 ns    2793129
PoolHost/deallocate/1024          270 ns        269 ns    2686872
FixedPoolHost/allocate/4          270 ns        269 ns    2512554
FixedPoolHost/deallocate/4        136 ns        136 ns    5244978
```
On develop the PoolHost benchmark doesn't even run. This is probably at least a 100x speedup.

davidbeckingsale:
@jdahm - did you figure out why the benchmarks weren't running on `develop`?

src/umpire/Umpire.hpp
```diff
@@ -73,6 +73,8 @@ int get_patch_version()
   return UMPIRE_VERSION_PATCH;
 }
 
+void print_all_records(std::ostream& os = std::cout);
```
jdahm: This method is useful for debugging, but would you like me to remove it?

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ and a dynamic pool for those that are larger.
 
 - Use RAII locks with `std::lock_guard`.
 
+- `DynamicPool` is significantly faster.
```
davidbeckingsale: This is a bit vague, can you make it more specific.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -21,374 +21,282 @@
...
 
-void
-DynamicPool::release()
+void DynamicPool::release()
 {
-  freeReleasedBlocks();
+  UMPIRE_LOG(Debug, "()");
+
+  // Coalesce first so that we are able to release the most memory possible
+  coalesce();
```
jdahm: I added a `coalesce()` call before releasing. Is that okay with you @mcfadden8?

jdahm:
Here are some benchmark results from an LC machine on this branch:
```Malloc/malloc/4                      78 ns         78 ns    8983648
Malloc/malloc/8                      78 ns         78 ns    8982935
Malloc/malloc/64                     77 ns         77 ns    9067000
Malloc/malloc/512                    95 ns         95 ns    7344900
Malloc/malloc/1024                  154 ns        154 ns    4544862
Malloc/free/4                        54 ns         54 ns   12868088
Malloc/free/8                        54 ns         54 ns   12868020
Malloc/free/64                       54 ns         54 ns   12873621
Malloc/free/512                      63 ns         63 ns   11186809
Malloc/free/1024                     74 ns         74 ns    9518914
FixedMallocPool/allocate/4            9 ns          9 ns   81367016
FixedMallocPool/deallocate/4          8 ns          8 ns   92529088
Host/allocate/4                     288 ns        288 ns    2379057
Host/allocate/8                     288 ns        288 ns    2430870
Host/allocate/64                    284 ns        284 ns    2463373
Host/allocate/512                   321 ns        321 ns    2187596
Host/allocate/1024                  376 ns        376 ns    1864352
Host/deallocate/4                   313 ns        313 ns    2237299
Host/deallocate/8                   313 ns        313 ns    2239217
Host/deallocate/64                  265 ns        265 ns    2653869
Host/deallocate/512                 266 ns        266 ns    2629635
Host/deallocate/1024                281 ns        281 ns    2493700
PoolHost/allocate/4                 454 ns        454 ns    1531950
PoolHost/allocate/8                 454 ns        454 ns    1523855
PoolHost/allocate/64                490 ns        490 ns    1476274
PoolHost/allocate/512               461 ns        461 ns    1509496
PoolHost/allocate/1024              507 ns        507 ns    1343538
PoolHost/deallocate/4              1051 ns       1050 ns     670326
PoolHost/deallocate/8               928 ns        928 ns     743320
PoolHost/deallocate/64              734 ns        734 ns     981486
PoolHost/deallocate/512             623 ns        623 ns    1007250
PoolHost/deallocate/1024            620 ns        620 ns     997630
FixedPoolHost/allocate/4            495 ns        495 ns    1418292
FixedPoolHost/deallocate/4          201 ns        201 ns    3498888```

I have been waiting for 10 minutes to no avail for the `PoolHost` benchmarks to run on `develop`.

File: src/umpire/strategy/DynamicPool.hpp [hunk 4]
```diff
@@ -99,7 +102,7 @@ class DynamicPool :
     std::size_t getReleasableSize() const noexcept;
```
jdahm: This method was measuring the total size of the free blocks before, even though some of those blocks were not the head of allocations and could not be released back to the resource. Thus, some of the strategy tests had to change.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -21,374 +21,277 @@
...
+  m_min_alloc_bytes{min_alloc_bytes},
+  m_align_bytes{align_bytes},
+  m_do_coalesce{coalesce_heuristic},
+  m_used_map{},
+  m_free_map{},
+  m_curr_bytes{0},
+  m_actual_bytes{initial_alloc_bytes},
+  m_highwatermark{initial_alloc_bytes}
+{
+  insertFree(m_allocator->allocate(initial_alloc_bytes), initial_alloc_bytes, true);
```
jdahm: This version also allocates the `initial_alloc_bytes` block in the constructor.

tests/integration/allocator_integration_tests.cpp
```diff
@@ -270,8 +271,8 @@ TEST(Allocation, DeallocateDifferent)
   auto& rm = umpire::ResourceManager::getInstance();
 
   auto alloc_one  = rm.getAllocator("HOST");
-  auto alloc_two = rm.makeAllocator<umpire::strategy::DynamicPool>(
-      "POOL_different_deallocate", alloc_one);
+  auto alloc_two = rm.makeAllocator<umpire::strategy::SizeLimiter>(
```
jdahm: This needs to be a SizeLimiter or else the assert on line 136 (another test) fails because it's checking the size of the record list.

tests/integration/replay/test_output.good
```diff
@@ -71,24 +71,34 @@ deallocate host_mixedpool_spec3(17)
...
-deallocate host_simpool_spec1(22)
-deallocate host_simpool_spec2(23)
+deallocate host_dyn_pool_defaults(21)
+deallocate host_dyn_pool_spec1(22)
+deallocate host_dyn_pool_spec2(23)
 deallocate MONOTONIC 1024(24)
 deallocate host_slot_pool(25)
 deallocate thread_safe_allocator(26)
 deallocate host_mixedpool_default(27)
+coalesce internal_dynamic_pool
```
jdahm: The coalesce calls were added. This is somewhat suspect...

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -21,382 +21,330 @@
 #include "umpire/Replay.hpp"
 
 #include <algorithm>
+#include <tuple>
+
+static std::size_t round_up(std::size_t num, std::size_t factor)
```
davidbeckingsale: Can be inline?

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: This is a good change, some users have been confused by the previous behavior.

src/umpire/strategy/DynamicPool.hpp
```diff
@@ -39,129 +38,117 @@ namespace strategy {
...
      */
     DynamicPool(
         const std::string& name,
         int id,
         Allocator allocator,
-        const std::size_t min_initial_alloc_size = (512 * 1024 * 1024),
+        const std::size_t initial_alloc_size = (512 * 1024 * 1024),
         const std::size_t min_alloc_size = (1 * 1024 *1024),
-        Coalesce_Heuristic coalesce_heuristic = heuristic_percent_releasable(100)) noexcept;
+        const int align_bytes = 16,
```
davidbeckingsale: Adding this new default argument in this position breaks current users. I think we should have multiple constructors that delegate down to this one. This way we can have one with alignment and default heuristic, and one with default alignment and a heuristic.

jdahm: Yes, I was worried about backward compatibility. Good idea 👍

src/umpire/strategy/DynamicPool.cpp
mcfadden8: If I understand correctly, the only thing that changed here is the name from `min_initial_alloc_size` to `initial_alloc_bytes`.  If this is so, the only change was the name (which is good), but not the signature.  The only signature change was the addition of the `align_bytes` optional argument at the end.

Please confirm that I understand this correctly.

src/umpire/strategy/DynamicPool.hpp
```diff
@@ -31,129 +30,116 @@ namespace strategy {
...
-     * \param min_initial_alloc_size The minimum size of the first allocation
-     *                               the pool will make.
-     * \param min_alloc_size The minimum size of all future allocations.
-     * \param coalesce_heuristic Heuristic callback function.
+     * \param name Name of this instance of the DynamicPool
+     * \param id Unique identifier for this instance
+     * \param initial_alloc_bytes Size the pool initially allocates
+     * \param min_alloc_bytes The minimum size of all future allocations
+     * \param align_bytes Number of bytes with which to align allocation sizes
+     * \param coalesce_heuristic Heuristic callback function
```
mcfadden8: Please add description of `align_bytes` parameter.

src/umpire/strategy/DynamicPool.cpp
jdahm: The `align_bytes` parameter was added to the signature of the constructor. This new version allocates a free block in the constructor, whereas the version on develop allocates that the first time an allocation is requested.

src/umpire/strategy/DynamicPool.hpp
jdahm: There was already one there, above `coalesce_heuristic`. I moved it below that so it's in order.

src/umpire/strategy/DynamicPool.cpp
mcfadden8: Aha!  Yes!! That is much better!  Thanks for clarifying that for me.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -13,382 +13,336 @@
...
-    iterPrev = iter;
+DynamicPool::~DynamicPool()
+{
+  // Free any unused blocks
+  for (auto& rec : m_free_map) {
+    void* addr;
+    bool is_head;
+    std::tie(addr, is_head) = rec.second;
+    // Deallocate if this is a head
+    if (is_head) m_allocator->deallocate(addr);
```
mcfadden8: Just to be sure I understand, this destructor intentionally could leak if callers have not freed their memory before the destructor is called.  Is this the intention here?  If so, should we add a debug message or something to indicate that all blacks were not on the free map?

jdahm: Yes, the intention is to leak used blocks.

jdahm: I suppose this could also cause application errors, if they intentionally did not deallocate something, and the DynamicPool is deleted. Although, this should only happen at the end of the program if used in the intended way.

mcfadden8: Okay.  I feel like leaving some bread crumbs behind might be useful.  Should we `UMPIRE_ERROR`, `UMPIRE_LOG`, and/or `UMPIRE_ASSERT` here?
@jdahm, @davidbeckingsale?

jdahm: I added an `UMPIRE_LOG`. Is that sufficient?

mcfadden8: This is probably a good start.  Could we add the allocator name and the number of bytes being leaked to the log?  I also think we should use `UMPIRE_LOG(Error` instead of `UMPIRE_LOG(Debug` for this situation.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -13,382 +13,341 @@
...
+  const std::size_t actual_bytes = round_up(bytes, m_align_bytes);
+  Pointer ptr{nullptr};
 
-void DynamicPool::releaseBlock(struct Block *curr, struct Block *prev) {
-  assert(curr != nullptr);
+  // Check if the previous block is a match
+  SizeMap::const_iterator iter{findFreeBlock(actual_bytes)};
 
-  UMPIRE_LOG(Debug, "(" << curr << ", " << prev << ")");
+  // This is optional, but it might help the growth of the pool...
```
mcfadden8: I think this is good.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -13,382 +13,341 @@
...
-      totalBytes -= curr->blockSize;
-      freed += curr->blockSize;
-      m_allocator->deallocate(curr->data);
-
-      if ( prev )   prev->next = curr->next;
-      else          freeBlocks = curr->next;
-
-      blockPool.deallocate(curr);
+    const int64_t left_bytes{static_cast<int64_t>(free_size - actual_bytes)};
+    if (left_bytes > m_align_bytes) {
```
mcfadden8: Just marking this with a comment so that we can discuss the block's lifecycle during our meeting today.

tests/integration/replay/test_output.good
jdahm: We will investigate this further.

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: This sounds reasonable. We decided that AllocationStrategy destructors _should not_ deallocate everything - this way, applications leaking memory would get correctly caught by tools like Valgrind.

jdahm: Resolving: I added an `UMPIRE_LOG(Error` that prints the number of leaks if there are any.

tests/integration/replay/test_output.good
jdahm: Logic fixes removed these 👍

PR #153, opened 2019-06-19
Fixed parsing of AllocationAdvisor allocator creation.

mcfadden8:


mcfadden8:
I was not able to add CI replay tests for this since AllocationAdvisor is a device-only interface.

PR #155, opened 2019-06-19
Make ENABLE_WARNINGS_AS_ERRORS off by default

davidbeckingsale:


jdahm:
Can we only make this OFF by default if compiling in Release mode?

davidbeckingsale:
No.

mcfadden8:
@davidbeckingsale, why can't we have warnings as error for develop builds?

mcfadden8:
I see that this got merged.  What was the rush to get this merged?

PR #156, opened 2019-06-19
Add PID to filenames

davidbeckingsale:


davidbeckingsale:
@mcfadden8 - I think adding the PID might make the second uid in the filename redundant. What do you think?

davidbeckingsale:
I think the filenames always need to keep the same format. Easier for sorting, tools to process, etc.

mcfadden8:
Okay, then I would keep the uid even though it will be 0 most of the time.  There is an ever-so-slight chance that a PID might collide.

davidbeckingsale:
@mcfadden8 - I'll update the documentation.

PR #157, opened 2019-06-19
Only create directories/files when replay or logging are enabled

davidbeckingsale:


tests/integration/io/iomanager_tests.cpp
```diff
@@ -30,14 +29,13 @@ int main(int argc, char** argv) {
   (void) argv;
 #endif
 
-  auto& rm = umpire::ResourceManager::getInstance();
+  umpire::util::MPI::initialize();
+  umpire::util::IOManager::initialize(true, true);
```
mcfadden8: It would be good to test the other three variants too:

- logging not set, replay not set (test that neither path/file created)
- logging not set, replay set (test that only path/replay file created)
- logging set, replay not set (test that only path/logging file created)

davidbeckingsale:
@mcfadden8 - I added the tests as suggested.

PR #166, opened 2019-06-28
Fix initialization order

jdahm:
This addresses the case when ResourceManager::getInstance() is called from a static function. `s_root_io_dir` and `s_file_basename` are not guaranteed to be initialized at that point.


jdahm:
When can `setOutputDir` be called? If it's called after `IOManager::initialize()`, then `s_*_filename` variables will not be updated, and the initialization is a static...

PR #174, opened 2019-07-10
Fix new allocator_benchmarks build with the Intel compiler

jdahm:


jdahm:
Glad it worked! The build errors also occurred with the Intel compiler, because certain methods were only partially overridden.

PR #178, opened 2019-07-12
Merge v1.0.0 changes to develop

davidbeckingsale:


CMakeLists.txt
```diff
@@ -11,7 +11,9 @@ cmake_policy(SET CMP0057 NEW)
 
 project(Umpire
   LANGUAGES CXX C
-  VERSION 0.3.5)
+  VERSION 1.0.0)
+
+set(UMPIRE_VERSION_RC "rc1")
```
davidbeckingsale: In `develop` I will make another PR to set this to be the git hash.

scripts/make_release_tarball.sh
```diff
@@ -7,7 +7,7 @@
 ##############################################################################
 
 TAR_CMD=gtar
-VERSION=0.3.5
+VERSION=1.0.0-rc1
```
davidbeckingsale: This needs to be changed to remove the rc1 prior to release.

CMakeLists.txt
mcfadden8: @davidbeckingsale, when will this other PR be created?  After this is merged into develop?

davidbeckingsale: UM-426

PR #183, opened 2019-07-18
Fix static initialization order

jdahm:


tests/integration/io/io_tests_runner.py
```diff
@@ -23,11 +23,12 @@ def check_output(name, file_object, expected):
     found = False
 
     for line in file_object:
-        if (expected in line.rstrip()):
+        print(line)
+        if expected in line.rstrip():
             found = True
 
-    if (not found):
-        print("{RED}[   ERROR]{END} Got {contents}".format(contents=contents, expected=expected, **formatters))
```
jdahm: This was a bug. `contents` did not exist!

src/umpire/util/io.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const std::string& base_dir,
+    const std::string& name,
+    const int pid,
+    const std::string& extension)
+{
+  int unique_id{0};
+  std::string filename;
+
+  do {
+    filename = base_dir + "/" + name +
```
jdahm: Please double-check the logic here. I removed the stringstream...

src/umpire/ResourceManager.hpp
```diff
@@ -261,9 +261,6 @@ class ResourceManager {
 
     std::mutex m_mutex;
 
-    static const std::string s_null_resource_name;
```
jdahm: I didn't think these variables needed to be exposed outside ResourceManager.

src/umpire/util/io.cpp
```diff
@@ -0,0 +1,181 @@
...
+      }
+      MPI::sync();
+    } else {
+      UMPIRE_ERROR("Cannot create output directory before MPI has been initialized. "
+                   "Please unset UMPIRE_IO_DIR in your environment");
+    }
+  }
+
+  if (enable_log) {
+    static std::ofstream s_log_ofstream{log_filename};
```
jdahm: I moved `s_log_ofstream` and `s_replay_ofstream` inside this scope. Let me know if you see a reason to move them for greater access.

jdahm:
As long as `initialize_io()` is not itself called at the static initialization step, then this is done! If we want to support that strainge use case, then we have a problem with the static `util::OutputBuffer`s. These are changed in `initialize_io()`, at which point they may not be initialized yet.

PR #186, opened 2019-07-22
Fix static initialization issue with constant memory

jdahm:


jdahm:
`allocator_integration_tests` ensures the constant memory resources work, so making the cuda constant memory array be static should be OK.

PR #187, opened 2019-07-23
Remove static pool from AllocationMap

jdahm:


jdahm:
I tested that this fixes the extended static initialization tests on the LC systems.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -277,6 +277,8 @@ void DynamicPool::mergeFreeBlocks()
 {
   using PointerMap = std::map<Pointer, SizeTuple>;
 
+  if (m_free_map.size() < 2) return;
```
jdahm: This is unrelated, but it's good to return here rather than build a map unnecessarily.

mcfadden8: This looks fine, although I would be inclined to move this check above the `using Pointermap` declaration on line 278.  I realize it won't make any runtime difference and it is a cosmetic/style thing.  So, if you want to keep the line here, that is fine too.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -51,23 +51,13 @@ DynamicPool::~DynamicPool()
...
-  }
+  // Warning if blocks are still in use
   if (m_used_map.size() > 0) {
     std::stringstream ss;
-    ss << "Pointers still in use at destruction time.";
-    if (num_heads > 0) {
-      ss << " This will cause " << num_heads << " leak(s).";
-    }
-    UMPIRE_LOG(Error, ss.str());
+    ss << "The following addresses have not been deallocated:";
```
davidbeckingsale: I think you should maybe check the size of the map, because you don't want to print a million leaks...

PR #188, opened 2019-07-23
Print leaked addresses in FixedPool a la DynamicPool

jdahm:


jdahm:
I tested the logic, and it correctly prints warnings like this (after changing the static initialization test):
```$ bin/tests/integration/static_initialization_test
[WARNING][Umpire/src/umpire/strategy/FixedPool.cpp:98]: ~FixedPool There are 2 addresses not deallocated at destruction. This will cause leak(s). Addresses:
	 0x10f4d1200 0x10f4d1400
```

PR #192, opened 2019-07-24
Fixing replay for MemoryResources

davidbeckingsale:
- [ ] Resources need to work with Replay
- [ ] Replay tests should fail if allocators/allocations are not found

### DIFF
--- a/tests/integration/replay/replay_tests.cpp
+++ b/tests/integration/replay/replay_tests.cpp
@@ -8,6 +8,7 @@
 #include <vector>
 #include <string>
 
+#include "umpire/config.hpp"
 #include "umpire/Allocator.hpp"
 #include "umpire/ResourceManager.hpp"
 #include "umpire/op/MemoryOperation.hpp"
@@ -26,6 +27,17 @@ public:
   replayTest() : testAllocations(3), allocationSize(16)
   {
     auto& rm = umpire::ResourceManager::getInstance();
+
+    allocatorNames.push_back("HOST");
+#if defined(UMPIRE_ENABLE_DEVICE)
+    allocatorNames.push_back("DEVICE");
+#endif
+#if defined(UMPIRE_ENABLE_UM)
+    allocatorNames.push_back("UM");
+#endif
+#if defined(UMPIRE_ENABLE_PINNED)
+    allocatorNames.push_back("PINNED");
+#endif
 
     rm.makeAllocator<umpire::strategy::MixedPool>(
           "host_mixedpool_default"
--- a/tools/util/ReplayInterpreter.cpp
+++ b/tools/util/ReplayInterpreter.cpp
@@ -123,7 +123,7 @@ int ReplayInterpreter::getSymbolicOperation( std::string& raw_line, std::string&
     }
     else {
       std::cerr << "Unknown Replay (" << m_json["event"] << ")\n";
-      return -1;
+      exit(1);
     }
 
     compare_ss << std::endl;
@@ -618,7 +618,7 @@ void ReplayInterpreter::replay_makeAllocator( void )
     }
     else {
       std::cerr << "Unknown class (" << type << "), skipping.\n";
-      return;
+      exit(1);
     }
   }
   else {
@@ -640,7 +640,7 @@ void ReplayInterpreter::replay_allocate( void )
 
   if ( n_iter == m_allocator_indices.end() ) {
     std::cerr << "Unknown allocator " << (void*)alloc_obj_p << std::endl;
-    return;           // Just skip unknown allocators
+    exit(1);
   }
 
   const AllocatorIndex& allocator_number = n_iter->second;
@@ -677,7 +677,7 @@ void ReplayInterpreter::replay_deallocate( void )
 
   if ( n_iter == m_allocator_indices.end() ) {
     std::cout << "Unable to find allocator for: " << m_json["payload"]["memory_ptr"] << " deallocation ignored" <<  std::endl;
-    return;           // Just skip unknown allocators
+    exit(1);
   }
 
   const AllocatorIndex& allocator_number = n_iter->second;
@@ -707,7 +707,7 @@ void ReplayInterpreter::replay_release( void )
 
   if ( n_iter == m_allocator_indices.end() ) {
     std::cout << "Unable to find allocator for: " << m_json["payload"]["memory_ptr"] << " release ignored" <<  std::endl;
-    return;
+    exit(1);
   }
 
   const AllocatorIndex& allocator_number = n_iter->second;
--- a/tools/util/ReplayInterpreter.cpp
+++ b/tools/util/ReplayInterpreter.cpp
@@ -37,7 +37,7 @@ void ReplayInterpreter::buildOperations(void)
       replay_makeAllocator();
     }
     else if ( m_json["event"] == "makeMemoryResource" ) {
-      ;
+      replay_makeMemoryResource();
     }
     else if ( m_json["event"] == "allocate" ) {
       replay_allocate();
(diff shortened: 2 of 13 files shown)


mcfadden8:
On it!

tools/util/ReplayInterpreter.cpp
```diff
@@ -123,7 +123,7 @@ int ReplayInterpreter::getSymbolicOperation( std::string& raw_line, std::string&
     }
     else {
       std::cerr << "Unknown Replay (" << m_json["event"] << ")\n";
```
jdahm: Could this be an UMPIRE_ERROR, or is cerr better?

davidbeckingsale: I agree with @jdahm - I think an UMPIRE_ERROR is appropriate here.

tools/util/ReplayInterpreter.cpp
```diff
@@ -618,7 +631,7 @@ void ReplayInterpreter::replay_makeAllocator( void )
     }
     else {
       std::cerr << "Unknown class (" << type << "), skipping.\n";
-      return;
+      exit(1);
```
davidbeckingsale: As above, these should `UMPIRE_ERROR` (for consistency)

tools/util/ReplayInterpreter.cpp
mcfadden8: I'm not sure I agree :).  I would rather have replay limit itself to Umpire interactivity as much as possible and only "replay" the allocations/deallocations.  I would rather not utilize other parts of Umpire if possible.

davidbeckingsale: Fair enough - please wrap this up in a `REPLAY_ERROR` macro to avoid duplicating the std::cerr, exit pattern.

tools/util/ReplayInterpreter.cpp
mcfadden8: As stated above, I'm not sure I agree for this tool.

davidbeckingsale: See my other comment 😄

tests/integration/replay/replay_tests.cpp
```diff
@@ -27,6 +28,17 @@ class replayTest {
...
+#if defined(UMPIRE_ENABLE_DEVICE)
+    allocatorNames.push_back("DEVICE");
+#endif
+#if defined(UMPIRE_ENABLE_UM)
+    allocatorNames.push_back("UM");
+#endif
+#if defined(UMPIRE_ENABLE_PINNED)
+    allocatorNames.push_back("PINNED");
+#endif
+
```
mcfadden8: It is easy enough to generate different comparison files for different configurations.

The tests need to somehow use a different "good" file to compare to based upon the compile-time options above.  The comparison file is currently specified by a bash script.  Are there flags that we can pass in to the bash script that will allow the script to dynamically use a different comparison file based upon what devices and capabilities are present?

tools/util/ReplayInterpreter.cpp
mcfadden8: See my comments on this above.  I think this needs additional discussion.

tools/util/ReplayInterpreter.cpp
mcfadden8: Yes, once we decide what to do, I will make the change everywhere.  This is an application.  Is UMPIRE_ERROR available to users of Umpire?

tests/integration/replay/replay_tests.cpp
davidbeckingsale: Yup, you should be able to pass arguments based on the CMake variables to the bash script. In CMake, the variables won't have the `UMPIRE_` prefix.

tools/util/ReplayOperationManager.cpp
```diff
@@ -907,21 +927,21 @@ void ReplayOperationManager::makeAllocateCont( uint64_t allocation_from_log )
...
-  m_cont_op = new ReplayOperation(m_allocator_array, m_alloc_operations);
+  m_cont_op = new ReplayOperation(m_allocator_array, m_alloc_operations, m_resource_names);
   m_cont_op->makeCoalesce(allocator_name);
   operations.push_back(m_cont_op);
 }
 
 void ReplayOperationManager::makeRelease( int allocator_num )
 {
-  m_cont_op = new ReplayOperation(m_allocator_array, m_alloc_operations);
+  m_cont_op = new ReplayOperation(m_allocator_array, m_alloc_operations, m_resource_names);
```
mcfadden8: There must be a better way of doing this.  I suppose what I should do is create a separate ReplayOperationEnvironment class for environment information set up by the ReplayOperationManager and then shared/utilized by each of the ReplayOperation objects.  I could then pass along a reference to the ReplayOperationEnvironment object to the operations.  Is this a pattern?

tools/util/ReplayInterpreter.cpp
davidbeckingsale: You said you didn't want to use the UMPIRE macro, which is fine. I think this should be wrapped in a replay specific macro to avoid errors due to repetition.

tools/util/ReplayInterpreter.cpp
davidbeckingsale: It is, but you don't need to use it. I just think having a macro for this "print to cerr, exit(1)" pattern is a good idea.

davidbeckingsale:
@mcfadden8 - can we review this together tomorrow? I have a couple of questions that might be easier explained in person.

tools/util/ReplayInterpreter.cpp
mcfadden8: Done.  Please mark resolved.

tools/util/ReplayInterpreter.cpp
mcfadden8: Done, please mark resolved.

tools/util/ReplayInterpreter.cpp
```diff
@@ -55,20 +56,19 @@ void ReplayInterpreter::buildOperations(void)
       if (   m_json["payload"]["major"] != UMPIRE_VERSION_MAJOR
           || m_json["payload"]["minor"] != UMPIRE_VERSION_MINOR
           || m_json["payload"]["patch"] != UMPIRE_VERSION_PATCH ) {
-        std::cerr << "Warning, version mismatch:\n"
+
+        REPLAY_ERROR("Warning, version mismatch:\n"
```
davidbeckingsale: Should we attempt to replay mismatched versions? Maybe only error if the major version changes (from now on).

mcfadden8: Please see [UM-418](https://rzlc.llnl.gov/jira/browse/UM-418), we will address this as part of a separate PR.

tools/util/ReplayOperationManager.cpp
mcfadden8: @davidbeckingsale, As discussed, I will be modifying replay to pass a reference to the replayManager to all of the replayOperations.

davidbeckingsale: At a later date, right?

tests/integration/replay/replay_tests.cpp
mcfadden8: This will be addressed via a separate PR [UM-419](https://rzlc.llnl.gov/jira/browse/UM-419)

tools/util/ReplayOperationManager.cpp
mcfadden8: Oops.

tools/util/ReplayMacros.hpp
```diff
@@ -0,0 +1,20 @@
...
+//
+// SPDX-License-Identifier: (MIT)
+//////////////////////////////////////////////////////////////////////////////
+#ifndef REPLAY_Macros_HPP
+#define REPLAY_Macros_HPP
+
+#include <cstdlib>
+#include <iostream>
+
+#define REPLAY_ERROR( msg )                                                 \
```
davidbeckingsale: 👍

PR #197, opened 2019-07-30
Use a block allocator in the std::map

jdahm:


jdahm:
MS VC doesn't seem to be conforming to (1) https://en.cppreference.com/w/cpp/container/map

PR #199, opened 2019-07-31
Make DynamicPool allocation behavior consistent with previous release

jdahm:


tests/unit/util/CMakeLists.txt
```diff
@@ -16,10 +16,24 @@ blt_add_test(
   NAME exception_tests
   COMMAND exception_tests)
 
+set (allocation_map_tests_depends
+  umpire gtest)
+
+if (ENABLE_CUDA)
+  set (allocation_map_tests_depends
```
mcfadden8: Was this a regression?  How did this new dependency come about and how is this related to this specific PR?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -15,6 +15,7 @@
 #include <cstdlib>
 #include <algorithm>
 #include <sstream>
+#include <new>
```
mcfadden8: Was this added to get std::bad_alloc{} throw?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -114,21 +115,56 @@ DynamicPool::findFreeBlock(std::size_t bytes) const
...
+    }
+    catch (...) {
+      UMPIRE_LOG(Error,
+                 "\n\tUnable to allocate from resource even after giving back free blocks.\n"
+                 "\tThrowing to let application know we have no more memory: "
+                 << getCurrentSize() << " Bytes still allocated to pool\n"
+                 << getFreeBlocks() << " Partially Free Blocks, "
+                 << getInUseBlocks() << " Used Blocks\n"
+        );
+      throw std::bad_alloc{};
```
mcfadden8: I'm not sure about changing the exception value here.  The previous implementation was intending to perform a re-throw of the previously thrown exception (by calling throw with no arguments).

tests/unit/util/CMakeLists.txt
jdahm: I don't know. I saw a HIP build error on the previous commit and added it. BLT isn't carrying over the dependencies from umpire correctly, because e.g. `cuda` implies `cuda_runtime`. It's complicated...

src/umpire/strategy/DynamicPool.cpp
jdahm: Yes.

src/umpire/strategy/DynamicPool.cpp
jdahm: I didn't know that throw without args threw the previous exception. Should I switch it back to that?

davidbeckingsale: Please switch it back - our exceptions contain additonal info that bad_alloc won't, e.g. CUDA error converted to its string description.

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: Please remove this.

src/umpire/strategy/DynamicPool.cpp
mcfadden8: Yes, please switch it back so that umpire exception information information is re-thrown.

PR #202, opened 2019-07-31
Add mixed pool to allocator_benchmarks

jdahm:
- [x] Merge #178


davidbeckingsale:
Did you fix the crashes?

jdahm:
The mixed pool is relatively small (a few MB). In that case, I think may initializes all tests (pools) before running the tests. Filtering to only run a few tests is a workaround. There are too many benchmarks to run if you compile with CUDA.

jdahm:
I'm checking again for the crash...

jdahm:
This should be ready to go now

PR #203, opened 2019-07-31
Using the same mechanism for all installation path evaluations

adrienbernede:
The following changes aim at correcting a difference in the behavior of the two sequences : 
* A classic sequence where the installation directory is specified at configuration time
`cmake .. -DCMAKE_INSTALL_PREFIX=../install`
`cmake --build . -j 12`
`make install`

* A newer sequence where the installation directory is specified during the installation phase
`cmake ..`
`cmake --build . -j 12`
`cmake --install . --prefix ../install`

Before the proposed changes, the newer sequence (valid with cmake 3.15+ only) would lead to some includes being installed in `/usr/local/include` because their destination was evaluated at configuration time, the rest being installed at the specified destination.

While the changes may not have an impact in most cases, unifying the mechanism for installation directory evaluation seemed valuable to me.

**Note 1:** I wasn’t sure how to write a test for this one.
**Note 2:** I found no usage of the `UMPIRE_INCLUDE_DIRS` in `umpire-config.cmake.in`


davidbeckingsale:
@adrienbernede - does `make install` (rather than `cmake --install`) still work with this change?

adrienbernede:
@davidbeckingsale - the short answer is yes. I verified it on my station, but I would definitely complete my evaluation with any test case you would give me.

The detailed answer is a bit different :
According to my tests, the behavior of `make install` hasn’t changed : 
* installing in a custom place with `CMAKE_INSTALL_PREFIX` is still working
* installing in the default location still fails on my station.

On my station, the default behavior is to install in `/usr/local`. However, installing an include directory there, while there is already one (native `/usr/local/include`) has cmake trying to change the permissions on this directory, and I don’t have permissions for that, while I have permissions for adding files and subdirectories.

IMO, this is due to the attempt to install the include directory instead of files and subdirectories in that directory. I will dig a little deeper here, but as far as can tell, the changes in this PR do not change the behavior of make --install.

adrienbernede:
Digging into this permission issue, it seems that the include directory installation here : 

`install(DIRECTORY`
`${CMAKE_BINARY_DIR}/include`
`DESTINATION ${CMAKE_INSTALL_PREFIX})` or `DESTINATION ".")`

Simply leads to the installation of include/umpire/config.hpp
I could simplify things a bit installing just that file instead of a directory.

jdahm:
We merged these changes into #227

PR #205, opened 2019-08-01
Add git hash as RC version

davidbeckingsale:


CMakeLists.txt
```diff
@@ -83,6 +83,11 @@ if (NOT BLT_LOADED)
   include(${BLT_SOURCE_DIR}/SetupBLT.cmake)
 endif()
 
+blt_git_hashcode (HASHCODE umpire_sha1
+                  RETURN_CODE rc
+                  SOURCE_DIR ${PROJECT_SOURCE_DIR})
+set (UMPIRE_VERSION_RC ${umpire_sha1})
+
```
mcfadden8: Will this also work when building from a tar file instead of a git workspace?

davidbeckingsale: No, but this will not appear in a `release`.

PR #206, opened 2019-08-01
Add test for UMPIRE_LOG_LEVEL env var

davidbeckingsale:


scripts/travis/build_and_test.sh
```diff
@@ -36,7 +36,7 @@ if [[ "$DO_BUILD" == "yes" ]] ; then
             cat Testing/Temporary/MemoryChecker.*.log
             exit 1
           fi
-      done < <(ctest -E replay\|io -T memcheck)
+      done < <(ctest -E replay\|io\tlog -T memcheck)
```
mcfadden8: I am not familiar with the wizardry going on with the escaped "|" and tab in this `done` statment.  What is it that this statement is doing?

davidbeckingsale: We're parsing the output of `ctest -T mecheck` in that loop. The `-E replay\|io\|log` is a regular expression for tests to exclude. Since python has about a million leaks, we exclude tests that use a python script.

PR #217, opened 2019-08-09
Feature/umpire ci

adrienbernede:
This feature aims at introducing a first GitLab CI pipeline designed for the LC GitLab instance.
The additional constraint was to design this pipeline so that we could share as much configuration as possible with other projects.

**First action**
The idea is to separate project specific and shareable configuration.

As for now, shareable information mainly consist in toolchain configuration, such as compiler paths, and is placed in the `.gitlab/conf` directory. More generally, the aim is to have .gitlab containing non-project specific gitlab configuration so that it can be easily moved elsewhere later on (submodule ?).

Project specific configuration, such as compiling flags and target release identification, remains at the same place.

Shareable and project configuration are linked by the directory layout which remains the same : system_type/toolchain

**Second action**
The `.gitlab-ci.yml` file is used by GitLab CI as the main configuration file. Using inclusion of files (`include:`) and job properties extensions (`extends:`) the pipeline configuration is split in several places in a modular manner.

`.gitlab-ci.yml` containins project specific configuration of the pipeline.

`scripts/gitlab` contains projects specific scripts for gitlab.

`.gitlab/ci` contains shareable configuration, such as configuration of jobs for different clusters.

In addition, an allocation mechanism is introduced so that on quartz only one resource allocation is required. Note that this allocation is currently using the pdebug queue.

**Discussion**
This is a first step toward a more complex CI process. 

Features included : 
- modular conception
- unique build script
- build and test on quartz and butte
- share a single resource allocation on quartz
- butte and/or quartz jobs can be deactivate using keywords in branch name (_qnone _bnone)

Feature not included : 
- split builds and tests (under development)
- leveling of test (extensive tests for release)
- optimizations using cache ?

**Known limits**
- Some builds or tests are ending in errors both for known or unknown issues. This jobs have been tagged as "allowed to fail" but should be fixed asap.
- We may want to limit the number of builds depending on the context. The Gitlab repo is currently configured mirror the main GitHub repo and trigger the CI for any change in the latter.

### DIFF
--- a/.gitlab/ci/build_quartz.yml
+++ b/.gitlab/ci/build_quartz.yml
@@ -1,0 +1,109 @@
+##############################################################################
+# Copyright (c) 2016-19, Lawrence Livermore National Security, LLC and Umpire
+# project contributors. See the COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (MIT)
+##############################################################################
+
+####
+# This is the share configuration of jobs for quartz
+
+####
+# In pre-build phase, allocate a node for builds
+allocate_resources_build_quartz:
+  tags:
+    - shell
+    - quartz
+  stage: allocate_resources
+  script:
+    - salloc -N 1 -c 36 -p pdebug -t 10 --no-shell --job-name=${BUILD_QUARTZ_ALLOC_NAME}
+  except:
+    - /_qnone/
+
+####
+# In post-build phase, deallocate resources
+# Note : make sure this is run even on build phase failure
+release_resources_build_quartz:
+  tags:
+    - shell
+    - quartz
+  stage: release_resources
+  script:
+    - scancel $(squeue -h --name=${BUILD_QUARTZ_ALLOC_NAME} --format=%A)
+  when: always
+  except:
+    - /_qnone/
+
+####
+# Generic qwartz job, extending build script
+.build_quartz:
+  variables:
+  tags:
+    - shell
+    - quartz
+  stage: build
+  extends: .build_quartz_script
+  except:
+    - /_qnone/
+
+
+####
+# Here are all quartz build jobs
+
+build_quartz_clang_3_9_1:
+  variables:
+    COMPILER: "clang_3_9_1"
+  extends: .build_quartz
+
+build_quartz_gcc_4_9_3:
+  variables:
+    COMPILER: "gcc_4_9_3"
+  extends: .build_quartz
+
+build_quartz_clang_4_0_0:
+  variables:
+    COMPILER: "clang_4_0_0"
+  extends: .build_quartz
+
+build_quartz_cudatoolkit_9_1:
+  variables:
+    COMPILER: "cudatoolkit_9_1"
+  extends: .build_quartz
+
+build_quartz_gcc_6_1_0:
+  variables:
+    COMPILER: "gcc_6_1_0"
+  extends: .build_quartz
+
+build_quartz_gcc_7_1_0:
+  variables:
+    COMPILER: "gcc_7_1_0"
+  extends: .build_quartz
+
+build_quartz_icpc_16_0_4:
+  variables:
+    COMPILER: "icpc_16_0_4"
+  extends: .build_quartz
+  allow_failure: true
+
+build_quartz_icpc_17_0_2:
+  variables:
+    COMPILER: "icpc_17_0_2"
+  extends: .build_quartz
+
+build_quartz_icpc_18_0_0:
+  variables:
+    COMPILER: "icpc_18_0_0"
+  extends: .build_quartz
+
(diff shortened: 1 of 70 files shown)


.gitlab/ci/build_quartz.yml
```diff
@@ -0,0 +1,109 @@
...
+  except:
+    - /_qnone/
+
+
+####
+# Here are all quartz build jobs
+
+build_quartz_clang_3_9_1:
+  variables:
+    COMPILER: "clang_3_9_1"
```
mcfadden8: Can the list of compilers be somehow kept in a common yml file?  It seems like these are currently duplicated among build_{machinename}.yml files.

.gitlab/conf/host-configs/blueos_3_ppc64le_ib_p9
```diff
@@ -0,0 +1 @@
+blueos_3_ppc64le_ib
```
mcfadden8: Is this file complete?

scripts/gitlab/build.sh
```diff
@@ -0,0 +1,20 @@
...
+
+BUILD_DIRECTORY="build_${SYS_TYPE}_${COMPILER}"
+CCONF="host-configs/${SYS_TYPE}/${COMPILER}.cmake" 
+
+rm -rf ${BUILD_DIRECTORY} 2>/dev/null
+mkdir ${BUILD_DIRECTORY} && cd ${BUILD_DIRECTORY}
+
+cmake \
+  -C ../.gitlab/conf/${CCONF} \
+  -C ../${CCONF} \
```
mcfadden8: Where should build-specific feature variations go?  (-DENABLE_TOOLS, -DENABLE_FORTAN, etc?)

.gitlab/ci/build_quartz.yml
adrienbernede: I should have explained this part of the configuration :
1. These parts of the build_{machinename}.yml files should be generated in the future.
2. As far as I know there is no way to factorize the compiler or machine name here :
  - No matrix mechanism at the moment in gitlab
  - Compiler could be specified by a variable, and job name is a variable, but I couldn’t extract a substring form it (no support for bash string processing)
3. I am not sure the naming here is permanent (I don’t think so actually)
4. Sharing some configuration between cluster would add another layer in the modular setup (.gitlab-ci.yml includes .gitlab/ci/build_{cluster}.yml which includes .gitlab/ci/{all_clusters.yml}). IMO it adds unnecessary complexity (at least at the moment) and exceptions will surely appear soon or later (why defining xl compilers for all clusters ?).

Please comment :)

scripts/gitlab/build.sh
adrienbernede: - All project specific information should go into the projects host-configs directory. In other word, everything that couldn’t be shared. This includes build system variables that may not exist in other projects.
 - Tool chain configuration (typically compilers paths) can be shared, hence the creation .gitlab/conf/{conf}.

Reminder : .gitlab it to be shared among projects, but no mechanism allows it at the moment and it is easier to maintain this way, since it is still in conception phase.

.gitlab/conf/host-configs/blueos_3_ppc64le_ib_p9
adrienbernede: Yes. 

This file is a symbolic link to `blueos_3_ppc64le_ib` directory so that the configuration for `blueos_3_ppc64le_ib_p9` is the same as `blueos_3_ppc64le_ib`.

The mechanism is the same as it the project host-configs directory.

host-configs/bgqos_0/clang_4_0_0.cmake
```diff
@@ -4,9 +4,6 @@
 #
 # SPDX-License-Identifier: (MIT)
 ##############################################################################
-set(CMAKE_CXX_COMPILER "/usr/apps/gnu/clang/r284961-stable/bin/bgclang++11" CACHE PATH "")
```
davidbeckingsale: These changes will break our current Bamboo builds

adrienbernede: Can I test bamboo builds on CZ ?

adrienbernede: Are bamboo tests using the same host-configs file as CZ at some point, or are they completely separated ?

davidbeckingsale: Nope, but look at the `scripts/bamboo/build_and_test.sh` and you will see the problem. We are using the project host-config to select compilers.

adrienbernede: This should now be corrected by the third commit.

PR #218, opened 2019-08-12
Fix NUMA strategy cast in ResourceManager::move call

jdahm:


src/umpire/ResourceManager.cpp
```diff
@@ -565,21 +565,8 @@ ResourceManager::reallocate(void* src_ptr, std::size_t size, Allocator allocator
 #if defined(UMPIRE_ENABLE_NUMA)
 static strategy::NumaPolicy* cast_as_numa_policy(Allocator& allocator) {
```
davidbeckingsale: Why don't we just use `unwrap_allocator` directly (I'm unsure what this function is helping with?)

jdahm: `unwrap_allocator` throws an error if it can't cast as that type

jdahm:
@davidbeckingsale is this better?

CHANGELOG.md
```diff
@@ -20,15 +20,19 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 
 ### Changed
 
-- Adjust notifications for CI jobs
-- Use git commit hash as RC version in develop builds
-- Update BLT submodule to fix warnings from CMake 3.14 and warnings from HIP library
+- Adjust notifications for CI jobs.
+
```
davidbeckingsale: Why adding blank lines?

jdahm: For consistency.

jdahm:
I'm going to double check that this works on LC, then merge.

PR #220, opened 2019-08-21
Switch pool creation to use size_t, not int in C API

davidbeckingsale:


src/umpire/interface/c_fortran/typesUmpire.h
```diff
@@ -1,9 +1,3 @@
-//////////////////////////////////////////////////////////////////////////////
-// Copyright (c) 2016-19, Lawrence Livermore National Security, LLC and Umpire
-// project contributors. See the COPYRIGHT file for details.
-//
-// SPDX-License-Identifier: (MIT)
-//////////////////////////////////////////////////////////////////////////////
```
mcfadden8: Did some tools change that caused this copyright to go missing?

src/umpire/interface/c_fortran/typesUmpire.h
```diff
@@ -30,11 +24,11 @@
 extern "C" {
 #endif
 
-struct s_UMP_SHROUD_capsule_data {
+struct s_umpire_SHROUD_capsule_data {
```
mcfadden8: Will these name changes (from `s_UMP_` to `s_umpire_`) cause any externally visible API changes?

src/umpire/interface/c_fortran/wrapResourceManager.cpp
```diff
@@ -214,7 +208,7 @@ size_t umpire_resourcemanager_get_size(umpire_resourcemanager * self,
 
 umpire_allocator * umpire_resourcemanager_make_allocator_pool(
     umpire_resourcemanager * self, const char * name,
-    umpire_allocator allocator, int initial_size, int block,
+    umpire_allocator allocator, size_t initial_size, size_t block,
```
mcfadden8: Shouldn't these be `std::size_t`?

src/umpire/interface/c_fortran/typesUmpire.h
davidbeckingsale: Nope, these are internal.

src/umpire/interface/c_fortran/typesUmpire.h
davidbeckingsale: Yup, I can fix that.

src/umpire/interface/c_fortran/wrapResourceManager.cpp
davidbeckingsale: Nope, because this is a C function, so no namespacing.

mcfadden8: Then shouldn't this file be renamed with a `.c` suffix instead of of `.cpp`?

davidbeckingsale:
@mcfadden8 - this is ready to go.

src/umpire/interface/c_fortran/typesUmpire.h
```diff
@@ -30,11 +30,11 @@
...
```
mcfadden8: Should the header of this file change too?

davidbeckingsale: I'm not sure I understand what you are saying.

davidbeckingsale:
I don't think tests allocating >4GiB will pass on Travis/Azure.

mcfadden8:
We have tests running for devices on bamboo.  If we added some device testing, that should cover it.

src/umpire/interface/c_fortran/typesUmpire.h
mcfadden8: The copyright/boilerplate header for this file seems different.

PR #221, opened 2019-08-22
Generalized Strategy.Device test for all resources

jdahm:


davidbeckingsale:
Neat! Does the `SizeLimiter` work well?

jdahm:
Yeah, the SizeLimiter throws an exception when the threshold is reached, and the DynamicPool catches that and tries again. Very cool!

tests/integration/strategy_tests.cpp
```diff
@@ -119,32 +119,21 @@ TEST_P(StrategyTest, Duplicate)
...
-TEST(Strategy, Device)
+TEST(DynamicPool, LimitedResource)
 {
   auto& rm = umpire::ResourceManager::getInstance();
 
-  auto allocator = rm.getAllocator("DEVICE");
-  void* alloc = nullptr;
+  const std::size_t max_mem{1024 * 1024 * 4};
 
-  // Determine how much memory we can allocate from device
```
mcfadden8: Why are we throwing this out?

jdahm: This part was integral for the test, to make the allocator `throw` internally further down. I replaced it by a `SizeLimiter` strategy, so we don't need this part any more.

jdahm:
This is ready to go!

davidbeckingsale:
@jdahm - was a JIRA ticket created w.r.t. testing for resource allocation failures?

jdahm:
Yes, it's #469 on JIRA!

PR #224, opened 2019-08-23
Don't try and test large allocs in constant memory

davidbeckingsale:


tests/integration/interface/allocator_c_tests.cpp
```diff
@@ -190,10 +190,23 @@ TEST_P(PoolAllocatorCTest, AllocateDeallocateNothing)
...
+#if defined(UMPIRE_ENABLE_PINNED)
+  , "PINNED"
+#endif
+};
+
 INSTANTIATE_TEST_CASE_P(
     Pools,
     PoolAllocatorCTest,
-    ::testing::ValuesIn(allocator_names));
+    ::testing::ValuesIn(pool_names));
```
mcfadden8: I'm not sure I follow how this fixes the problem?  Is the PINNED pool name being checked somewhere?

davidbeckingsale: The new tests allocate over 4Gib, and were added to ensure we are passing unsigned values through the C interface. The constant memory resource only supports 64k, so that's why the test was failing.

PR #226, opened 2019-08-23
Feature/umpire ci

adrienbernede:
This feature aims at introducing a first GitLab CI pipeline designed for the LC GitLab instance.
The additional constraint was to design this pipeline so that we could share as much configuration as possible with other projects.

First action
---
The idea is to separate project specific and shareable configuration.

As for now, shareable information mainly consist in toolchain configuration, such as compiler paths, and is placed in the `.gitlab/conf` directory. More generally, the aim is to have .gitlab containing non-project specific gitlab configuration so that it can be easily moved elsewhere later on (submodule ?).

Project specific configuration, such as compiling flags and target release identification, remains at the same place.

Shareable and project configuration are linked by the directory layout which remains the same : system_type/toolchain

Second action
---
The `.gitlab-ci.yml` file is used by GitLab CI as the main configuration file. Using inclusion of files (`include:`) and job properties extensions (`extends:`) the pipeline configuration is split in several places in a modular manner.

`.gitlab-ci.yml` containins project specific configuration of the pipeline.

`scripts/gitlab` contains projects specific scripts for gitlab.

`.gitlab/ci` contains shareable configuration, such as configuration of jobs for different clusters.

In addition, an allocation mechanism is introduced so that on quartz only one resource allocation is required. Note that this allocation is currently using the pdebug queue.

Discussion
---
This is a first step toward a more complex CI process.

Features included :

 - modular conception
 - unique build script
 - build and test on quartz and butte
 - share a single resource allocation on quartz
 - butte and/or quartz jobs can be deactivate using keywords in branch name (_qnone _bnone)

Feature not included :

 - split builds and tests (under development)
 - leveling of tests (extensive tests for release)
 - optimizations using cache ?

Known limits
---
 - Some builds or tests are ending in errors both for known or unknown issues. This jobs have been tagged as "allowed to fail" but should be fixed asap.
 - We may want to limit the number of builds depending on the context. The Gitlab repo is currently configured mirror the main GitHub repo and trigger the CI for any change in the latter.

Q&A:
---
@mcfadden8
Can the list of compilers be somehow kept in a common yml file? It seems like these are currently duplicated among build_{machinename}.yml files.

@adrienbernede
I should have explained this part of the configuration :

These parts of the build_{machinename}.yml files should be generated in the future.
As far as I know there is no way to factorize the compiler or machine name here :
No matrix mechanism at the moment in gitlab
Compiler could be specified by a variable, and job name is a variable, but I couldn’t extract a substring form it (no support for bash string processing)
I am not sure the naming here is permanent (I don’t think so actually)
Sharing some configuration between cluster would add another layer in the modular setup (.gitlab-ci.yml includes .gitlab/ci/build_{cluster}.yml which includes .gitlab/ci/{all_clusters.yml}). IMO it adds unnecessary complexity (at least at the moment) and exceptions will surely appear soon or later (why defining xl compilers for all clusters ?).

---
@mcfadden8
Is this file complete?

@adrienbernede
Yes.

This file is a symbolic link to blueos_3_ppc64le_ib directory so that the configuration for blueos_3_ppc64le_ib_p9 is the same as blueos_3_ppc64le_ib.

The mechanism is the same as it the project host-configs directory.

---
@mcfadden8
Where should build-specific feature variations go? (-DENABLE_TOOLS, -DENABLE_FORTAN, etc?)

  @adrienbernede
All project specific information should go into the projects host-configs directory. In other word, everything that couldn’t be shared. This includes build system variables that may not exist in other projects.
Tool chain configuration (typically compilers paths) can be shared, hence the creation `.gitlab/conf/{conf}`.
Reminder : .gitlab it to be shared among projects, but no mechanism allows it at the moment and it is easier to maintain this way, since it is still in conception phase.


adrienbernede:
Thanks @davidbeckingsale !

Given the CI results, it’s worth making some additional comments:

1. You can see we have the **feedback from GitLab-CI directly in GitHub** ! The sync is working, great news !
2. You can also see the the repo on GitLab-CI is named "umpire-test-synchro". This is because I was testing the mirroring capability. In fact, I have to GitLab repo mirroring the GitHub repo. The two are using a slightly different configuration, and I was expecting what is happening right now: **one is reporting results, while the other isn’t**. Interesting result, this is not very clear in GitLab documentation so it was worth the try.
3. You can also see that the GitLab-CI pipeline failed ! No worries, I found the issue, and this is because I have those **2 mirror repository, triggering the same pipeline** at the same time, and more specifically creating 2 srun allocations. This should be fine once I delete one of the mirror repo, and then I will strengthen the mechanism used to retrieve the srun allocation.
4. I have some code coverage failing, and I can’t find the details, can someone help me with this ?

PR #228, opened 2019-08-28
[WIP] Clean up DynamicPool

jdahm:


davidbeckingsale:
@mcfadden8 - I changed the destination of this to be #234 , please review and merge.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -115,7 +120,7 @@ DynamicPool::findFreeBlock(std::size_t bytes) const
   return iter;
 }
 
-void* DynamicPool::allocateFromResource(std::size_t bytes)
+void* DynamicPool::allocateFromStrategy(std::size_t bytes)
```
mcfadden8: @jdahm and @davidbeckingsale, I thought I  asked this question before, but I cannot find it now.  Why are we changing the name from `allocateFromResource` to `allocateFromStrategy`?  I thought that `allocateFromResource` was more clear as it at least indicated that it was performing the allocation at least from something that is close to the actual `Resource` than the calling object.  Since the calling object is already a Strategy, it is confusing to me seems recursive.

If `allocateFromResource` isn't right, can we come up with a different, less confusing name?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -153,9 +158,19 @@ void* DynamicPool::allocateFromResource(std::size_t bytes)
       throw;
     }
   }
+
+  // Add to count
+  m_actual_bytes += bytes;
+
```
mcfadden8: Was this moved from someplace?  Or was it previously not be accounted for correctly?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -360,15 +378,14 @@ void DynamicPool::mergeFreeBlocks()
...
+  UMPIRE_LOG(Debug, "() Free blocks after: " << getFreeBlocks());
 }
 
 std::size_t DynamicPool::releaseFreeBlocks()
 {
   UMPIRE_LOG(Debug, "()");
 
-  // Coalesce first so that we are able to release the most memory possible
-  mergeFreeBlocks();
-
```
mcfadden8: Why is this being removed from here?

mcfadden8:
@jdahm, there are multiple changes in this PR ranging from name changes to helper function creation to the removal  of `mergeFreeBlocks` from `releaseFreeBlocks` function.  I'd like to have each of the change-sets rationalized and want to make sure we understand what, if any, changes to the pool life-cycle the changes present.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -115,7 +120,7 @@ DynamicPool::findFreeBlock(std::size_t bytes) const
   return iter;
 }
 
-void* DynamicPool::allocateFromResource(std::size_t bytes)
```
jdahm: `Resource` didn't make sense here, since `m_allocator` is an `AllocationStrategy`

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -153,9 +158,19 @@ void* DynamicPool::allocateFromResource(std::size_t bytes)
       throw;
     }
   }
+
+  // Add to count
+  m_actual_bytes += bytes;
```
jdahm: I changed the `(de)allocateFromStrategy()` calls to update m_actual_bytes at the source of allocation, to avoid mistakes.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -309,6 +324,8 @@ void DynamicPool::mergeFreeBlocks()
 
   using PointerMap = std::map<Pointer, SizeTuple>;
 
+  UMPIRE_LOG(Debug, "() Free blocks before: " << getFreeBlocks());
```
jdahm: I thought this logging statement may be useful. What do you think?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -406,15 +422,10 @@ void DynamicPool::coalesce()
     const std::size_t released_bytes{releaseFreeBlocks()};
     // Deallocated and removed released_bytes from m_free_map
 
-    // If this removed anything from the map, re-allocate a single large chunk
+    // If this removed anything from the map, re-allocate a single large chunk and insert to free map
```
jdahm: The rounding here isn't needed because we always allocate rounded sizes, so `released_size` should always be rounded.

jdahm:
> I'd like to have each of the change-sets rationalized and want to make sure we understand what, if any, changes to the pool life-cycle the changes present.

I gave short explanations for each change. Let me know if they need to be better explained and we can discuss them at the meeting later.

src/umpire/strategy/DynamicPool.cpp
mcfadden8: I guess it might be clearer if we could distinguish allocations from pool versus allocations from someplace else to grow the pool.  I don't like strategy since the pool is a strategy.  If resource doesn't work (I'm not sure why), then can you suggest a different name?  If not, then I would rather have it not change and remain resource.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -79,7 +79,12 @@ DynamicPool::~DynamicPool()
     std::size_t whole_bytes;
     std::tie(addr, is_head, whole_bytes) = rec.second;
     // Deallocate if this is a whole block
-    if (is_head && bytes == whole_bytes) m_allocator->deallocate(addr);
+    if (is_head && bytes == whole_bytes) deallocateFromStrategy(addr, bytes);
+  }
+
+  if (m_used_map.size() == 0) {
```
mcfadden8: Why can't this be `UMPIRE_ASSERT(m_used_map.size() == 0)` ?

src/umpire/strategy/DynamicPool.cpp
jdahm: `releaseFreeBlocks()` is a private method that is only called from `coalesce()` and `allocateFromStrategy()`. `coalesce()` calls `mergeFreeBlocks()` explicitly, so this is an unnecessary operation in that call path.

I agree that this changes the behavior from `allocateFromStrategy()` in the case where it cannot allocate from the resource/strategy. Would you like me to add a call to `mergeFreeBlocks()` there? I'm not sure how much we've tested that ability yet, beyond the relatively simple unit test.

jdahm: Added it!

src/umpire/strategy/DynamicPool.cpp
jdahm: Because that would be wrong. We don't always have that behavior if someone forgets to delete a pointer. If all allocations are freed, there are then certain checks we can make, which happen in debug mode.

src/umpire/strategy/DynamicPool.cpp
jdahm: I switched these to `(de)allocateFromResource()`!

davidbeckingsale: How about `allocateBlock`?

mcfadden8: I like the suggestion from @davidbeckingsale!  Please change it to `(de)allocateBlock()`!

PR #231, opened 2019-09-03
Bugfix/1.0.0/dynamic pool merging

jdahm:


src/umpire/strategy/DynamicPool.cpp
```diff
@@ -305,6 +320,8 @@ void DynamicPool::mergeFreeBlocks()
 
   using PointerMap = std::map<Pointer, SizeTuple>;
 
+  UMPIRE_REPLAY("() Free blocks before: " << getFreeBlocks());
```
mcfadden8: @davidbeckingsale, @jdahm, I think we should limit the changes in this PR to just what is required for the fix.  I'd rather not add new replay semantics here.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -356,15 +374,14 @@ void DynamicPool::mergeFreeBlocks()
...
   }
+
+  UMPIRE_REPLAY("() Free blocks after: " << getFreeBlocks());
 }
 
 std::size_t DynamicPool::releaseFreeBlocks()
 {
   UMPIRE_LOG(Debug, "()");
 
-  // Coalesce first so that we are able to release the most memory possible
```
mcfadden8: Why is this being removed from here and how does it pertain to the fix?

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: Agree.

Adding asserts or logging would be okay however.

src/umpire/strategy/DynamicPool.cpp
jdahm: This was getting called twice through `coalesce`. `releaseFreeBlocks` should literally only release free blocks.

mcfadden8:
Why are we renaming `allocateFromResource()` and `deallocateFromResource()` to `allocateFromStrategy()` and `deallocateFromStrategy()`?

If at all possible, I would prefer that we limit changes to only those that are absolutely required for the upcoming patch.  Limiting the changes will make it easier to review and to look at again later if needed.

src/umpire/strategy/DynamicPool.cpp
jdahm: Oops, I meant that only to be a log. I'll fix that!

src/umpire/strategy/DynamicPool.cpp
mcfadden8: This is kind of a slippery slope to me.  I now understand why `::allocateFromResource()` was calling `::releaseFreeBlocks()` instead of `::release()`.

If possible, I would like to avoid general clean-up in a patch release.

PR #232, opened 2019-09-04
Fix bug where DynamicPool was leaking blocks (v1.0.1)

davidbeckingsale:


src/umpire/strategy/DynamicPool.cpp
```diff
@@ -180,28 +180,27 @@ void* DynamicPool::allocate(std::size_t bytes)
...
     m_actual_bytes += alloc_bytes;
     m_curr_bytes += rounded_bytes;
 
-    const int64_t left_bytes{static_cast<int64_t>(
-        m_min_alloc_bytes - rounded_bytes)};
+    const std::size_t left_bytes{alloc_bytes - rounded_bytes};
 
     // Add free
-    if (left_bytes > m_align_bytes)
+    if (left_bytes > 0)
```
davidbeckingsale: This was the cause of the issue. The other changes make the code consistent.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -42,8 +42,8 @@ DynamicPool::DynamicPool(const std::string& name,
   m_actual_bytes{initial_alloc_bytes},
```
jdahm: This should also get rounded.

davidbeckingsale:
@mcfadden8 - this should be ready to go

mcfadden8:
@davidbeckingsale, Why are build tests failing?

RELEASE_NOTES.md
```diff
@@ -1,3 +1,7 @@
+# v1.0.1
+
+- Fixed a bug in DynamicPool where memory could be leaked.
```
mcfadden8: Could you please be more descriptive here? What conditions would cause the leak?  I also think that the Jira ticket should be referenced.

scripts/make_release_tarball.sh
```diff
@@ -7,7 +7,7 @@
 ##############################################################################
 
 TAR_CMD=gtar
-VERSION=1.0.0
+VERSION=1.0.1
```
mcfadden8: Where is our "magic" version symbol maintained?  I presume it does NOT need to be updated since this is simply a patch-level release.  Is that correct?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -33,17 +33,17 @@ DynamicPool::DynamicPool(const std::string& name,
                          const int align_bytes) noexcept :
   AllocationStrategy(name, id),
   m_allocator{allocator.getAllocationStrategy()},
-  m_min_alloc_bytes{min_alloc_bytes},
+  m_min_alloc_bytes{round_up(min_alloc_bytes, align_bytes)},
```
mcfadden8: Where do we confirm that `align_bytes` is valid?  What if I set `align_bytes=21`, will that work?

mcfadden8: Where do we validate `min_alloc_bytes`?  Is `min_alloc_bytes=0` expected to work?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -33,17 +33,17 @@ DynamicPool::DynamicPool(const std::string& name,
...
   m_align_bytes{align_bytes},
   m_coalesce_heuristic{coalesce_heuristic},
   m_used_map{},
   m_free_map{},
   m_curr_bytes{0},
-  m_actual_bytes{initial_alloc_bytes},
+  m_actual_bytes{round_up(initial_alloc_bytes, align_bytes)},
```
mcfadden8: Where do we validate `initial_alloc_bytes`? Is `initial_alloc_bytes=0` expected to work?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -33,17 +33,17 @@ DynamicPool::DynamicPool(const std::string& name,
...
   m_highwatermark{0}
 {
-  insertFree(m_allocator->allocate(initial_alloc_bytes),
-             initial_alloc_bytes, true, initial_alloc_bytes);
+  const std::size_t bytes{round_up(initial_alloc_bytes, align_bytes)};
```
mcfadden8: We need tests for edge-conditions of allocation sizes so that byte counts of 1 through `align_bytes-1` are tried.  I'm not sure if we have tested how well `round_up` works.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -33,17 +33,17 @@ DynamicPool::DynamicPool(const std::string& name,
...
+  insertFree(m_allocator->allocate(bytes), bytes, true, bytes);
```
mcfadden8:
```suggestion
  insertFree(m_allocator->allocate(rounded_to_alignment_bytes), rounded_to_alignment_bytes, true, rounded_to_alignment_bytes);
```

It does not need to be `rounded_to_alignment_bytes`, but I think that `bytes` by itself is not descriptive enough.

It could be `rounded_bytes` like it is in other parts of the code.

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -197,11 +196,10 @@ void* DynamicPool::allocate(std::size_t bytes)
     m_actual_bytes += alloc_bytes;
     m_curr_bytes += rounded_bytes;
 
-    const int64_t left_bytes{static_cast<int64_t>(
-        m_min_alloc_bytes - rounded_bytes)};
+    const std::size_t left_bytes{alloc_bytes - rounded_bytes};
```
mcfadden8: Without this change, aren't allocations greater than `m_min_alloc_bytes` problematic?  It seems that this would also cause problems.  Are we able to test this somehow?

src/umpire/strategy/DynamicPool.cpp
mcfadden8: yikes.  I'm not sure what the intent of the previous line was.

mcfadden8: I still could see `m_min_alloc_bytes - rounded_bytes` being something huge if `rounded_bytes > m_min_alloc_bytes`.

tests/integration/strategy_tests.cpp
```diff
@@ -389,11 +389,43 @@ TEST(ReleaseTest, Works)
   void* ptr_two = alloc.allocate(1024);
   alloc.deallocate(ptr_two);
 
+  ASSERT_EQ(alloc.getCurrentSize(), 62);
   EXPECT_NO_THROW(alloc.release());
 
+  ASSERT_LE(alloc.getActualSize(), 64);
+
```
mcfadden8: Why isn't this failing?  Should `alloc.getActualSize() == (1024+64)`?

tests/integration/strategy_tests.cpp
```diff
@@ -389,11 +389,43 @@ TEST(ReleaseTest, Works)
...
+  ASSERT_EQ(allocator.getCurrentSize(), 0);
+  ASSERT_GE(allocator.getActualSize(), 0);
+
+  allocator.release();
+
+  ASSERT_EQ(allocator.getCurrentSize(), 0);
+  ASSERT_EQ(allocator.getActualSize(), 0);
+}
+
+
```
mcfadden8: As mentioned in the code above, I think we should add a `TEST(ReleaseTest, AlignmentTest)` that would set the alignment size and then check allocations from 1 to alignment_bytes and make sure the getCurrentSize()/getActualSize() return believable results.

RELEASE_NOTES.md
davidbeckingsale: Agree.

scripts/make_release_tarball.sh
davidbeckingsale: Not merged yet.

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: What does "valid" mean?

davidbeckingsale: No. And nowhere. I thought you only wanted a 1 line PR?

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: Okay.

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: Okay.

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: No. Allocations greater than `m_min_alloc_bytes` are exactly sized, so left_bytes would be > 0. In this case, left_bytes will now be 0. In any case, neither of these values would trigger the old `left_bytes > m_align_bytes` condition (which is correct).

tests/integration/strategy_tests.cpp
davidbeckingsale: No, `release` is called, and ptr_two has been deallocated.

scripts/make_release_tarball.sh
mcfadden8: Oh right.

src/umpire/strategy/DynamicPool.cpp
jdahm: Yes, I think we should UMPIRE_ERROR() if someone passes 0 as min_alloc_size, or align_bytes.

src/umpire/strategy/DynamicPool.cpp
mcfadden8: We make the rules for that since it is our API.  I will write a Jira ticket on this.  I'm no positive, but it it would not surprise me if round_up assumed align_bytes at least needed to be a power of 2.

mcfadden8: I just wanted to point it out here since I am now thinking about the API.  I wrote a Jira ticket and we can decide where to address this elsewhere.

mcfadden8: I wrote a Jira ticket.  We can decide when/if/where to address this later.

src/umpire/strategy/DynamicPool.cpp
mcfadden8: I wrote a Jira ticket to decide what the contract needs to be for all of these parameters.  We can decide when/if/where/how to address these separately.

src/umpire/strategy/DynamicPool.cpp
mcfadden8: um..  okay?

So, this caused a problem because the (evidently small) leak would cause the containing block to become hostage and would not ever be coallesced or released.  Is that right?

davidbeckingsale: Yes - then repeating this over many many many allocations caused the large amounts we were seeing.

src/umpire/strategy/DynamicPool.cpp
mcfadden8: @davidbeckingsale, were you going to round this too as suggested by @jdahm ?

src/umpire/strategy/DynamicPool.cpp
```diff
@@ -180,28 +180,27 @@ void* DynamicPool::allocate(std::size_t bytes)
...
 
-    if (left_bytes > m_align_bytes) {
+    if (left_bytes > 0) {
       insertFree(static_cast<unsigned char*>(ptr) + rounded_bytes, left_bytes,
                  false, whole_bytes);
     }
   } else {
     // Allocate new block -- note this does not check whether alignment is met
     const std::size_t alloc_bytes{std::max(rounded_bytes, m_min_alloc_bytes)};
+    std::cout << "Allocating " << alloc_bytes << " from resource for " << rounded_bytes << std::endl;
```
mcfadden8: I assume we want to remove this right?

tests/integration/strategy_tests.cpp
mcfadden8: Can you add this test?  I think that adding a test would be good for the patch release if it is not too difficult.

tests/integration/strategy_tests.cpp
mcfadden8: Oh duh.  I missed the release() call.  I suppose we could consider adding another ASSERT_EQ before the release for getActualSize()?

src/umpire/strategy/DynamicPool.cpp
davidbeckingsale: Done already.

mcfadden8: sorry, I don't no why my browser sometimes keeps old crap around

tests/integration/strategy_tests.cpp
davidbeckingsale: As discussed, we will test this elsewhere.

RELEASE_NOTES.md
davidbeckingsale: I have updated the description.

mcfadden8:
Thanks @davidbeckingsale, the changes look great!  I plan to test this a little more and will approve afterwards.

PR #234, opened 2019-09-05
Restore pool behavior where blocks are merged more often

davidbeckingsale:


mcfadden8:
@davidbeckingsale, I had to create new "good" replay logs for replay_test since behavior changes from previously and coalesce is called more frequently now that we are merging blocks at each deallocation.  The default heuristic is set to 100% which means a coalesce will be done when 100% of the bytes in the pool are freeable (none are allocated).

Thus far, I have not been able to update the replay logs for HIP yet as I am no longer able to get builds to work on crayz (is there someplace else I should try?)

Another behavior change that should be checked is for `release()`.  When `release()` is called, the `initial`  block size that was allocated during construction is also returned to the device.  I'm not 100% certain, but I believe that the next allocate will cause the first block that gets allocated to be of 'minimum block size` which does change the characteristic of the pool and we need to decide whether that is desired behavior or not as it is different from 0.3.5.

davidbeckingsale:
This looks good. Let's try and regenerate a hip replay together in person tomorrow.

davidbeckingsale:
This is good to go.

PR #236, opened 2019-09-05
Fix getActualSize for all Allocators

davidbeckingsale:


tests/integration/allocator_integration_tests.cpp
```diff
@@ -130,6 +130,28 @@ TEST_P(AllocatorTest, get_allocator_records)
   m_allocator->deallocate(data);
 }
 
+TEST_P(AllocatorTest, getCurrentSize)
```
mcfadden8: Does this capture all allocators and strategies?

davidbeckingsale: It captures all basic allocators. We don't have tests for all strategies yet, but the DynamicPool is already tested.

PR #237, opened 2019-09-06
Add symbol to prevent linking across major versions

davidbeckingsale:


davidbeckingsale:
This is ready to go @mcfadden8 - I renamed the symbol to be `umpire_ver_<major>_found`, so that the linker error will be a bit more descriptive.

src/umpire/Allocator.inl
```diff
@@ -0,0 +1,57 @@
...
+    UMPIRE_LOG(Info, "Deallocating a null pointer");
+    return;
+  } else {
+    m_allocator->deallocate(ptr);
+  }
+}
+
+} // end of namespace umpire
+
+#endif // UMPIRE_Allocator_INL
```
mcfadden8: I assume that creating these as inline functions definitely is an API change even if it is simply a relocation.  Does this mean that we need to change the major version number of the API?

OTOH, I suppose binaries built using the previous version of the header file will fail to link because they are expecting to find `allocate` and `deallocate` in the library rather than inline.

src/umpire/util/FixedMallocPool.cpp
```diff
@@ -15,6 +15,7 @@
 #include <cstring>
 #include <cstdlib>
 #include <algorithm>
+#include <cstddef>
```
mcfadden8: It appears that this was the only change to this file.  Was there a side-effect change in a different file that caused this one to need to change?

src/umpire/util/FixedMallocPool.hpp
```diff
@@ -7,8 +7,6 @@
 #ifndef UMPIRE_FixedMallocPool_HPP
 #define UMPIRE_FixedMallocPool_HPP
 
-#include "umpire/Allocator.hpp"
```
mcfadden8: Why isn't this needed?

davidbeckingsale: It doesn't use an Allocator.

src/umpire/util/FixedMallocPool.cpp
davidbeckingsale: Needs `std::size_t` - removing the `Allocator.hpp` file below meant this is now required.

src/umpire/Allocator.inl
davidbeckingsale: Agree. I'm not sure the best thing to do. We have to make these (or some function) inline to ensure that the symbol gets referenced. We can discuss this in person today.

davidbeckingsale:
Agree!

PR #239, opened 2019-09-06
Add more functions to C/Fortran API

davidbeckingsale:


src/umpire/interface/c_fortran/typesUmpire.h
```diff
@@ -26,17 +28,35 @@ struct s_umpire_allocator {
...
 
-struct s_umpire_resourcemanager {
+struct s_umpire_strategy_allocationadvisor {
     void *addr;     /* address of C++ memory */
     int idtor;      /* index of destructor */
 };
-typedef struct s_umpire_resourcemanager umpire_resourcemanager;
+typedef struct s_umpire_strategy_allocationadvisor umpire_strategy_allocationadvisor;
+
+struct s_umpire_strategy_dynamicpool {
```
mcfadden8: While here, should we also create an interface to `DynamicPoolList`?  I think we should.

davidbeckingsale:
@mcfadden8 - added the list pool, and tests. I think this is ready.

cmake/SetupCompilerFlags.cmake
```diff
@@ -4,6 +4,9 @@
 #
 # SPDX-License-Identifier: (MIT)
 ##############################################################################
+message(STATUS "Setting C standard to 99")
+set(CMAKE_C_STANDARD 99)
```
ax3l: If you write
```C++
if(NOT DEFINED CMAKE_C_STANDARD)
    message(STATUS "Setting C standard to 99") 
    set(CMAKE_C_STANDARD 99) 
endif()
```
users can overwrite this setting with their own value from the command-line, e.g. to build for C11.

davidbeckingsale: See #266

PR #241, opened 2019-09-11
Added option to perform operations as they are interpreted

mcfadden8:


tools/replay.cpp
```diff
@@ -34,6 +34,9 @@ int main(int argc, char* argv[])
      , "Input file created by Umpire library with UMPIRE_REPLAY=On"
      , cxxopts::value<std::string>(), "FILE"
     )
+    (  "l, largefile"
```
davidbeckingsale: I don't think this is a good name for the option - it should be more descriptive of what the option does, not _why_ it does it.

tools/util/ReplayInterpreter.hpp
```diff
@@ -34,6 +34,7 @@ class ReplayInterpreter {
     using AllocatorIndexMap = std::unordered_map<AllocatorFromLog, AllocatorIndex>;
     using AllocationAllocatorMap = std::unordered_map<AllocationFromLog, AllocatorIndex>;
 
+    ReplayOperationManager* m_operation_mgr;
```
davidbeckingsale: Why is this a pointer?

mcfadden8:
Closing PR.  This needs to be resolved in a different way.  More specifically, REPLAY needs to emit and consume binary data.

PR #243, opened 2019-09-13
Feature/lassen ci

adrienbernede:
This PR aims at moving GitLab jobs from butte to lassen:
- It is still possible to trigger butte jobs.
- It is now possible to exclude lassen and quartz by environment variable.


adrienbernede:
If you merge this one, pipelines will pass ! ;)

.gitlab/ci/build_lassen.yml
```diff
@@ -0,0 +1,100 @@
...
+  tags:
+    - shell
+    - lassen
+  stage: build
+  extends: .build_blueos_3_ppc64le_ib_p9_script
+  except:
+    refs:
+      - /_lnone/
+    variables:
+      - $UMPIRE_CI_LASSEN == "OFF"
```
davidbeckingsale: Why is this variable set to `"OFF"`? I don't quite understand how these variables work.

adrienbernede: This is in fact a test for the except exclusion to apply.
It reads this like:
Jobs using `.build_lassen` property will be run `except` if _either_ ref contains `_lnone` _or_ `$UMPIRE_CI_LASSEN == "OFF")`.

.gitlab/ci/build_butte.yml
```diff
@@ -14,9 +14,10 @@
     - shell
     - butte
   stage: build
-  extends: .build_butte_script
-  except:
-    - /_bnone/
+  extends: .build_blueos_3_ppc64le_ib_script
+  only:
+    variables:
+      - $UMPIRE_CI_BUTTE == "ON"
```
adrienbernede: In butte case, we want the jobs to be deactivated by default. So the only condition applies.
It reads like:
Jobs using `.build_butte` property will be run `only` _if_ `$UMPIRE_CI_BUTTE == "ON"`.

adrienbernede: Note: major difference between only and except:
 - In `only:`, all categories (refs, variables) must have at least one condition fulfilled.
 - In `except:`, one fulfilled condition is enough to prevent the job from running.

https://docs.gitlab.com/ee/ci/yaml/README.html#onlyexcept-advanced

adrienbernede:
@mcfadden8 please let me know if you are OK (or not) with the changelog.md update.

adrienbernede:
I also solved the conflict with develop in changelog.

.gitlab/ci/build_butte.yml
davidbeckingsale: Okay I understand. I was reading the statement as a variable assignment, not as a condition. Thanks for clarifying.

PR #247, opened 2019-09-18
Resetting CMAKE_CXX_EXTENSIONS to ON for PGI compiler.

mcfadden8:
The current version of BLT is setting CMAKE_CXX_EXTENSIONS to OFF for all compilers.  This causes the `-A` option to be added to PGI compiler options which breaks builds for Umpire.

This workaround will reset the CMAKE_CXX_EXTENSIONS flag to ON for PGI compilers.


mcfadden8:
The problem is that the actual binaries produced do not run!

```
$ ./bin/tools/replay --help
terminate called after throwing an instance of 'std::regex_error'
  what():  regex_error
[1]    148158 abort (core dumped)  ./bin/tools/replay --help
```

davidbeckingsale:
Woohoo! 👍

PR #250, opened 2019-09-19
Moved (de)allocate implementation back into library where it belongs

mcfadden8:


davidbeckingsale:
Don't forget we need to merge this fix to develop also.

mcfadden8:
Closing as this turns out to be an nvcc problem (2720046)

PR #258, opened 2019-10-03
Add operations to "unset" memory advice

### LINKED ISSUE #251: AllocationAdvisor needs to revoke advice in deallocate if using pooled allocator
**Describe the bug**

If using a pooled allocator AllocationAdvisor will leave advice for pages that were "freed".

**To Reproduce**

N/A

**Expected behavior**

AllocationAdvisor will revoke advice for buffers deallocated by a pooled allocator.

davidbeckingsale:
Fixes #251


davidbeckingsale:
@mcfadden8 - this is ready to go.

src/umpire/ResourceManager.cpp
```diff
@@ -709,4 +710,18 @@ ResourceManager::getZeroByteAllocator()
   return m_allocators_by_name[s_zero_byte_pool_name];
 }
 
+std::shared_ptr<op::MemoryOperation>
+ResourceManager::getOperation(
+    const std::string& operation_name,
+    Allocator src_allocator,
```
mcfadden8: Will the following suggestion work?

```suggestion
    const Allocator& src_allocator,
```

src/umpire/ResourceManager.cpp
```diff
@@ -709,4 +710,18 @@ ResourceManager::getZeroByteAllocator()
...
+    Allocator dst_allocator)
```
mcfadden8: Same question here:
```suggestion
    const Allocator& dst_allocator)
```

src/umpire/ResourceManager.cpp
davidbeckingsale: Should do, let me try 😄

PR #259, opened 2019-10-03
Update json tpl to v3.7.0 

davidbeckingsale:


mcfadden8:
Is there a driver for this?

davidbeckingsale:
I thought it would be worthwhile to try and stay up to date, particularly getting any fixes from the upstream project.

mcfadden8:
I'm not sure about this.  I believe in the 'if it ain't broke...' saying...  If we can point to a particular fix that is needed by replay or our build environment, that might convince me.

PR #264, opened 2019-10-14
Update exit code for branch deletion

davidbeckingsale:


.github/workflows/pull_request.yml
```diff
@@ -10,3 +10,4 @@ jobs:
       uses: jessfraz/branch-cleanup-action@b3a2b299e1ce42dbcbef5c4a0b9e97b8068154b8
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        NO_BRANCH_DELETED_EXIT_CODE: "0"
```
mcfadden8: How does the branch ever get deleted? (which line actually causes the removal to happen)?

davidbeckingsale: It's in the action `jessfrazz/branch-cleanup-action`. That action contains a script to delete branches, that uses the environment variable I have just added.

I'm hoping this might fix the issue.

PR #267, opened 2019-10-18
Moved source for replay to replay sub-directory of tools

mcfadden8:
This is simply to better organize the `replay` and `replaydiff` programs.  Under the `tools` directory, I made the following organizational changes:

1. I renamed `util/` to `replay/`
2. I moved `replay.cpp` and `replaydiff.cpp` to `replay` directory

This is in preparation for more significant changes to the replay implementation and I wanted to do this as a separate step so that future merge work will be less of a headache.


mcfadden8:
@davidbeckingsale, I am assuming that I do not need to update the CHANGELOG for this since this is an internal source code directory organizational change that does not have external impact.

I think this is ready for review and I am ready to merge as soon as it has been approved.

mcfadden8:
@davidbeckingsale, would you review one more time please?  I decided to forgo implementing replay as a library until a later phase since it is not related to the speedup work I am currently working on.

PR #277, opened 2019-11-04
Add replay entries for copy, memset, reallocate, move

davidbeckingsale:
Example of output for "copy":

```
{ "kind":"replay", "uid":46718, "timestamp":1573159858540844000,  "event": "copy", "payload": { "src": "0x7fd9a7801028", src_offset: "40", "dest": "0x7fd9a780002c", dst_offset: "44",  "size": 4, "src_allocator_ref": "0x7fd9a5c02c60", "dst_allocator_ref": "0x7fd9a5c02c60" }  }
```


src/umpire/ResourceManager.cpp
```diff
@@ -482,6 +482,20 @@ void ResourceManager::copy(void* dst_ptr, void* src_ptr, std::size_t size)
     size = src_size;
   }
 
+  UMPIRE_REPLAY( 
+      R"( "event": "copy", "payload": { "src": )"
```
mcfadden8: What is `R`?

PR #278, opened 2019-11-06
Replay now creates index of replays

mcfadden8:
Like python, generate a an compiled version of the text replay that is indexed for faster replays.


davidbeckingsale:
🎉

PR #279, opened 2019-11-11
Splitting builds and tests

adrienbernede:
In this PR, I "simply" split builds and tests for quartz builds.

This requires building in predefined directories (not specific to each
runner).

Also, we are not using artifacts, so there is no dependency
relationship in that sense between build and test jobs.
However, GitLab introduced a new dependency relationship between jobs,
allowing to start a job in a later stage if all its dependencies have
run. (see "needs:" keyword)

Prevent the job from cloning when not necessary, setting GIT_STRATEGY
to "none".


adrienbernede:
GitLab Pipeline failed probably because of some delay causing the "release resources" to occur after the job was terminated "due to time limit".
That is to say it is not related to my feature (passed in pre-PR tests).

PR #284, opened 2019-11-23
UM-538 update HIP good replay results for NUMA

mcfadden8:


mcfadden8:
@davidbeckingsale, this is ready for review (tested on corona).

PR #291, opened 2019-12-17
Feature/um 541 provide recipe for creating a thread safe allocator

mcfadden8:


mcfadden8:
@davidbeckingsale, I think this one is ready for review as well.

PR #292, opened 2019-12-23
Handle reallocate case where size == 0

mcfadden8:


tests/integration/operation_tests.cpp
```diff
@@ -364,6 +364,135 @@ TEST_P(ReallocateTest, RealocateNull)
...
+
+  // valid ptr, zero size
+  reallocated_array =
+    static_cast<float*>(
+      rm.reallocate(   static_cast<void*>(reallocated_array)
+                    ,  reallocated_zero_size*sizeof(float)));
+
+  ASSERT_EQ(   source_allocator->getSize(reallocated_array)
+             , reallocated_zero_size*sizeof(float));
+
```
kennyweiss: Thanks @mcfadden8.

It looks like this will fix the case where one attempts to call ``reallocate(0)`` on an array with an existing allocation (containing a non-zero number of elements or a ``nullptr``).

I think there is still an issue if you attempt to call ``reallocate(N)`` (w/ ``N>0``) on an array whose current size is 0.

E.g. if I add the following test here, right after calling ``reallocate(0)``

```cpp
  // valid ptr, non-zero size after zero size
  reallocated_array =
    static_cast<float*>(
      rm.reallocate(   static_cast<void*>(reallocated_array)
                    ,  reallocated_size_1*sizeof(float)));

  ASSERT_EQ(   source_allocator->getSize(reallocated_array)
             , reallocated_size_1*sizeof(float));
```

I get the following error:
```
>ctest -R operation_tests -VV
...
18: Test command: <Umpire>/build-umpire/bin/operation_tests
18: [==========] Running 18 tests from 5 test cases.
18: [ RUN      ] Reallocate/ReallocateTest.ReallocateNullZero/0
*** Error in `<Umpire>/build-umpire/bin/operation_tests': realloc(): invalid pointer: 0x00002aaaaaad1000 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7f804)[0x2aaaab78d804]
/lib64/libc.so.6(realloc+0x389)[0x2aaaab794359]
<Umpire>/build-umpire/bin/operation_tests[0x455715]
...
======= Memory map: ========
...
1/1 Test #18: operation_tests ..................***Exception: Other  0.01 sec
```

kennyweiss: For reference, here is what I got from ``std::realloc`` using gcc@4.9:

```cpp
#include <cstdlib>
#include <new>
#include <cassert>
#include <iostream>
 

void printArray(int* arr, int sz)
{
    std::cout<<"Array: "
             <<"{ address:" << &arr
             <<", ptr val:" << arr << "} -- ";
    
    for(int i=0; i< sz; ++i)    { std::cout << arr[i] << " ";   }
    std::cout << "\n" << std::endl;
}

int main()
{
    std::cout<<"*** Allocating arr -- 5 ***" << std::endl;
    int* arr = static_cast<int*>( std::malloc(5 * sizeof(int)) );
    for(int i=0; i<5; ++i)    { arr[i] = i;   }
    printArray(arr, 5);
    
    std::cout<<"*** Reallocating arr -- 10 ***" << std::endl;
    arr = static_cast<int*>( std::realloc(arr, 10 * sizeof(int)) );     // <-- realloc(10)
    for(int i=0; i<10; ++i)    { arr[i] = 10 + i;   }
    printArray(arr, 10);
    
    std::cout<<"*** Reallocating arr -- 0 ***" << std::endl;
    arr = static_cast<int*>( std::realloc(arr, 0) );                    // <-- realloc(0)
    printArray(arr, 0);

    // Allocate a different array before reallocating arr 
    // to see if heap address of `arr` will change after realloc(0)
    int* arr2 = static_cast<int*>( std::malloc(5 * sizeof(int)) );
    for(int i=0; i<5; ++i)    { arr2[i] = 1000 + i;   }

    std::cout<<"*** Reallocating arr -- 10 ***" << std::endl;
    arr = static_cast<int*>( std::realloc(arr, 10 * sizeof(int)) );     // <-- realloc(10)
    for(int i=0; i<10; ++i)    { arr[i] = 20 + i;   }
    printArray(arr, 10);

    std::free(arr);
    std::free(arr2);
}
```

In case it is not clear, the sequence is:
* ``arr = malloc(5)``
* ``arr = realloc(arr,10)``
* ``arr = realloc(arr,0)``
* ``arr2 = malloc(5)``
* ``arr = realloc(arr,10)``

The call to ``arr2`` was added to see what happens if you allocate something else between ``realloc(arr,0)`` and ``realloc(arr,10)``

Output:
```
*** Allocating arr -- 5 ***
Array: { address:0x7fff6a0c0eb8, ptr val:0xbf2c30} -- 0 1 2 3 4 

*** Reallocating arr -- 10 ***
Array: { address:0x7fff6a0c0eb8, ptr val:0xbf2c30} -- 10 11 12 13 14 15 16 17 18 19 

*** Reallocating arr -- 0 ***
Array: { address:0x7fff6a0c0eb8, ptr val:0} -- 

*** Reallocating arr -- 10 ***
Array: { address:0x7fff6a0c0eb8, ptr val:0xbf2c30} -- 20 21 22 23 24 25 26 27 28 29 
```

Note that ``arr``'s memory is in the same location after ``realloc(0)`` and ``realloc(10)``.

mcfadden8: Thanks!
I seem to have missed this case.  Will investigate and fix this today.

src/umpire/ResourceManager.cpp
```diff
@@ -565,7 +565,10 @@ ResourceManager::reallocate(void* src_ptr, std::size_t size)
...
   void* dst_ptr = nullptr;
 
-  if (!src_ptr) {
+  if (!src_ptr || size == 0) {
+    if (src_ptr)
+      m_default_allocator->deallocate(src_ptr);
+
     dst_ptr = m_default_allocator->allocate(size);
   } else {
     auto& op_registry = op::MemoryOperationRegistry::getInstance();
```
kennyweiss: Thanks @mcfadden8!

I think there is still an issue with this ``else`` branch.

* Line 566 sets ``dest_ptr = nullptr``.
* Line 587 dereferences this null pointer

```cpp
   void* dst_ptr = nullptr; 
   ...
    op->transform(src_ptr, &dst_ptr, alloc_record, alloc_record, size);
```

Note: This is the code path that my function in axom is taking.

tests/integration/operation_tests.cpp
mcfadden8: This problem was caused by the HOST specialization of `operation` was blindly calling `::realloc` with the pointer to the zero-length space.  Umpire has its own semantics for zero length allocations and the pointer passed to `::realloc` caused the crash.  Since Umpire maintains its own semantics for pointers to zero-length objects, we needed to check for this inside of the HOST specialization of the `op-operation` for reallocate.

Thanks for the additional test.  It has also been added to the test program.

mcfadden8: As discussed in person, I will go ahead and resolve this conversation as it is working now.

src/umpire/ResourceManager.cpp
mcfadden8: As discussed in person, I will go ahead and resolve this conversation as it is working now.

mcfadden8:
@davidbeckingsale, this is ready for review.

src/umpire/ResourceManager.cpp
kennyweiss: Thanks for the updates @mcfadden8 -- I verified that your changes work in my application.

PR #294, opened 2020-01-09
Pull in radiuss-ci configurations & scripts as a submodule

davidbeckingsale:


davidbeckingsale:
@mcfadden8 - this switches us to use the radiuss-ci repository for our CI configurations (so that things can be shared between multiple projects)

mcfadden8:
This is timely but means that it is going to be painful for me to add support for new compilers.  Is this supposed to replace what is also in the scripts/bamboo directory?

davidbeckingsale:
New compilers would need to be added here: https://github.com/LLNL/radiuss-ci

The `scripts/bamboo` location is for project-specific options. 

I can see why this might be a bit painful, but the idea is that the radiuss-ci repo will contain host-configs for all compilers we care about, hopefully maintained by someone else (@adrienbernede)

adrienbernede:
> New compilers would need to be added here: https://github.com/LLNL/radiuss-ci
> 
> The `scripts/bamboo` location is for project-specific options.
> 
> I can see why this might be a bit painful, but the idea is that the radiuss-ci repo will contain host-configs for all compilers we care about, hopefully maintained by someone else (@adrienbernede)

Well, let’s say that hopefully this gets maintained _in only one open location_. So anyone can contribute for the benefit of any project relying on this mechanism.

adrienbernede:
@davidbeckingsale I intend to [merge my changes in radiuss-ci](https://github.com/LLNL/radiuss-ci/pull/2). Then I would change the submodule to point on master branch (in umpire).

.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "src/umpire/tpl/conduit"]
 	path = src/umpire/tpl/conduit
 	url = https://github.com/LLNL/conduit.git
+[submodule ".gitlab/radiuss-ci"]
```
davidbeckingsale: Could we put the submodule in the directory `.radiuss-ci` instead of `.gitlab/radiuss-ci` - it will get rid of the extra `gitlab` in the path, and also make it more obvious that the shared configurations can be used for more than just gitlab.

adrienbernede: OK. `.gitlab` directory will still exists right, though, confirm?

adrienbernede:
> Can you please verify that we are not missing any configs - @mcfadden8 recently added a few.

I’m on it. I see you merged master into it, It answers my question "can I rebase" :)

adrienbernede:
Not necessarily in fact. I would prefer to rebase then.

adrienbernede:
Updated with new compilers (including ci)
Nice and clean history

.gitmodules
davidbeckingsale: I think the .gitlab directory can be removed now that we have all the compiler configurations in the radiuss-ci repo.

adrienbernede: .gitlab also contains the "ci" directory, with the pipeline description for LC machines.
These files will be partly generated in the near future (waiting for a new feature in Gitlab)

adrienbernede: So, as soon as they are partly generated, the remaining stuff could be reintegrated into .gitlab-ci.yml

adrienbernede: Except if you want to anticipate and have a large .gitlab-ci.yml in the meantime

davidbeckingsale: Ah, I understand. Let's keep the `.gitlab` directory, and move the RADIUSS submodule to `./.radiuss-ci`

adrienbernede:
LGTM

davidbeckingsale:
This is looking good, thanks @adrienbernede 👍

PR #295, opened 2020-01-09
Added CI support BLUEOS latest xl, gcc, and clang compilers

mcfadden8:
Added CI of Umpire plus Cuda, Numa, C, and Fortran with the following compilers on BLUEOS:

- (Upstream) Clang 9.0.0
- Clang (Default, presently clang/ibm-2019.10.03)
- GCC (Default, presently gcc/4.9.3)
- GCC 8.3.1
- XL (Default, presently xl/xl/2019.08.20)

The compilers labeled with "Default" will follow whatever the default of `module load compler` yields so it will change automatically when the default is updated.


mcfadden8:
@davidbeckingsale, this is ready for review.

adrienbernede:
@mcfadden8 @davidbeckingsale,

### Concerns
 - Each default is most likely a duplicate of another configuration.
 - These default host_config files are not compatible with #294.

### Alternate solution
I tend to think that it would be better to simply maintain a table specifying the default:

[toolchain_defaults.txt]
gcc gcc_4_9_3
xl xl_2019_08_20
...

This table could be directly embedded in script/gitlab/build_and_test.sh, which can conveniently be used to build umpire on LC:

`COMPILER=GCC_4_9_3 scripts/gitlab/build_and_test.sh`

It would then be easy to treat the case

`COMPILER=gcc_default scripts/gitlab/build_and_test.sh`

or even

`COMPILER=gcc scripts/gitlab/build_and_test.sh`

mcfadden8:
While I agree that the default will likely be a duplicate of another specific configuration, it does allow for CI to instantly match whenever the default compiler is changed.  This will allow us to quickly determine whether there is a problem with new defaults as they come into existence.  I don't like the idea of having to maintain a separate table that has to be updated each time the compiler defaults are changed.

adrienbernede:
I now understand what your need is. I thought you were talking about Umpire defaults, but I should have read more carefully.

Then I guess you expect the default config to be maintained in radiuss-ci. And maybe in an automated manner, since you would like to be sure to follow the changes of "module load".

This idea has its strengths. I am still not 100% convinced, because it seems to me that you could determine the default on the fly, precisely using "module" commands. But maybe that’s not perfect neither.

PR #297, opened 2020-01-13
Use macro rather than constexpr for version number

davidbeckingsale:
cc @gzagaris


gzagaris:
@davidbeckingsale, Thanks!

PR #303, opened 2020-01-16
Map "UM" to device memory when using HIP

davidbeckingsale:
Device memory can be accessed from the host, so can fulfill the role of unified memory for many use cases. An alternative mapping would be to use pinned host memory, but this would result in poor performance when accessing data on the device.


davidbeckingsale:
@mcfadden8 - can I get a review on this?

PR #304, opened 2020-01-23
Fixed corner case where 0-byte allocation was reallocated to another …

mcfadden8:
…0-byte allocation

### DIFF
--- a/src/umpire/ResourceManager.cpp
+++ b/src/umpire/ResourceManager.cpp
@@ -540,75 +540,73 @@ void ResourceManager::memset(void* ptr, int value, std::size_t length)
 void*
 ResourceManager::reallocate(void* current_ptr, std::size_t new_size)
 {
-  UMPIRE_LOG(Debug, "(current_ptr=" << current_ptr << ", new_size=" << new_size << ")");
-
-  UMPIRE_REPLAY(
-      R"( "event": "reallocate", "payload": { "ptr": ")"
-      << current_ptr
-      << R"(", "size": )"
-      << new_size
-      << R"( })"
-  );
-
-  void* new_ptr;
-
-  if ( current_ptr == nullptr || new_size == 0 ) {
-    if (current_ptr != nullptr )
-      m_default_allocator->deallocate(current_ptr);
-
-    new_ptr = m_default_allocator->allocate(new_size);
-  } else {
-    auto& op_registry = op::MemoryOperationRegistry::getInstance();
-
-    auto alloc_record = m_allocations.find(current_ptr);
-
-    if (current_ptr != alloc_record->ptr) {
-      UMPIRE_ERROR("Cannot reallocate an offset ptr (ptr="
-          << current_ptr << ", base=" << alloc_record->ptr);
-    }
-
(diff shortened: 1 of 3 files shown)


src/umpire/ResourceManager.cpp
```diff
@@ -540,76 +540,74 @@ void ResourceManager::memset(void* ptr, int value, std::size_t length)
...
+  //
+  if ( current_ptr == nullptr ) {
     new_ptr = allocator.allocate(new_size);
-  } else {
+  }
+  else {
     auto alloc_record = m_allocations.find(current_ptr);
+    auto alloc = Allocator(alloc_record->strategy);
 
+#ifdef TAKEN_OUT_AND_NEED_TO_DISCUSS
```
mcfadden8: @davidbeckingsale, this is the check that we need to discuss.  The previous API would perform this check to make sure that that the caller didn't try to reallocate with the wrong `Allocator`

src/umpire/ResourceManager.cpp
```diff
@@ -540,76 +540,74 @@ void ResourceManager::memset(void* ptr, int value, std::size_t length)
...
-
-    auto op = op_registry.find("REALLOCATE",
-        alloc_record->strategy,
-        alloc_record->strategy);
-
-    op->transform(current_ptr, &new_ptr, alloc_record, alloc_record, new_size);
-  }
-
-  return new_ptr;
+  return reallocate(current_ptr, new_size, Allocator(m_default_allocator));
```
mcfadden8: @davidbeckingsale This particular change simplified this method by having it call the `reallocate` with the `m_default_allocator`.  While the code de-duplication was good, it did break a check that was being made within the called function.  See my comment below.

src/umpire/ResourceManager.cpp
davidbeckingsale: Can the check be moved up, and guarded if the current ptr is null?

mcfadden8: Yes.  Just a sec

mcfadden8:
@davidbeckingsale, I think this should be ready for revew now.

mcfadden8:
I will finish fixing the replay problems in a separate branch.

src/umpire/ResourceManager.cpp
```diff
@@ -540,76 +540,82 @@ void ResourceManager::memset(void* ptr, int value, std::size_t length)
...
-  void* new_ptr;
-
-  if ( current_ptr == nullptr || new_size == 0 ) {
-    if (current_ptr != nullptr )
-      m_default_allocator->deallocate(current_ptr);
-
-    new_ptr = m_default_allocator->allocate(new_size);
-  } else {
-    auto& op_registry = op::MemoryOperationRegistry::getInstance();
+  Allocator allocator;
```
davidbeckingsale: This could be a strategy pointer rather than the object, and then you can just construct the object inline when you call reallocate?

Having uninitialized variables is bad.

mcfadden8: agreed.  that was ugly..  fixed

PR #305, opened 2020-01-24
Add support for allocation on OpenMP devices

davidbeckingsale:


davidbeckingsale:
It provides 'DEVICE' when compiling with `ENABLE_OPENMP_TARGET=On`, so gets tested the same as CUDA or HIP.

mcfadden8:
> This looks good. I think you should add a replay test for the OpenMP resource to make sure that the newly added resource creation is understood and tested.

9c6a98c addresses my concern.  Thanks!

davidbeckingsale:
@mcfadden8 - now the Bamboo build is passing, this is ready.

PR #311, opened 2020-02-21
Add sanitizers to DynamicPoolMap and DynamicPoolList

davidbeckingsale:


src/umpire/util/memory_sanitizers.hpp
```diff
@@ -0,0 +1,30 @@
+//////////////////////////////////////////////////////////////////////////////
+// Copyright (c) 2016-20, Lawrence Livermore National Security, LLC and Umpire
+// project contributors. See the COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (MIT)
+//////////////////////////////////////////////////////////////////////////////
+#ifndef UMPIRE_memory_sanitizers_HPP
+#define UMPIRE_memory_sanitizers_HPP
+
+#if defined(__clang__)
```
mcfadden8: Don't we want to make this optionally be compiled in?  I thought I saw something about a 73% slowdown?

davidbeckingsale: The ASAN macros themselves are no-ops when compiled without `-fsanitize=address`

mcfadden8:
This looks great!  Would you add tests for this?  I'd like to see what it does for both stores and loads if possible.

davidbeckingsale:
How do you think it would best be tested?

src/umpire/util/memory_sanitizers.hpp
davidbeckingsale: I will add an option to guard it also, since it adds a conditional for the `Platform` check.

mcfadden8:
With a signal handler?

PR #318, opened 2020-02-28
Documenting Gitlab CI Pipeline

adrienbernede:
Currently, Umpire can use Gitlab CI to test builds on LC systems. But maybe this is not documented enough. so I introduced more descriptive comments in gitlab ci yml files.

Now, I don’t know if I should go further than that, and how.

PS: Gitlab CI now needs to be triggered manually with a sign-off comment "LGTM". It runs automatically on master though.


adrienbernede:
LGTM

PR #323, opened 2020-03-12
Windows build fixes - mostly shared library definitions.

robinson96:
This update gets most of our win32 related changes into umpire (with the exception of a VS build system, which I'm assuming you don't actually want - but let me know if you do).


src/umpire/Allocator.inl
```diff
@@ -11,45 +11,21 @@
 
 #include "umpire/Allocator.hpp"
 
-#include "umpire/util/Macros.hpp"
-#include "umpire/Replay.hpp"
-
 namespace umpire {
 
-inline void*
```
davidbeckingsale: Does inline have to be removed? We need the symbol `umpire_ver_2_found` to show up in user code for our mismatch detection to work.

robinson96: I'd be surprised if it did (it could have been Al trying to circumvent the mismatch detection before he figured out that he needed to change our windows specific config file)

davidbeckingsale:
@robinson96  - this PR breaks a whole load of builds (including our builds on LC machines). It looks like what you want is a shared windows build? I can work on this locally.

robinson96:
OK, I have the CMake definition of the export macros in the right place now. I'll leave it to you to prettify it. Now, on to tackling CHAI divergence!

robinson96:
Approved PR 334 instead of this one.

Issue #321, opened 2020-03-10
Support compiling with clang and -stdlib=libc++


brunner6:
**Describe the bug**

There is some code to demangle C++ names in tools/replay/ReplayInterpreter.cpp that relies on a non-standard GCC extension, `abi::__cxa_demangle`.  When compiling with clang and using the -stdlib=libc++ option, which uses LLVM's implementation of the standard library, this function is not available.

**To Reproduce**

Compile the code with
```
module load clang/9.0.0
cmake  ../Umpire -DCMAKE_BUILD_TYPE=Release  -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS="-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0" -DCMAKE_C_FLAGS="-DGTEST_HAS_CXXABI_H_=0" -DENABLE_CUDA=OFF -DENABLE_FORTRAN=OFF -DENABLE_C=ON  -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=TRUE -DENABLE_GTEST=TRUE -DENABLE_GMOCK=TRUE -DENABLE_FRUIT=FALSE -DENABLE_EXAMPLES=TRUE -DENABLE_BENCHMARKS=TRUE -DCMAKE_VERBOSE_MAKEFILE=TRUE -DBLT_SOURCE_DIR=../blt -DENABLE_TOOLS=ON
```

**Expected behavior**

The code compiles with clang and libc++

**Compilers & Libraries (please complete the following information):**
 - clang (any version) with the -stdlib=libcxx flag
 - you also need to Define `-DGTEST_HAS_CXXABI_H_=0` in the C/C++ flags to get around the same exact issue in google test.

**Additional context**

This compiler set/option is one that KULL uses regularly to test.

I have a branch with with this patch on it, but I can't push to the repo:
```
diff --git a/tools/replay/ReplayInterpreter.cpp b/tools/replay/ReplayInterpreter.cpp
index d0e5859b..b8a9e383 100644
--- a/tools/replay/ReplayInterpreter.cpp
+++ b/tools/replay/ReplayInterpreter.cpp
@@ -16,7 +16,8 @@
 #include "ReplayFile.hpp"
 #include "umpire/tpl/json/json.hpp"
 
-#if !defined(_MSC_VER)
+#include <ciso646>
[...]

PR #324, opened 2020-03-18
Feature/backtrace release

mcfadden8:
- Backtrace information is now included in Umpire Error exception string
- "-DENABLE_ALLOCATION_BACKTRACE" added to cause backtraces to be kept for each active Umpire allocation.
- Resource Manager destructor now logs an error message (if logging enabled) for all leaked allocations found.  This information will also include the backtrace information if it is enabled.
- Added a guard around the demangling code so that `-stdlib=libc++` will build.  This fixes issue #321
- Added `-stdlib=libc++` CI to BLUEOS bamboo


davidbeckingsale:
@mcfadden8 - the Windows build is failing.

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 - Add support for multiple CUDA devices. These devices are detected and
   registered as "DEVICE_N", where N is the device number.
 
+- Allocation backtrace may be enabled by building umpire with 
+  -DENABLE_ALLOCATION_BACKTRACE
+
+- Umpire exceptions now include backtrace information in exception string.
```
davidbeckingsale:
```suggestion
- Umpire exceptions now include backtrace information in the exception string.
```

cmake/SetupUmpireThirdParty.cmake
```diff
@@ -48,3 +48,7 @@ if (ENABLE_SLIC AND ENABLE_LOGGING)
                         LIBRARIES ${SLIC_LIBRARY} ${SLIC_UTIL_LIBRARY}
                       )
 endif ()
+
+blt_register_library( NAME backtrace_symbols
```
davidbeckingsale: This needs to be guarded I think, because it won't work on every system right?

examples/CMakeLists.txt
```diff
@@ -30,6 +30,16 @@ blt_add_executable(
   DEPENDS_ON ${example_depends})
 list(APPEND umpire_examples allocator)
 
+blt_add_executable(
```
davidbeckingsale: This needs to be guarded - only build if `ENABLE_ALLOCATION_BACKTRACE` is On.

examples/backtrace_example.cpp
```diff
@@ -0,0 +1,60 @@
+//////////////////////////////////////////////////////////////////////////////
```
davidbeckingsale: I'm a bit unclear what the example is supposed to be showing. Can you clarify (in reply to this is fine).

src/umpire/ResourceManager.cpp
```diff
@@ -137,6 +139,16 @@ ResourceManager::ResourceManager() :
 ResourceManager::~ResourceManager()
 {
   for (auto&& allocator : m_allocators) {
+    if (allocator->getCurrentSize() != 0) {
+      std::stringstream ss;
+
+      umpire::print_allocator_records(Allocator{allocator.get()}, ss);
+
+      UMPIRE_LOG(Error, allocator->getName() << " Allocator still has "
```
davidbeckingsale: Error means it will be printed to stdout by default. Do we want to add a variable to control where it gets printed?

src/umpire/util/Macros.hpp
```diff
@@ -87,12 +88,14 @@
 
 #define UMPIRE_USE_VAR(x) static_cast<void>(x)
 
-
 #define UMPIRE_ERROR( msg )                                        \
 {                                                                  \
-  UMPIRE_LOG(Error, msg);                                          \
+  umpire::util::Backtrace backtrace;                               \
```
davidbeckingsale: Just to confirm, we now _always_ get a backtrace on error?

CHANGELOG.md
```diff
@@ -24,6 +29,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 
 ### Fixed
 
+- Clang builds using `-stdlib=lib++` option have been fixed.
```
davidbeckingsale:
```suggestion
- Clang builds using `-stdlib=libc++` option have been fixed.
```

tools/replay/replaydiff.cpp
```diff
@@ -60,6 +65,13 @@ int main(int argc, char* argv[])
     std::cerr << "Miscompare:" << std::endl;
     return -2;
   }
+#else
+  UMPIRE_USE_VAR(argc);
+  UMPIRE_USE_VAR(argv);
+  std::cerr << "This program requires the ability to demangle C++" << std::endl
```
davidbeckingsale:
```suggestion
  std::cerr << "This program requires the ability to demangle C++ symbols." << std::endl
```

tools/replay/replaydiff.cpp
```diff
@@ -60,6 +65,13 @@ int main(int argc, char* argv[])
...
+    << "However, this program was compiled with stdlib=c++ which does " << std::endl
```
davidbeckingsale:
```suggestion
    << "However, this program was compiled with -stdlib=libc++ which does " << std::endl
```

examples/backtrace_example.cpp
mcfadden8: `alloc_leak_example` prints a backtrace for each active allocation.  Further, that memory is leaked because it was never free.  It will also show that the resource manager logs the leaked allocations along with backtraces for who allocated them

`umpire_exception_example` shows that a backtrace is included with the thrown exception string returned by `exc.what()`.

src/umpire/ResourceManager.cpp
mcfadden8: I think this behaviour changed at some point.  I am not seeing this come to stdout.  I only see it in the log file.

src/umpire/util/Macros.hpp
mcfadden8: Yes.  Except on platforms it is not supported (like Umpire).

examples/CMakeLists.txt
mcfadden8: But it shows both the case were ALLOCATION_BACKTRACE is optionally on and the case when exceptions are generated containing backtrace information.

davidbeckingsale: Okay, I understand.

examples/CMakeLists.txt
```diff
@@ -30,6 +30,16 @@ blt_add_executable(
...
+  NAME backtrace_example
+  SOURCES backtrace_example.cpp
+  DEPENDS_ON ${example_depends})
+list(APPEND umpire_examples backtrace_example)
+
+blt_add_target_link_flags(
```
davidbeckingsale: This needs guarding for MSVC?

src/umpire/util/Macros.hpp
davidbeckingsale: I assume you mean "like Windows".

tools/replay/replay.cpp
```diff
@@ -85,6 +90,12 @@ int main(int argc, char* argv[])
     time_span = std::chrono::duration_cast<std::chrono::duration<double>>(t2 - t1);
     std::cout << "Running replay took " << time_span.count() << " seconds." << std::endl;
   }
-
+#else
+  UMPIRE_USE_VAR(argc);
+  UMPIRE_USE_VAR(argv);
+  std::cerr << "This program requires the ability to demangle C++" << std::endl
+    << "However, this program was compiled with stdlib=c++ which does " << std::endl
```
davidbeckingsale:
```suggestion
    << "However, this program was compiled with -stdlib=libc++ which does " << std::endl
```

examples/CMakeLists.txt
davidbeckingsale: Adding `rdynamic` on Windows issues a warning about unknown flag, so it should be guarded. See here: https://dev.azure.com/davidbeckingsale/Umpire/_build/results?buildId=2769&view=results

src/umpire/util/Macros.hpp
mcfadden8: lol - oops. yes, like Windows

examples/backtrace_example.cpp
mcfadden8: @davidbeckingsale, may I mark this conversation resolved?

src/umpire/util/Macros.hpp
mcfadden8: @davidbeckingsale, is this resolved now?

davidbeckingsale: Yup.

PR #326, opened 2020-03-23
Add a new pool strategy

davidbeckingsale:


src/umpire/Replay.hpp
```diff
@@ -43,13 +46,13 @@ class Replay {
...
 
-    if (typeid(firstArg) != typeid(umpire::strategy::DynamicPool::CoalesceHeuristic)) {
+    //if (typeid(firstArg) != typeid(umpire::strategy::DynamicPool::CoalesceHeuristic)) {
       m_argument_number++;
       if ( m_argument_number != 1 )
         ss << ", ";
 
       ss << "\"" << firstArg << "\"";
-    }
+    //}
```
mcfadden8, line 55: I suggest we remove this 'if' since it seems to be working (or is it?, did we confirm that?)

src/umpire/util/size_map.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include <iostream>
+
+#include <cmath>
+#include <list>
+
+#include <algorithm>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnarrowing"
```
mcfadden8, line 18: Is this required and is it only needed for GCC?

PR #329, opened 2020-03-31
Adding filter mechanism to build_and_test for ci [feature/host-config-filter]

adrienbernede:
In Umpire, moving to the same filtering mechanism for hostconfig files as in RAJA (https://github.com/LLNL/RAJA/pull/737) shouldn’t be to much of a hassle.

This PR is an attempt to do so, and seems to work fine.

TODO:
Currently this is easy because umpire do not really use "filters". Indeed, there is one project host-config file per generic/toolchain host-config file. I’ll do the change necessary for the example described in https://github.com/LLNL/radiuss-ci/commit/f4e490e571af8341c1305362f0894856c4ff9ad4#commitcomment-38158500


adrienbernede:
LGTM

scripts/gitlab/build_and_test.sh
```diff
@@ -27,39 +30,86 @@ then
...
         echo "ERROR: Host-config file ${generic_conf} does not exist" && exit 1
     fi
 
-    umpire_conf="${project_dir}/${conf_suffix}"
-    if [[ ! -f ${umpire_conf} ]]
+    project_conf="${project_dir}/host-configs/${sys_type}/${host_config}"
+    if [[ ! -f ${project_conf} ]]
     then
-        echo "ERROR: Host-config file ${umpire_conf} does not exist" && exit 1
+        echo "ERROR: Host-config file ${project_conf} does not exist" && exit 1
```
mcfadden8, line 107: Can we make the existence of a project configuration file optional?  Right now, we have a bunch of empty project configuration files.

scripts/gitlab/build_and_test.sh
```diff
@@ -27,39 +30,86 @@ then
...
     fi
 
     cmake \
       -C ${generic_conf} \
-      -C ${umpire_conf} \
+      -C ${project_conf} \
```
mcfadden8, line 112: I don't see how this will perform the two filters required for [https://github.com/LLNL/radiuss-ci/commit/f4e490e571af8341c1305362f0894856c4ff9ad4#commitcomment-38158500](https://github.com/LLNL/radiuss-ci/commit/f4e490e571af8341c1305362f0894856c4ff9ad4#commitcomment-38158500).  Won't this only overlay one of the filters?

scripts/gitlab/build_and_test.sh
adrienbernede: Let’s say all the project host-config files for gcc* are empty. Then the filter mechanism allows you to create only one, named `gcc_X.cmake` for all the matching toolchain.
So there should be much less empty files needed.

scripts/gitlab/build_and_test.sh
adrienbernede: In `project_conf`, the suffix `conf_suffix` correspond to the filter found for the `CONFIGURATION` variable passed.

So, if you ask for `gcc_8_1_0__stdc++`, `__stdc++` will be seen as a _tuning_ suffix, and `gcc_8_1_0` will have to match a toolchain host-config file.

Then the script will look for the filter matching `gcc_8_1_0` with tuning `__stdc++`.

In Umpire there are not filter yet, there is exactly one project host-config file per toolchain. But if you decide to merge all gcc host-config files into `gcc_X.camke` and `gcc_X_stdc++.cmake`, then the script would pick-up the latter.

You may also read the "Third action" section in https://github.com/LLNL/RAJA/pull/737.

davidbeckingsale:
This needs the fix from #333, but I'm going to merge it anyway since it's unrelated.

PR #331, opened 2020-04-02
Ignore .vscode project directory

mcfadden8:
For now, I'd like to ignore the workspace settings control directory the vscode places in the open workspace.  We may wish to revisit this later if we decide to put project-specific settings in there that we would like other (vscode) developers to see.  But for now it is just an annoyance.


davidbeckingsale:
Thanks @mcfadden8 !

PR #334, opened 2020-04-08
Add shared Windows build to Azure

davidbeckingsale:


davidbeckingsale:
@robinson96 - check this out, these are the changes required to have a shared build work on Windows (via CMake).

The def files get auto-generated and could be copied if you want to build without CMake. You should be able to maintain these in your Visual Studio project.

PR #341, opened 2020-05-04
Make DeviceAllocator trap on out of memory

davidbeckingsale:


mcfadden8:
@davidbeckingsale, Don't (some) device allocators already do this?

[Cuda Malloc](https://github.com/LLNL/Umpire/blob/develop/src/umpire/alloc/CudaMallocAllocator.hpp#L30-L36)

Can/Should we enforce common behavior across the various devices where OOM may occur?

davidbeckingsale:
This is for the `DeviceAllocator` that is used inside kernels. The host-side functions that allocate memory already have error code in them.

PR #345, opened 2020-05-15
Feature/um 146 shared memory mpi.v2

mcfadden8:
@davidbeckingsale, this is still a work in progress.  We need to discuss how to best implement `getSharedAllocator`.


src/umpire/SharedMemoryAllocator.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <string>
+
+namespace umpire {
+
+SharedMemoryAllocator::SharedMemoryAllocator(strategy::SharedAllocationStrategy* allocator) noexcept:
+  m_allocator(allocator)
+{
+}
+
+void* SharedMemoryAllocator::allocate(std::size_t bytes)
```
davidbeckingsale, line 22: I thought that SharedMemoryAllocator's _had_ to allocate with a name?

src/umpire/SharedMemoryAllocator.cpp
```diff
@@ -0,0 +1,175 @@
...
+  UMPIRE_LOG(Debug, "() returning " << m_allocator);
+  return m_allocator;
+}
+
+Platform SharedMemoryAllocator::getPlatform() noexcept
+{
+  return m_allocator->getPlatform();
+}
+
+void SharedMemoryAllocator::set_foreman(int id)
```
davidbeckingsale, line 152: Why does the foreman need "setting"? Should it be a constructor param?

PR #348, opened 2020-06-09
CI Clean-up

adrienbernede:
No functionality change in this PR.

- Removes mentions of UMPIRE in variable naming (makes the yaml file easier to use in another project).
- Removes the start job, not necessary in newer Gitlab. The `needs` key can now be given an empty list in which case the job will start immediately.


adrienbernede:
LGTM

PR #350, opened 2020-06-11
Fix exit with error if ctest failed

adrienbernede:
I noticed that a unit test for intel 16.0.4 on quartz was failing but this wasn’t caught by CI.

This is due to a mistake in exit status management in build_and_test.sh script.

This PR fixes that.


adrienbernede:
LGTM

PR #357, opened 2020-06-23
Introduce Uberenv and use host-config files generated with spack

adrienbernede:
This branch introduces Uberenv in the project in order to generate host-config files.

- [X] Add uberenv
- [X] Convert toss3 hostconfigs (quartz)
- [x] Convert blueos hostconfigs (lassen)
- [x] Convert blueos hostconfigs (butte)
~~- [ ] Drop radiuss-ci~~
- [X] Adapt CI to use new hostconfigs


adrienbernede:
LGTM

adrienbernede:
LGTM

adrienbernede:
LGTM

adrienbernede:
Note: we can’t drop radiuss-ci and old host-config files at the moment because they are used in bamboo.
Note2: host-config files now contains the machine name. This makes sense since they may have fine tuning info like the cuda arch.

Instead, I want to work on generating host-config on-the-fly in CI, and make them available as artifacts.

adrienbernede:
@davidbeckingsale I think the PR is ready as-is.

Generating host-config on-the-fly isn’t a feature we need immediately. In addition, I tested it and since it implies downloading and using Spack, it significantly slows down the pipeline, so we probably want to keep that for a specific scenario.

adrienbernede:
> Hey @adrienbernede - this looks good. Why are we still maintaining generated host-configs if they can be automatically regenerated by Spack? Also, can those configs actually be regenerated, or do they contain any extra additions?

Your question makes sense. As I said in my previous comment, generating them is slow and would slow down the CI for PRs unnecessarily. In particular it would require downloading Spack.

They are exactly as generated. That’s the point, they should not need any tweaking.

adrienbernede:
Versioning generated files is not a best practice, I know :)

davidbeckingsale:
@adrienbernede - can you fixup the conflict?

adrienbernede:
Definitely, I did not see there was one.

adrienbernede:
LGTM

mcfadden8:
@adrienbernede, I just started reviewing this (as @davidbeckingsale requested).  The initial comment says "WIP: At the moment, only toss3 hostconfig files have been ported.".  Is this outdated?  I notice some checkboxes below that indicate blueos has been ported?  Could you clarify?

mcfadden8:
> Note: we can’t drop radiuss-ci and old host-config files at the moment because they are used in bamboo.
> Note2: host-config files now contains the machine name. This makes sense since they may have fine tuning info like the cuda arch.
> 
> Instead, I want to work on generating host-config on-the-fly in CI, and make them available as artifacts.

I thought that one of the benefits of implementing this PR was that we could stop using bamboo.  Is this incorrect?  To rephrase, I would like to stop using bamboo ASAP.

davidbeckingsale:
> > Note: we can’t drop radiuss-ci and old host-config files at the moment because they are used in bamboo.
> > Note2: host-config files now contains the machine name. This makes sense since they may have fine tuning info like the cuda arch.
> > Instead, I want to work on generating host-config on-the-fly in CI, and make them available as artifacts.
> 
> I thought that one of the benefits of implementing this PR was that we could stop using bamboo. Is this incorrect? To rephrase, I would like to stop using bamboo ASAP.

I think we can, but @adrienbernede wanted to make sure we didn't _have_ to.

adrienbernede:
@mcfadden8 Initial comment fixed, you were right.
@davidbeckingsale is correct: I didn’t want to break bamboo, and since I can’t trigger bamboo, I preferred leaving the "old" host-config in place at first.

mcfadden8:
> @mcfadden8 Initial comment fixed, you were right.
> @davidbeckingsale is correct: I didn’t want to break bamboo, and since I can’t trigger bamboo, I preferred leaving the "old" host-config in place at first.

Okay.  I am very motivated to move off of bamboo. :)

adrienbernede:
Note:
I couldn’t test the documentation, it wouldn’t build on my machine.

PR #359, opened 2020-06-25
Add aligned allocator

davidbeckingsale:


src/umpire/strategy/AlignedAllocator.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+  if (m_allocator->getPlatform() != Platform::host) {
+    UMPIRE_ERROR("Cannot construct AlignedAllocator from non-host Allocator.");
+  }
+}
+
+void* 
+AlignedAllocator::allocate(std::size_t bytes)
+{
+  std::size_t total_bytes = bytes+sizeof(void*)+m_alignment-1;
```
mcfadden8, line 32: This only works for alignments that are a power of 2.  I suggest checking/throwing an error otherwise.

src/umpire/strategy/AlignedAllocator.cpp
```diff
@@ -0,0 +1,72 @@
...
+  UMPIRE_LOG(Debug, "requested: " << bytes << " actual: " << bytes+m_alignment-1);
+
+  uintptr_t ptr{reinterpret_cast<uintptr_t>(m_allocator->allocate(total_bytes))};
+  uintptr_t aligned_ptr{static_cast<uintptr_t>((reinterpret_cast<uintptr_t>(ptr) + sizeof(void*) + (m_alignment-1)) & m_mask)}; 
+  uintptr_t* header = (uintptr_t*) (aligned_ptr - sizeof(void*));
+  *header = ptr;
```
mcfadden8, line 42: I don't think this assignment will work for memory that the HOST does not have access to.

tests/integration/strategy_tests.cpp
```diff
@@ -1483,3 +1485,80 @@ TEST(NumaPolicyTest, Location) {
...
+
+  p1 &= align;
+  p2 &= align;
+
+  ASSERT_NE(p1, p2);
+  ASSERT_TRUE( p1 == 0 || p1 == align);
+  ASSERT_TRUE( p2 == 0 || p2 == align);
+}
+
+TEST(AlignedAllocator, AllocateAlign8)
```
mcfadden8, line 1505:
```suggestion
TEST(AlignedAllocator, AllocateAlign256)
```

tests/integration/strategy_tests.cpp
```diff
@@ -1483,3 +1485,80 @@ TEST(NumaPolicyTest, Location) {
...
+{
+  unsigned int align = 256;
+  auto& rm = umpire::ResourceManager::getInstance();
+  auto alloc = rm.makeAllocator<umpire::strategy::AlignedAllocator>(
+    "aligned_allocator_8", rm.getAllocator("HOST"), align);
```
mcfadden8, line 1510:
```suggestion
    "aligned_allocator_256", rm.getAllocator("HOST"), align);
```

src/umpire/strategy/AlignedAllocator.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+void* 
+AlignedAllocator::allocate(std::size_t bytes)
+{
+  std::size_t total_bytes = bytes+sizeof(void*)+m_alignment-1;
+  UMPIRE_LOG(Debug, "requested: " << bytes << " actual: " << bytes+m_alignment-1);
+
+  uintptr_t ptr{reinterpret_cast<uintptr_t>(m_allocator->allocate(total_bytes))};
+  uintptr_t aligned_ptr{static_cast<uintptr_t>((reinterpret_cast<uintptr_t>(ptr) + sizeof(void*) + (m_alignment-1)) & m_mask)}; 
+  uintptr_t* header = (uintptr_t*) (aligned_ptr - sizeof(void*));
```
mcfadden8, line 41: This line assumes that the minimum alignment size is whatever is required for a uintptr_t*.  I believe that this means that the alignment cannot be 1, 2, or (possibly) 4 (it certainly cannot be 1 :-)).  I suggest that we update the constructor to make certain the alignment is a power of 2 that is >= 8.

PR #364, opened 2020-07-01
Updating Fortran bindings with private helpers

davidbeckingsale:


src/umpire/interface/c_fortran/wrapResourceManager.cpp
```diff
@@ -1,22 +1,15 @@
...
 // project contributors. See the COPYRIGHT file for details.
 //
 // SPDX-License-Identifier: (MIT)
 #include "wrapResourceManager.h"
+#include <stdlib.h>
 #include <string>
 #include "umpire/Allocator.hpp"
 #include "umpire/ResourceManager.hpp"
 
-#include "umpire/strategy/AllocationAdvisor.hpp"
```
davidbeckingsale, line 12: @ltaylor16 - I'm not sure why these lines get removed, but now this fails to compile...

ltaylor16:
It looks like an older version of Shroud was used.  The old files say Shroud 0.11.0 and the new files do not have a version but should say 0.12.1

davidbeckingsale:
> It looks like an older version of Shroud was used. The old files say Shroud 0.11.0 and the new files do not have a version but should say 0.12.1

Wow, I really need to learn how to install Python tools... thanks Lee, sorry for bothering you.

ltaylor16:
Looking at the changes, there are some derived types which need to be private since they may be different depending on which version of Shroud generated the files.
Let me update Shroud before you merge.

PR #365, opened 2020-07-09
Fix for using external camp

adayton1:
This pull request adds some fixes for using external camp.


src/umpire/tpl/CMakeLists.txt
```diff
@@ -114,7 +114,6 @@ if (NOT TARGET camp)
     blt_register_library(
       NAME camp
       DEPENDS_ON cuda_runtime
-      LIBRARIES camp
```
davidbeckingsale, line 117: @adayton1 - why is this removed?

adayton1: Doesn't this try to add -lcamp to the link line?

adayton1: @white238 @gzagaris camp is a header only library, so should `LIBRARIES camp` be left out of the call to `blt_register_library`?

white238: So it does a bunch of possible things:

1) If it is a BLT object library (which is done by blt_add_library(...OBJECT TRUE...)) it circumvents the normal CMake object libraries that have a lot of side effects that are very unhelpful.

2) If it is a BLT registered library, it adds all the known information from the previous blt_register_library

(the one its probably doing) 3) The normal case calls target_link_library() (which is a normal CMake function), which if its a CMake target, inherits information following the CMake normal rules, or if it is not a CMake target, it blindly adds -lcamp to the link line in hopes that it will find it in the system or an already added -L directory.

If you need more clarification I can look into it further.

davidbeckingsale: @white238 - camp provides a proper CMake target regardless of whether it's "external" or not. The only extra requirement is the `CAMP_HAVE_CUDA` define to make sure it enables stuff that's guarded in the headers.

What's the best way to do this?

davidbeckingsale:
@adayton1 - can you give this a try?

davidbeckingsale:
@adayton1 - just wanted to confirm you are happy with this after our email discussion? I will go ahead and update the installed CMake config in another PR to handle the CHAI issue.

davidbeckingsale:
LGTM

adayton1:
This looks good to me now. I tried to review it, but github is smart enough to not let me review my own pull request :)

adayton1:
Another thought occurred to me: Could camp configure a header so that we don't have to add defines? That would also guarantee consistency across projects.

davidbeckingsale:
> Another thought occurred to me: Could camp configure a header so that we don't have to add defines? That would also guarantee consistency across projects.

I replied to the issue you opened on camp.

PR #369, opened 2020-07-20
Bugfix/um 645 alignedallocator dynamicpool not giving aligned allocations

mcfadden8:


mcfadden8:
This problem is no longer reproducible.

PR #373, opened 2020-07-22
Created new tests/applications location for standalone test applications of Umpire

mcfadden8:


tests/applications/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+blt_add_executable(
+  NAME pool_allocator_stress
+  SOURCES pool_allocator_stress.cpp
+  DEPENDS_ON ${test_applications_depends})
+list(APPEND umpire_test_applications pool_allocator_stress)
+
+if (NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+  blt_add_target_link_flags(
+    TO pool_allocator_stress
+    FLAGS "-rdynamic -ldl")
```
davidbeckingsale, line 18: We can use `CMAKE_DL_LIBS` instead of `-ldl`: https://cmake.org/cmake/help/latest/variable/CMAKE_DL_LIBS.html

mcfadden8: Done!

mcfadden8: Wait.  This isn't right.  Since there can be multiple library names, I need to find a different way to do this.  I'm not sure how to do this with `blt_add_target_link_flags`.  I must need to use something different.

mcfadden8: Okay @davidbeckingsale, please go ahead and review.  I think this should be ready.

tests/CMakeLists.txt
```diff
@@ -7,4 +7,5 @@
 add_subdirectory(tpl)
 add_subdirectory(unit)
 add_subdirectory(integration)
-add_subdirectory(tools)
\ No newline at end of file
+add_subdirectory(tools)
+add_subdirectory(applications)
```
davidbeckingsale, line 11: I think we want this guarded with some CMake option and not included by default?

mcfadden8: `ENABLE_DEVELOPER_DEFAULTS` perhaps?

davidbeckingsale: That's fine with me

mcfadden8: Done @davidbeckingsale - ready for another review..

PR #375, opened 2020-07-23
Feature/um 650 add umpire poison memory region checks to quickpool

mcfadden8:


mcfadden8:
Should be good for final checks now.

davidbeckingsale:
I remember why it's guarded... The `UMPIRE_POISON` macros have to check the platform (adds an if statement), so that's why I made sure it could be disabled. Do you think that really matters?

mcfadden8:
Hmm.

mcfadden8:
Almost there.  I think I have the compilers on bamboo working.

mcfadden8:
Okay @davidbeckingsale , this is ready.

PR #378, opened 2020-07-24
Host config files: full configuration + generated in CI.

### LINKED ISSUE #356: Regroup builds and unit tests in Gitlab CI
I would like to merge builds with their associated tests in quartz CI.

I think separation of builds and unit-tests of quartz only has proved to be of little interest, and it’s a pain to maintain (duplication).

This would only make it less straightforward to say at a glance if the job failed in build or in test, but we could improve error message in the log to help with this.

Note:
- For a project managed with GitLab, one could export the tests reports to have them formatted in a Merge Request. Making build and unit-test separation of even smaller interest for projects on Gitlab.
- A similar result (reporting) could be achieved with a CDash server, I think.

### LINKED ISSUE #367: Hostconfig and Spack package: Where should we draw the line
### The context

In the Spack package, we can change a lot of parameters to configure the build.
All these parameters are not transposed to the host-config file because it would make them too specific, and naming them (for example) would be a pain.

Instead, a host-config file aims at configuring the toolchain, and how the projects adapts to it.

In practices:

#### Everything "host-config" goes into the `hostconfig` function:

https://github.com/LLNL/Umpire/blob/7e4f2f3c30fc3470b6f3fa2808b837dee2c47939/scripts/uberenv/packages/umpire/package.py#L119

We find there the flags, the compilers paths, the cuda info, etc.

#### Everything "build-config" goes into the `cmake_args` function:

https://github.com/LLNL/Umpire/blob/7e4f2f3c30fc3470b6f3fa2808b837dee2c47939/scripts/uberenv/packages/umpire/package.py#L255

We find there the languages support, tests mode, openmp option, etc.

### The issue

I wanted to point this out to the team because in order to manually build the project, picking a host config may not be enough, unless one wants to rely on the defaults value of the project (if they are set). Instead, one may have to specify some variables on command line.

Also some parameters may be misplaced:
- "deviceconst" may move from `hostconfig` to `cmake_args`,
- "openmp" may move from `cmake_args` to `hostconfig`. 

### Solution

I would appreciate in someone could review the different parameters and discuss with me if it should be placed in the hostconfig, or specified on command line (typically).

### Takaway

A host-config file does not suffice to reproduce a Spack build, although I may have said it was.

adrienbernede:
Resolves #367 
Resolves #356

There may seem to be a lot of changes, but the CI functions remains the same, only improved:

- Merge build and tests for quartz jobs.

- Drop external build directory anymore (used for split build and test -> dropped, and recording purpose -> we can use artifacts instead if we need to).

- Split CI configuration in "templates" + "jobs", preparing for job generation and also logical split.

- Generate host config files on the fly in CI. Performances issues where solved cloning only one commit in Spack and using it in /dev/shm.

- All CMake parameters in Spack are now places in host-config. ENABLE_TESTS was removed because "False" case is broken if not setting ENABLE_BENCHMARKS accordingly. So ENABLE_TESTS is left to the default value: "True".

- Drop unused code in build_and_test.sh script.

- Use full hash in host-config name (could be reduced to partial hash).

Note:

At this stage, we could drop the host-config directory.


.gitlab-ci.yml
```diff
@@ -59,30 +59,23 @@ stages:
...
     - echo ${JOBID}
-    - srun $( [[ -n "${JOBID}" ]] && echo "--jobid=${JOBID}" ) -t 10 -N 1 -n 1 -c 4 scripts/gitlab/build_and_test.sh --build-only
-
-.test_toss_3_x86_64_ib_script:
-  script:
-    - echo ${ALLOC_NAME}
-    - export JOBID=$(squeue -h --name=${ALLOC_NAME} --format=%A)
-    - echo ${JOBID}
-    - srun $( [[ -n "${JOBID}" ]] && echo "--jobid=${JOBID}" ) -t 10 -N 1 -n 1 -c 4 scripts/gitlab/build_and_test.sh --test-only
+    - srun $( [[ -n "${JOBID}" ]] && echo "--jobid=${JOBID}" ) -t 10 -N 1 scripts/gitlab/build_and_test.sh
```
mcfadden8, line 62: Will this result in a single failure for either build or test?  One of the things I would like to see is a (more) clear indication of whether the failure is due to an infrastructure (git failures), build, or test failure.  Idealy this would be clearly shown in the subject line of the email.

host-configs/hc-butte-blueos_3_ppc64le_ib_p9-xl@beta2019.06.20-cuda-7ymlfxgrfplb5tztddd6ta46zc2iypwg.cmake
```diff
@@ -0,0 +1,41 @@
+###################
+# Generated host-config - Edit at own risk!
```
mcfadden8, line 2: Perhaps this was already discussed somewhere previously, but why are generated files being checked in to the repo?

adrienbernede: See PR comment: We could now drop them :)

At first, generating them was costly in CI (because starting from scratch for each job meant downloading Spack and configuring it). I optimized the mechanism so that it’s now much faster in CI.

Spack is still downloaded each time, but this clone has been optimized (single branch, single commit).

.gitlab-ci.yml
adrienbernede: We could surely improve that. Please open an issue describing what you would like to have, that’s the best way for me to keep track of tasks.

host-configs/hc-butte-blueos_3_ppc64le_ib_p9-xl@beta2019.06.20-cuda-7ymlfxgrfplb5tztddd6ta46zc2iypwg.cmake
mcfadden8: Great to hear.  Let's drop them.

.gitlab-ci.yml
mcfadden8: #381

host-configs/hc-butte-blueos_3_ppc64le_ib_p9-xl@beta2019.06.20-cuda-7ymlfxgrfplb5tztddd6ta46zc2iypwg.cmake
adrienbernede: Done.

Reminder: old host-config files are still used in Bamboo, so I left them here for now.

adrienbernede:
LGTM

mcfadden8:
@adrienbernede, would you please send me a link to the gitlab CI pages?  I would like to review the output produced by your most recent commit.  Additionally, I would like to start getting the email notifications from that CI.

Thanks,
Marty

adrienbernede:
> @adrienbernede, would you please send me a link to the gitlab CI pages? I would like to review the output produced by your most recent commit. Additionally, I would like to start getting the email notifications from that CI.
> 
> Thanks,
> Marty

image

This is how to access GitLab directly from GitHub.
The result is: https://lc.llnl.gov/gitlab/radiuss/Umpire/pipelines/14328

adrienbernede:
Note:

`scripts/gitlab/generate_host_configs.py` and `scripts/gitlab/list_of_specs.json` are in fact unused.
They allow the generate all or a selection of the host-config files in the list of specs.
They could be useful in the near future to generate the jobs in CI, or if we want to move to python instead of bash as script language.

Should I remove them from the repo for now?

mcfadden8:
I would leave those for now.  I am not sure how @davidbeckingsale feels about this, but I would be completely fine with using python instead of bash.

Also, do we understand why one of the builds is failing? https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/82629/raw

adrienbernede:
LGTM

davidbeckingsale:
LGTM

adrienbernede:
LGTM

adrienbernede:
@davidbeckingsale I don’t know what’s wrong exactly.
This passed yesterday, but I thought I could just add a small non critical fix, but it turns out it’s more not as trivial as it seems.
So I back to yesterday version of the branch, expect now it fails in Azure, and it seems to be a recurring issue...
Do you know more about it?

davidbeckingsale:
> @davidbeckingsale I don’t know what’s wrong exactly.
> This passed yesterday, but I thought I could just add a small non critical fix, but it turns out it’s more not as trivial as it seems.
> So I back to yesterday version of the branch, expect now it fails in Azure, and it seems to be a recurring issue...
> Do you know more about it?

This will be fixed by #382 - I had to update the hip docker container which is causing this failure.

.gitlab/butte-jobs.yml
```diff
@@ -5,92 +5,67 @@
...
   variables:
-    HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-clang@9.0.0.cmake"
+    SPEC: "%clang@4.0.0"
   extends: .build_and_test_on_butte_advanced
 
-gcc_8_3_1 (build and test on butte):
+clang_9_0_0_libcpp (build and test on butte):
   variables:
-    HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-gcc@8.3.1.cmake"
+    SPEC: "%clang@9.0.0"
```
davidbeckingsale, line 49: Hey @adrienbernede - this is the only thing I'm confused about, which is the spec for clang_9 with libc++ `%clang%9.0.0`? Shouldn't the libc++ be a variant?

adrienbernede: Good catch!
That’s something I want to solve in the future: there is no guarantee that the name of the job describe the spec.

davidbeckingsale: Okay, I see. So in this case, the build is clang9, but the job is just named wrong? (it's not a problem, just wondering)

adrienbernede: Fixed.
Note: this wasn’t in the CI list of specs before, I may have introduced it to show-case variant use, in replacement of clang@9.0.0 alone. If you want to keep clang@9.0.0 too, just tell me so.

adrienbernede:
LGTM

.gitlab/butte-jobs.yml
adrienbernede: I added the variant (changed the spec) instead of changing the job name :).

adrienbernede:
LGTM

davidbeckingsale:
@adrienbernede - I'll merge this when CI passes.

PR #384, opened 2020-07-29
Allow compilation on Windows.

agcapps:
On Windows, we need to include direct.h and call _mkdir().


agcapps:
Thanks @davidbeckingsale .  No, I haven't tried building with C++17.  The code I'm working with is at C++11 (considering officially moving to C++14).  Since I'm not an authorized Umpire contributor, will you merge this PR?  Thanks.

PR #394, opened 2020-08-10
Add missing Caliper dependency in various subdirectories

daboehme:
Compiler complained about missing Caliper includes as the Caliper dependency was missing in some subdirectories' CMake scripts.


davidbeckingsale:
Thanks @daboehme !

PR #396, opened 2020-08-12
Apply clang-format to all code

davidbeckingsale:


davidbeckingsale:
@mcfadden8 @kab163 - take a look and comment if there's anything formatted in a way you dislike.

src/umpire/Allocator.cpp
```diff
@@ -6,91 +6,80 @@
 //////////////////////////////////////////////////////////////////////////////
 #include "umpire/Allocator.hpp"
 
+#include "umpire/Replay.hpp"
```
mcfadden8, line 9: Sorting header files too!  Nice.

src/umpire/Allocator.cpp
```diff
@@ -6,91 +6,80 @@
...
+                << m_allocator << "\" }");
 
   UMPIRE_LOG(Debug, "");
 
   m_allocator->release();
 }
 
-std::size_t
-Allocator::getSize(void* ptr) const
+std::size_t Allocator::getSize(void* ptr) const
```
mcfadden8, line 35: I actually prefer this style of keeping the method on the same line as the return type when possible.

davidbeckingsale: Haha this is the one thing I don't really like 😄

src/umpire/Replay.cpp
```diff
@@ -5,50 +5,45 @@
...
-  replayEnabled =  enable_replay;
+  replayEnabled = enable_replay;
 }
 
-void Replay::logMessage( const std::string& message )
+void Replay::logMessage(const std::string& message)
 {
-  if ( !replayEnabled )
-    return;   /* short-circuit */
+  if (!replayEnabled) return; /* short-circuit */
```
mcfadden8, line 41: hmm.  this is the only one I see that may be problematic @davidbeckingsale.  I would prefer:
```suggestion
  if (!replayEnabled) {
    return; /* short-circuit */
  }
```

src/umpire/Replay.cpp
```diff
@@ -5,50 +5,45 @@
...
 
   umpire::replay() << message;
 }
 
-bool Replay::replayLoggingEnabled()
-{
-  return replayEnabled;
-}
+bool Replay::replayLoggingEnabled() { return replayEnabled; }
```
mcfadden8, line 46: Same comment as above.  I would prefer:
```suggestion
bool Replay::replayLoggingEnabled()
{
  return replayEnabled;
}
```

davidbeckingsale:
@mcfadden8 - I think this addresses your comments.

src/umpire/Replay.cpp
mcfadden8: @davidbeckingsale, this one still seems to be present.  Can we modify it to add curly braces around open if statements?

src/umpire/DeviceAllocator.cpp
```diff
@@ -12,33 +12,31 @@
...
   auto device_alloc = rm.getAllocator(umpire::resource::Device);
 
-  m_counter = static_cast<unsigned int*>(
-    device_alloc.allocate(sizeof(unsigned int)));
+  m_counter =
+      static_cast<unsigned int*>(device_alloc.allocate(sizeof(unsigned int)));
   rm.memset(m_counter, 0);
 }
 
 __host__ __device__
```
kab163, line 29: Line 29 has `__host__ __device` on it's own line while Line 39 had this changed. Should this be changed?

davidbeckingsale: I think it remains split due to line length.

kab163: Yea, that seems likely. If you see DeviceAllocator.hpp, line 30, I think that looks a lot better. But this is ok :p

kab163:
So I wanted to double check: if statements, classes, and for loops have 
`if (something) {
...
}`
but functions have
`int main () 
{
...
}`
that seems to be the trend.

davidbeckingsale:
LGTM

src/umpire/Replay.cpp
mcfadden8: Resolving this since the reformatting tool only makes white-space changes by design.

davidbeckingsale:
LGTM

PR #402, opened 2020-08-13
Bugfix/um 678 memory leaks

kab163:


davidbeckingsale:
LGTM

src/umpire/strategy/QuickPool.hpp
```diff
@@ -99,6 +99,12 @@ class QuickPool : public AllocationStrategy, private mixins::AlignedAllocation {
     {
     }
 
+    ~pool_allocator()
+    {
+      if (pool != nullptr)
```
davidbeckingsale, line 104: This can probably be `if (pool)` but I'm not too bothered either way.

src/umpire/strategy/QuickPool.hpp
```diff
@@ -99,6 +99,12 @@ class QuickPool : public AllocationStrategy, private mixins::AlignedAllocation {
...
+        delete(pool);
```
mcfadden8, line 105: The only question I had here was whether we can use RAII instead of having to perform a `new/delete` pair?  I don't believe that pointer is required.

tests/integration/strategy_tests.cpp
```diff
@@ -470,15 +470,13 @@ TEST(ThreadSafeAllocator, HostStdThread)
...
-    threads.push_back(std::thread([=, &allocator, &thread_allocs] {
-      for (int j = 0; j < N; ++j) {
-        thread_allocs[i] = allocator.allocate(1024);
-        ASSERT_NE(thread_allocs[i], nullptr);
-        allocator.deallocate(thread_allocs[i]);
-        thread_allocs[i] = allocator.allocate(1024);
-        ASSERT_NE(thread_allocs[i], nullptr);
-      }
+  for (std::size_t i = 0; i < N; i++) {
+    threads.push_back(std::thread([=, &allocator, &thread_allocs] {
```
mcfadden8, line 474: The intention of this test was to make certain that thread synchronization is working as expected for the allocator.  So the idea was to have each thread perform a large-ish (N) set of operations against the allocator with the hope that it would cause a lot of synchronization.  This change will cause the thread to only execute one sequence of allocate/deallocate.

Do we know why the original `for (int j = 0; j < N; ++j)) {` loop was causing a memory leak?

tools/replay/ReplayOperationManager.cpp
```diff
@@ -37,7 +37,14 @@ ReplayOperationManager::ReplayOperationManager( ReplayFile::Header* Operations )
 {
 }
 
-ReplayOperationManager::~ReplayOperationManager() { }
+ReplayOperationManager::~ReplayOperationManager()
```
mcfadden8, line 40: Nice catch here.

tests/integration/strategy_tests.cpp
kab163: I don't know for sure the exact cause, but I used ThreadSanitizer on this and found that this file had data races. I think that somewhere in those data races, a memory leak was occurring - and specifically on the second allocate call). Removing that loop solved the data races (except one minor one) and the leaks with it. This is documented in more detail on the JIRA ticket UM-678 (what the memory leak and data race errors were).

mcfadden8: Okay thanks @kab163, let's continue the discussion in UM-678.

src/umpire/strategy/QuickPool.hpp
kab163: Good idea. I could try play around with the pool pointer and see if it can be left a normal variable, not a pointer.

tests/integration/strategy_tests.cpp
```diff
@@ -470,15 +470,13 @@ TEST(ThreadSafeAllocator, HostStdThread)
   std::vector<void*> thread_allocs{N};
   std::vector<std::thread> threads;
 
-  for (std::size_t i = 0; i < N; ++i) {
-    threads.push_back(std::thread([=, &allocator, &thread_allocs] {
-      for (int j = 0; j < N; ++j) {
-        thread_allocs[i] = allocator.allocate(1024);
-        ASSERT_NE(thread_allocs[i], nullptr);
-        allocator.deallocate(thread_allocs[i]);
-        thread_allocs[i] = allocator.allocate(1024);
```
davidbeckingsale, line 479: This line is what caused the leak.

src/umpire/strategy/QuickPool.hpp
kab163: Ok - I took out the pointer and it seems to pass all the AddressSanitizer/LeakSanitizer tests.. I can do another push and y'all can check it out. I assume it is best to remove the pointer?

davidbeckingsale: Yes, please.

kab163:
LGTM

src/umpire/strategy/QuickPool.hpp
```diff
@@ -95,27 +95,27 @@ class QuickPool : public AllocationStrategy, private mixins::AlignedAllocation {
...
     using difference_type = std::ptrdiff_t;
 
-    pool_allocator() : pool{new util::FixedMallocPool{sizeof(Value)}}
+    pool_allocator() : pool{util::FixedMallocPool{sizeof(Value)}}
     {
     }
 
     /// BUG: Only required for MSVC
     template <typename U>
-    pool_allocator(const pool_allocator<U>& other) : pool{other.pool}
```
kab163, line 104: This might be wrong... (the & might be needed actually)

davidbeckingsale: Yeah, it needs to be a reference.

kab163:
LGTM!

kab163:
LGTM

tests/integration/strategy_tests.cpp
```diff
@@ -470,15 +470,15 @@ TEST(ThreadSafeAllocator, HostStdThread)
   std::vector<void*> thread_allocs{N};
   std::vector<std::thread> threads;
 
-  for (std::size_t i = 0; i < N; ++i) {
+  for (std::size_t i = 0; i < N; i++) {
```
kab163, line 473: I don't think that change was needed, but it shouldn't hurt either...

kab163:
LGTM

src/umpire/strategy/QuickPool.hpp
```diff
@@ -95,7 +95,7 @@ class QuickPool : public AllocationStrategy, private mixins::AlignedAllocation {
     using size_type = std::size_t;
     using difference_type = std::ptrdiff_t;
 
-    pool_allocator() : pool{new util::FixedMallocPool{sizeof(Value)}}
+    pool_allocator() : pool{util::FixedMallocPool{sizeof(Value)}}
```
davidbeckingsale, line 98: Unintended copy construction here.

davidbeckingsale:
LGTM

kab163:
LGTM

davidbeckingsale:
> LGTM

I beat you to it! 😂

kab163:
LGTM

kab163:
@mcfadden8 I will merge this, ya?

PR #409, opened 2020-08-24
Feature/use different workflow for develop

mcfadden8:
I think that this should stop doing CHANGELOG checking for pushes into the develop branch.

I split the `push.yml` file into two:
- `push-all-branches.yml` will run Static Analysis and ABI Compatability Check jobs for `push` events
- `push-all-but-develop.yml` will run the Check CHANGELOG job for `push` events

This will cause the following to happen:
- `push` events for all branches will run jobs specfied in both `push-all-branches.yml` and `push-all-but-develop-branches.yml` files
- `push` events for `develop` branch will only run jobs specified in `push-all-branches.yml` file


mcfadden8:
LGTM

mcfadden8:
LGTM (ugh)

mcfadden8:
LGTM

PR #411, opened 2020-08-26
Fixup ABI checker

davidbeckingsale:


.github/actions/abi-compat/entrypoint.sh
```diff
@@ -7,7 +7,8 @@ cmake -DENABLE_DEVELOPER_DEFAULTS=On -DCMAKE_CXX_FLAGS="-Og" -DCMAKE_BUILD_TYPE=
 make -j 3 umpire
 cd ..
 
-git checkout origin/main
```
mcfadden8, line 10: So what tid this wind up doing then?

davidbeckingsale: Nothing, for some reason...

.github/actions/abi-compat/entrypoint.sh
```diff
@@ -18,4 +19,4 @@ cd ..
 abi-dumper build-main/lib/libumpire.so -o umpire-main.dump -lver 3.0.0
 abi-dumper build/lib/libumpire.so -o umpire-dev.dump -lver 4.0.0
 
-abi-compliance-checker -l umpire -old umpire-main.dump -new umpire-dev.dump
+abi-compliance-checker -l umpire -old umpire-main.dump -new umpire-dev.dump || true
```
mcfadden8, line 22: Where we will see the failure indication if the abi-compliance-checker fails?

davidbeckingsale: Good point. Let me consider how to do this.

PR #420, opened 2020-09-03
Fix QuickPool NVTX + ASAN calls

anis-ladram:
Allocation size passed to NVTX is incorrect, causing compute-sanitizer not to report incorrect write.

### DIFF
--- a/src/umpire/strategy/QuickPool.cpp
+++ b/src/umpire/strategy/QuickPool.cpp
@@ -40,6 +40,7 @@ QuickPool::~QuickPool()
 void* QuickPool::allocate(std::size_t bytes)
 {
   UMPIRE_LOG(Debug, "allocate(" << bytes << ")");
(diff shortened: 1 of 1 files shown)


davidbeckingsale:
Thanks @anis-ladram - we need to fix this in multiple places so will get that worked up in another branch.

mcfadden8:
Hello @anis-ladram and @davidbeckingsale, I have a fix here (https://github.com/LLNL/Umpire/pull/431) that is intended to address this problem for all three of our primary pools (QuickPool, DynamicPoolList, and DyanmicPoolMap) as well as the aligned allocator that they all share.  @anis-ladram, if you have time and are interested in trying this branch out, I would be interested in hearing any feedback you might have.

Thanks,

Marty

anis-ladram:
@mcfadden8 sure thing! just left some comments on the review :)

PR #430, opened 2020-09-08
Feature/um 716 improve destruction resilliance of primary pools

mcfadden8:
1. Added logging to primary pools to better reflect pool life cycle
2. Added (missing) REPLAY of coalesce operation to `QuickPool`
3. Modified destructors of primary pools to try/catch/log/ignore deallocation failures from the device which may happen if destruction occurrs after the device's runtime has already shut down.

### DIFF
--- a/src/umpire/strategy/DynamicPoolMap.cpp
+++ b/src/umpire/strategy/DynamicPoolMap.cpp
@@ -35,10 +35,19 @@ DynamicPoolMap::DynamicPoolMap(
       m_next_minimum_pool_allocation_size{
           aligned_round_up(next_minimum_pool_allocation_size)}
 {
+  UMPIRE_LOG(Debug, " { "
+    << "Name(\"" << name << "\")"
+    << ", Id(" << id << ")"
+    << ", For(\"" << allocator.getName() << "\")"
+    << " 1st_alloc(" << m_first_minimum_pool_allocation_size << ")"
+    << ", next_alloc(" << m_next_minimum_pool_allocation_size << ")"
+    << ", align(" << alignment << ")"
+    << " }");
 }
 
 DynamicPoolMap::~DynamicPoolMap()
 {
+  UMPIRE_LOG(Debug, "Releasing free blocks to device");
   // Get as many whole blocks as possible in the m_free_map
   mergeFreeBlocks();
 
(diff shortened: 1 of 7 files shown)


mcfadden8:
LGTM

src/umpire/strategy/DynamicPoolMap.cpp
```diff
@@ -35,10 +35,20 @@ DynamicPoolMap::DynamicPoolMap(
       m_next_minimum_pool_allocation_size{
           aligned_round_up(next_minimum_pool_allocation_size)}
 {
+  UMPIRE_LOG(Debug, " { "
```
davidbeckingsale, line 38: Can you reformat this to look like how the debug constructor log normally looks? I think the format you used before is arg=val for all the arguments.

src/umpire/strategy/DynamicPoolMap.cpp
```diff
@@ -301,6 +234,129 @@ MemoryResourceTraits DynamicPoolMap::getTraits() const noexcept
...
+  try {
+    aligned_deallocate(ptr);
+  }
+  catch (...) {
+    if (m_is_destructing) {
+      //
+      // Ignore the error if we are destructing as the resource to which we
+      // are releasing to may have gone away
+      //
+      UMPIRE_LOG(Error, "Pool is destructing, Exception Ignored");
```
davidbeckingsale, line 352: I think you should print the exception message. The underlying allocator will (and should) always exist. The comment above doesn't reflect the real issue this is fixing.

src/umpire/strategy/DynamicSizePool.hpp
```diff
@@ -226,10 +216,27 @@ class DynamicSizePool : private umpire::strategy::mixins::AlignedAllocation {
...
+        try {
+          aligned_deallocate(curr->data);
+        }
+        catch (...) {
+          if (m_is_destructing) {
+            //
+            // Ignore the error if we are destructing as the resource to which we
+            // are releasing to may have gone away
+            //
+            UMPIRE_LOG(Error, "Pool is destructing, Exception Ignored");
```
davidbeckingsale, line 234: Same comment as other pool.

src/umpire/strategy/DynamicSizePool.hpp
```diff
@@ -282,44 +289,53 @@ class DynamicSizePool : private umpire::strategy::mixins::AlignedAllocation {
         m_next_minimum_pool_allocation_size{
             aligned_round_up(next_minimum_pool_allocation_size)}
   {
+    UMPIRE_LOG(Debug, " { "
```
davidbeckingsale, line 292: Reformat message please.

src/umpire/strategy/QuickPool.cpp
```diff
@@ -28,16 +27,26 @@ QuickPool::QuickPool(const std::string& name, int id, Allocator allocator,
       m_next_minimum_pool_allocation_size{
           aligned_round_up(next_minimum_pool_allocation_size)}
 {
+  UMPIRE_LOG(Debug, " { "
```
davidbeckingsale, line 30: Reformat.

src/umpire/strategy/QuickPool.cpp
```diff
@@ -207,7 +216,22 @@ void QuickPool::release()
...
+      try {
+        aligned_deallocate(chunk->data);
+      }
+      catch (...) {
+        if (m_is_destructing) {
+          //
+          // Ignore the error if we are destructing as the resource to which we
+          // are releasing to may have gone away
+          //
+          UMPIRE_LOG(Error, "Pool is destructing, Exception Ignored");
```
davidbeckingsale, line 229: Fix comment, log the exception message.

davidbeckingsale (changes requested):

src/umpire/strategy/DynamicPoolMap.cpp
mcfadden8: I will fix the comment, but the underlying umpire exception is already printed right (by the call to UMPIRE_ERROR)

davidbeckingsale: Yup, you're correct.

mcfadden8: and sorry, but the comment is correct.  I will change it to: "...the resource to which we are releasing from...".  The 'resource' could be anything that offers a mechanism external to Umpire to cause it to be reset.

src/umpire/strategy/QuickPool.cpp
mcfadden8: Are you asking me to distinguish between an Umpire exception and some other type of exception?  If it is an Umpire exception, the message will be logged by UMPIRE_ERROR.

I don't understand what you are looking for in the comment.  Please suggest an alternative?

mcfadden8: Done

src/umpire/strategy/QuickPool.cpp
mcfadden8: Done

src/umpire/strategy/DynamicSizePool.hpp
mcfadden8: Done

src/umpire/strategy/DynamicSizePool.hpp
mcfadden8: Done

src/umpire/strategy/DynamicPoolMap.cpp
mcfadden8: Done

src/umpire/strategy/DynamicPoolMap.cpp
mcfadden8: Done

mcfadden8:
LGTM

kab163 (approved):
Looks good!

davidbeckingsale (approved):

PR #431, opened 2020-09-08
Bugfix/um 718 umpire pool asan poisoning is not posoning all necessary regions

mcfadden8:


mcfadden8:
LGTM

src/umpire/strategy/mixins/AlignedAllocation.inl
```diff
@@ -17,28 +17,30 @@ namespace mixins {
...
-    return size + (m_alignment - 1) - (size - 1) % m_alignment;
+  return size + (m_alignment - 1) - (size - 1) % m_alignment;
 }
 
 inline void* AlignedAllocation::aligned_allocate(std::size_t size)
 {
-    std::size_t total_bytes{ size + m_alignment - 1 };
-    uintptr_t ptr{ reinterpret_cast<uintptr_t>(m_allocator->allocate(total_bytes)) };
+  std::size_t total_bytes{ size + m_alignment };
+  uintptr_t ptr{ reinterpret_cast<uintptr_t>(m_allocator->allocate(total_bytes)) };
```
anis-ladram, line 26: Before doing this, why don't you check if the underlying allocator already produces aligned addresses? Because of this, when using CUDA, and despite our 256-bytes aligned memory guarantee (c.f. [CUDA programming guide](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses), section 5.3.2.1.1), you might allocate too much memory (e.g. if size == 4096, you will allocate 4096 + 16 == 4112).

This might be a large change considering the way you stack allocation strategies, don't know if it's really worth...

src/umpire/strategy/mixins/AlignedAllocation.inl
```diff
@@ -17,28 +17,30 @@ namespace mixins {
...
 
-    uintptr_t alignment{ m_alignment - 1 };
-    uintptr_t mask{ m_mask };
-    void *aligned_ptr{ reinterpret_cast<void*>( (ptr + alignment) & mask ) };
+  UMPIRE_POISON_MEMORY_REGION(m_allocator, reinterpret_cast<void*>(ptr), total_bytes);
 
-    base_pointer_map[aligned_ptr] = reinterpret_cast<void*>(ptr);
+  uintptr_t alignment{ m_alignment - 1 };
+  uintptr_t mask{ m_mask };
+  void *aligned_ptr{ reinterpret_cast<void*>( (ptr + alignment) & mask ) };
```
anis-ladram, line 32: c.f. comment on line 26.

src/umpire/strategy/mixins/AlignedAllocation.inl
```diff
@@ -17,28 +17,30 @@ namespace mixins {
...
 {
-    std::size_t total_bytes{ size + m_alignment - 1 };
-    uintptr_t ptr{ reinterpret_cast<uintptr_t>(m_allocator->allocate(total_bytes)) };
+  std::size_t total_bytes{ size + m_alignment };
+  uintptr_t ptr{ reinterpret_cast<uintptr_t>(m_allocator->allocate(total_bytes)) };
 
-    uintptr_t alignment{ m_alignment - 1 };
-    uintptr_t mask{ m_mask };
-    void *aligned_ptr{ reinterpret_cast<void*>( (ptr + alignment) & mask ) };
+  UMPIRE_POISON_MEMORY_REGION(m_allocator, reinterpret_cast<void*>(ptr), total_bytes);
```
anis-ladram, line 28: Isn't memory already marked for ASAN in top-level allocation strategies like QuickPool? You could say that you poison memory only when you're sure this is the low-level allocator (e.g. CudaMallocAllocator). Maybe I don't understand everything here.

src/umpire/strategy/mixins/AlignedAllocation.inl
mcfadden8: This is a fair question.  This mixin is intended for pool strategies.  I wasn't concerned about the additional size since this allocation will be made for large chunks of memory (several megabytes to gigabytes range) to be added to the pool.

src/umpire/strategy/mixins/AlignedAllocation.inl
mcfadden8: If the memory is aligned, then the base_pointer and aligned pointer will simply be the same.

src/umpire/strategy/mixins/AlignedAllocation.inl
mcfadden8: Yes, but some of this memory may have been missed if an alignment would have been necessary.  It does seem wasteful to poison memory twice however.  I think I will change this to continue to poison the memory region here in the mixin since it knows the base pointer and total size of what was allocated.  I will be providing an update to the pools to no longer perform the redundant poison after teh aligned_allocate call.

mcfadden8:
LGTM

src/umpire/strategy/mixins/AlignedAllocation.inl
anis-ladram: > some of this memory may have been missed if an alignment would have been necessary

I'm not sure what you mean by that. I think we should poison the whole memory, not only the aligned part. Doing that in the lowest level (i.e. the one where cudaMalloc is actually called) guarantees that the whole memory is poisoned, and that poisoning is not called twice.

mcfadden8: You are correct.  However, for a pool strategy, this is the lowest level.  Let me try to explain:

If an Allocator is built directly on a resource (CudaMallocAllocator), then the memory allocated must not be poisoned because the application calling allocate will start to use the memory right away.

If an Allocator is built on top of a pool strategy, then the pool will return memory from the pool it is managing.  This pool will separately be filled by chunk allocations from CudaMallocAllocator when needed.  When new chunks are given to the pool, they will more thank likely be (much) larger than the calling application is requesting.  So, those chunks should initially be given to the pool as poisoned.

mcfadden8: So, the lowest level allocator will be called by two consumers.  By the application directly and by a pool.  When the lowest level allocator is called directly from the application, it cannot be poisoned.  When it is called by the pool, it should be poisoned.

Does this make sense?

anis-ladram: Makes sense! Thank you for that explanation 😊

mcfadden8:
LGTM

mcfadden8:
LGTM

src/umpire/strategy/mixins/AlignedAllocation.inl
mcfadden8: Since applications may ask for arbitrary alignments ranging from a small power of 2 to potentially page-size multiples, this is the mechanism we are using which is wasteful as noted above, but is small compared relative to the sizes being allocated by the pool which is in the multi-MiB/GiB range.  Let me know if this explanation sounds reasonable and I will mark this conversation as resolved.

mcfadden8:
LGTM

mcfadden8:
Also Addresses #420 and #433

davidbeckingsale (approved):

src/umpire/strategy/DynamicPoolMap.cpp
```diff
@@ -106,7 +104,7 @@ void* DynamicPoolMap::allocateBlock(std::size_t bytes)
                                  << umpire::util::backtracer<>::print(bt));
     }
 #endif
-    ptr = aligned_allocate(bytes);
+    ptr = aligned_allocate(bytes);  // Will POISON
```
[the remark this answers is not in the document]
davidbeckingsale: Not too worried, thanks!

PR #434, opened 2020-09-14
Adding fixes to solve warning messages

kab163:
After recompiling on Lassen, I didn't get the same warning messages as before. However, I still kept the same fixes and things will compile with no warnings/errors. For the record, the error we got before was:
`/g/g0/belcher6/Umpire/src/umpire/resource/MemoryResourceTypes.hpp(46): warning: missing return statement at end of non-void function "umpire::resource::resource_to_string"

/g/g0/belcher6/Umpire/src/umpire/resource/MemoryResourceTypes.hpp(57): warning: missing return statement at end of non-void function "umpire::resource::string_to_resource"`
So we added an "Unknown" Resource Type to return if the end of the function was reached. This should solve the problem that came up before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 
 - Fixed Allocator overrun problem in replay tool
 
+- Removed warnings from MemoryResourceTypes header file
+
 ## [v4.0.1] - 2020-09-03
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


kab163:
LGTM

davidbeckingsale (changes requested):
Looks good but can you update the CHANGELOG please?

kab163:
LGTM

davidbeckingsale (approved):

PR #444, opened 2020-09-18
Task/um 696 allocator accessibility

kab163:
Notes:
1) Change required for HIP within blt stuff, but I couldn't include that modification in my branch - see comment in Umpire.cpp 
2) Table comment for is_accessible
3) Use of "UNIQUE" in example test - see comment in allocAccess.cpp file
4) Couldn't test HIP stuff because machine is still down, but it does compile at least


kab163:
LGTM

kab163:
LGTM

src/umpire/Umpire.cpp
```diff
@@ -98,6 +113,166 @@ bool pointer_contains(void* left_ptr, void* right_ptr)
...
+  switch(p) {
+    case (cPlatform::host):
+    {
+      if(findResource(a) == myResource::UNKNOWN
+       || findResource(a) == myResource::DEVICE_CONST)
+        return false;
+
+#if defined(UMPIRE_ENABLE_CUDA)
+      //check DEVICE case for (Umpire) CUDA
+      return is_host_pageable_from_cuda();
```
davidbeckingsale, line 182: Can you merge these into one `is_host_pageable` function?

src/umpire/Umpire.hpp
```diff
@@ -122,6 +122,21 @@ bool pointer_overlaps(void* left, void* right);
  */
 bool pointer_contains(void* left, void* right);
 
+/*!
+ * \brief Check whether or not an Allocator is accessible from a given platform.
+ *
+ *\param camp::Platform p  
+ *\param umpire::Allocator a 
+ */
+bool isAccessible(Platform p, Allocator a);
```
davidbeckingsale, line 131:
```suggestion
bool is_accessible(Platform p, Allocator a);
```

Use snake_case for free functions.

kab163:
LGTM

src/umpire/Umpire.hpp
```diff
@@ -122,6 +122,21 @@ bool pointer_overlaps(void* left, void* right);
...
+
+/*!
+ * \brief Helper function to find the resource type of a given allocator
+ *
+ *\param umpire::allocator a  
+ */
+MemoryResourceTraits::resource_type findResource(Allocator a);
```
davidbeckingsale, line 138: If you want to keep this public, can you rename to `get_resource`. I think that "find" makes it sound like it has to search for something, but we already know the resource.

Alternatively, you could move this into an anonymous namespace in Umpire.cpp so it won't be exposed to users.

davidbeckingsale (changes requested):

src/umpire/Umpire.hpp
kab163: This is a good catch! I didn't see that. (And I think odd that it didn't give me a warning about it)

kab163:
LGTM

kab163:
LGTM

src/umpire/Umpire.cpp
```diff
@@ -13,8 +13,20 @@
...
 #include "umpire/ResourceManager.hpp"
+#include "umpire/util/MemoryResourceTraits.hpp"
+#include "camp/resource/platform.hpp"
 #include "umpire/config.hpp"
 
+#if defined(UMPIRE_ENABLE_CUDA)
+  #include <cuda_runtime_api.h>
+#endif
+
+//make sure "/opt/rocm-3.6.0/hsa/include/" is
```
kab163, line 24: Not sure how or if i can edit the blt stuff - is this the best I can do for this?

tests/integration/allocator_accessibility.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "gtest/gtest.h"
+#include "gtest/gtest-death-test.h"
+#include "umpire/Allocator.hpp"
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Umpire.hpp"
+#include "umpire/config.hpp"
+#include "umpire/util/MemoryResourceTraits.hpp"
+
+//Need to enable gtest death tests with 
+// -DENABLE_GTEST_DEATH_TESTS=On
```
kab163, line 16: This is another thing I'm unsure about. Whenever this test is added to the CMakeLists.txt, this flag must be turned on. Maybe we can add this flag to the Developer Defaults? Or something else to make it automatic (I don't think the tests will run currently)

tests/integration/allocator_accessibility.cpp
```diff
@@ -0,0 +1,175 @@
...
+    else {
+      double* data = static_cast<double*>(m_allocator->allocate(m_size * sizeof(double)));
+      ASSERT_DEATH(test(data, m_size, dev), "");
+    }
+  }
+}
+#endif
+
+/////////////////////////////
+//Sycl test not yet available
```
kab163, line 163: This could be left as is, or I could mimic the OpenMP Target test just so that something is here. For now, there's just this placeholder.

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM for the win!!

kab163:
LGTM

kab163:
Plz work

kab163:
LGTM

kab163:
LGTM for the win!!

kab163:
Documenting our CUDA and MPI problem:
Tests with CUDA work if ran with  lrun -n 1 --smpiargs="-disable_gpu_hooks" ./bin/allocator_accessibility_tests
More info here:
https://docs.olcf.ornl.gov/systems/summit_user_guide.html#cuda-hook-error-when-program-uses-cuda-without-first-calling-mpi-init

kab163:
LGTM

kab163:
LGTM

src/umpire/ResourceManager.hpp
```diff
@@ -84,6 +84,17 @@ class ResourceManager {
    */
   Allocator getDefaultAllocator();
 
+  /*!
+   * \brief Get the existing Resources.
```
mcfadden8, line 88:
```suggestion
   * \brief Get the names for existing Resources.
```

src/umpire/Umpire.cpp
mcfadden8: Is this a change that is required in blt?  If so, has an issue been written up there?  I'm not sure these comments belong in the source code.

src/umpire/Umpire.cpp
```diff
@@ -98,6 +113,141 @@ bool pointer_contains(void* left_ptr, void* right_ptr)
...
 
+bool is_host_pageable()
+{
+#if defined(UMPIRE_ENABLE_HIP)
+  hipDeviceProp_t props;
+  int hdev = 0;
+  hipGetDevice(&hdev);
+
+  //Check whether HIP can map host memory.
+  hipGetDeviceProperties(&props, hdev);
```
mcfadden8, line 124: Can these device-implementation-specific calls be placed in to static methods of the specific resource instead of being placed here?

examples/allocator_access.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include "umpire/Allocator.hpp"
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Umpire.hpp"
+
+using cPlatform = camp::resources::Platform;
+
+bool testAccess(umpire::Allocator a)
+{
+  if(is_accessible(cPlatform::host, a)) {
```
mcfadden8, line 18: Shouldn't this call have a namespace qualifier?  I am curious why this compiles.  Is there a `using umpire` somewhere?  if so, that doesn't seem right to me.

```suggestion
  if(umpire::is_accessible(cPlatform::host, a)) {
```

tests/integration/allocator_accessibility.cpp
```diff
@@ -0,0 +1,174 @@
...
+    delete m_allocator;
+  }
+  
+  umpire::Allocator* m_allocator;
+  size_t m_size = 42;
+};
+
+TEST_P(AllocatorAccessibilityTest, AccessibilityFromPlatform)
+{
+  if(is_accessible(cPlatform::host, *m_allocator)) {
```
mcfadden8, line 119: Similar question to another comment made in a different file.  Shouldn't this have a namespace qualified?  I'm confused why it compiles.
```suggestion
  if(umpire::is_accessible(cPlatform::host, *m_allocator)) {
```

mcfadden8 (changes requested):

src/umpire/ResourceManager.hpp
kab163: Oooo good catch!

src/umpire/Umpire.cpp
kab163: You're right. I wasn't sure how to handle this, but if I didn't add that path, the "hsa/hsa.h" file couldn't be found (needed for hip stuff). I think it's a blt change requirement, but I'm not sure how to make edits to the blt submodule stuff. I agree that the comment doesn't belong there, it was kind of a placeholder for figuring it out later lol

@davidbeckingsale any thoughts?

davidbeckingsale: You can now set the CMake variable HIP_RUNTIME_INCLUDE_DIRS - take a look here for an example: https://github.com/LLNL/Umpire/blob/develop/Dockerfile#L69

kab163: Great, I'll remove this comment then. Thanks!

examples/allocator_access.cpp
kab163: The is_accessible function is housed in the Umpire.hpp file, which is included. I assume that's why it doesn't complain of an error. I could try to do some kind of statement like `using access = <path to is_accessible function>` similar to how I did the `cPlatform` statement instead (I'd have to test it, haven't tried as of yet). Is one better than the other?

tests/integration/allocator_accessibility.cpp
kab163: The umpire.hpp file is included, I assume that's why this works. The is_accessible function is not a member of the umpire namespace, so if I did something like that, it'd have to make it something like in line 14 or 15. Am I explaining this right? I feel like I'm making this confusing, but do you see what I'm talking about?

src/umpire/Umpire.cpp
davidbeckingsale: They cannot be static since some info (e.g. device number) would be a property of a specific resource instance.

tests/integration/allocator_accessibility.cpp
mcfadden8: Oh, yes.  I get it.  I think that these functions need to be part of the umpire namespace.

src/umpire/Umpire.cpp
kab163: Good idea, stay tuned! :)

tests/integration/allocator_accessibility.cpp
davidbeckingsale: The function is in the `umpire` namespace. I think there might a `using namespace umpire` line somewhere that makes this work.

davidbeckingsale: Adding a `using` declaration like on line 14 or 15 is not necessary if you explicitly write out the fully qualified name.

mcfadden8: If there is a `using umpire` statement somewhere, I would like to get it removed (especially if it is in a header).  using namespaces should be discouraged in libraries.

kab163: Oh, well Umpire.hpp has the umpire namespace, maybe when I include that file, it kinda gets added in I guess

mcfadden8: uh oh.  IMHO, that is a problem.

kab163:
LGTM

kab163:
LGTM

kab163:
TODO: make default to return false instead of returning true in isAccessible() functions; double check cases; update Accessibility Table documentation so there's a device-hip, device-cuda, etc. version.

kab163:
LGTM

kab163:
LGTM

scripts/uberenv/packages/umpire/package.py
```diff
@@ -319,6 +319,8 @@ def hostconfig(self, spec, prefix, py_site_pkgs_dir=None):
                                         hip_root))
             cfg.write(cmake_cache_entry("HIP_CLANG_PATH",
                                         rocm_root + '/llvm/bin'))
+            cfg.write(cmake_cache_entry("HIP_HIPCC_FLAGS",
+                                        '--amdgpu-target=gfx906'))
```
mcfadden8, line 323: Was this change required for the allocator accessibility check or was it needed for something else?

src/umpire/Umpire.hpp
```diff
@@ -122,6 +122,19 @@ bool pointer_overlaps(void* left, void* right);
...
+ *
+ * This function describes which allocators should be accessible
+ * from which CAMP platforms. Information on platform/allocator
+ * accessibility can be found at
+ * <https://umpire.readthedocs.io/en/develop/features/allocator_accessibility.html>
+ *
+ *\param camp::Platform p  
+ *\param umpire::Allocator a 
+ */
+bool is_accessible(Platform p, Allocator a);
```
mcfadden8, line 136: Shouldn't this be:
```suggestion
bool is_accessible(camp::Platform p, Allocator a);
```

scripts/uberenv/packages/umpire/package.py
kab163: It was needed so that when we are from the perspective of the host platform, for example, we can ask if the right device resource is accessible (I think like the main device, not DEVICE::1 or something). When this flag wasn't included, I got some odd HIP errors about device usage. This just makes it so that the flag is always included.

src/umpire/Umpire.cpp
```diff
@@ -98,6 +99,20 @@ bool pointer_contains(void* left_ptr, void* right_ptr)
   }
 }
 
+bool is_accessible(Platform p, Allocator a)
```
mcfadden8, line 102: I'm not sure how the `camp` namespace is being made available.  But since this is a library function, shouldn't the camp namespace be explicitly stated?
```suggestion
bool is_accessible(camp::Platform p, Allocator a) 
```

src/umpire/Umpire.cpp
```diff
@@ -12,6 +12,7 @@
 #include <iterator>
 #include <sstream>
 
+#include "umpire/resource/MemoryResource.hpp"
 #include "umpire/ResourceManager.hpp"
 #include "umpire/config.hpp"
```
mcfadden8, line 17: Shouldn't the implementation file also include the camp.hpp file?

src/umpire/Umpire.hpp
kab163: I include the Platform.hpp file, so "Platform p" is really doing "camp::resources::Platform p" - I don't think it's needed, but I may be wrong.

mcfadden8: Oh.  I see.  The comment at line 133 says "camp::Platform".   Platform, umpire::Platform, and camp::Platform are going to be confusing.  I think Platform.hpp existed before camp came around.  My personal opinion is we should somehow deprecate Platform.hpp and instead use camp::Platform.

src/umpire/alloc/MallocAllocator.hpp
```diff
@@ -50,6 +58,47 @@ struct MallocAllocator {
     UMPIRE_LOG(Debug, "(ptr=" << ptr << ")");
     ::free(ptr);
   }
+
+  bool isHostPageable()
+  {
+#if defined(UMPIRE_ENABLE_HIP)
+    hipDeviceProp_t props;
+    int hdev = 0;
+    hipGetDevice(&hdev);
```
mcfadden8, line 67: Similar to what was done with device-specific calls in ResourceManager, can these calls also be moved down into the resource?

mcfadden8: Or are we saying that things in umpire/alloc are the same as things in umpire/resources?  (@davidbeckingsale , I am not sure?)

davidbeckingsale: They have to stay here.

src/umpire/Umpire.cpp
davidbeckingsale: We bring the Platform enum into the umpire namespace too.

mcfadden8 (changes requested):
I think this is looking great!  I especially like the way the implementation looks in ResourceManager.cpp.  I just had a few questions.

scripts/uberenv/packages/umpire/package.py
davidbeckingsale: It's required because the accessibility tests generate device code. In the past we were only calling the HIP API, so this flag wasn't needed.

mcfadden8 (approved):

src/umpire/alloc/MallocAllocator.hpp
kab163: I think this is the best it can get. So this file is looking from the host resource memory perspective and asking if the cuda/hip platform is accessible. So this technically is down in the resource, but because i have to ask from the host resource if it can access the device platforms, I have to use the device-specific calls. If you look at the HipDeviceMemoryResource.cpp file, you'll see an example of how I ask, from the device memory resource perspective, if the host platform is accessible (so the reverse of this file, in regard to hip). 

This is some very detailed stuff, so we can have a webex or something to discuss it if you'd like.

kab163: @davidbeckingsale, any thoughts?

src/umpire/Umpire.hpp
kab163: Wait, maybe I don't lol let me look..

src/umpire/alloc/MallocAllocator.hpp
davidbeckingsale: I think we should discuss this next Tuesday, but I'm happy with how this works right now. This isn't resource-specific since a camp::Platform doesn't define things like the device number, so we just have to do the best way can to determine whether or not the host memory is accessible from that platform.

src/umpire/Umpire.hpp
davidbeckingsale: We do use camp::Platform, but umpire::Platform is an alias.

kab163:
LGTM

kab163:
LGTM

kab163:
TODO: find out more about why hip is_accessible() function fix was necessary; change the getResourceNames() func. in ResourceManager()

kab163:
LGTM

src/umpire/ResourceManager.hpp
```diff
@@ -84,6 +84,17 @@ class ResourceManager {
...
   Allocator getDefaultAllocator();
 
+  /*!
+   * \brief Get the names for existing Resources.
+   *
+   * The Resource Manager dynamically populates available memory resource types with
+   * m_resource_names based on what's available. This is used for determining 
+   * allocator accessibility.
+   *
+   * \return The private member, m_resource_names.
```
davidbeckingsale, line 94: This comment is out of date now.

src/umpire/resource/HipDeviceMemoryResource.cpp
```diff
@@ -65,6 +65,30 @@ std::size_t HipDeviceMemoryResource::getHighWatermark() const noexcept
...
+  else
+    return false;
+}
+
+bool HipDeviceMemoryResource::isAccessibleFrom(Platform p) noexcept
+{
+  if(p == Platform::hip)
+    return true;
+  else if(p == Platform::host)
+    return isHostPageable();
```
davidbeckingsale, line 87: I don't think you need to check this. As far as I know, HIP device memory will always be accessible from the host.

davidbeckingsale (changes requested):
Couple of quick comments.

kab163:
LGTM

src/umpire/resource/HipDeviceMemoryResource.cpp
kab163: I updated this check so that if its the host or hip it will return true, otherwise return false

kab163:
LGTM

src/umpire/alloc/MallocAllocator.hpp
```diff
@@ -50,6 +58,35 @@ struct MallocAllocator {
...
+    //Device supports coherently accessing pageable memory
+    //without calling cudaHostRegister on it
+    cudaDeviceGetAttribute(&pageableMem,
+              cudaDevAttrPageableMemoryAccess, cdev);
+    if(pageableMem)
+      return true;
+    else
+      return false;
+#endif
+    return false; //shouldn't reach this
```
mcfadden8, line 78: You could actually drop the "else return false" from lines 75 and 76 above and the code would just fall through to this line and return false if not pageable or UMPIRE_ENABLE_CUDA is not defined.  Then you could also remove the comment and not have to worry about knowing that the only time this function is called is if the platform is Platform::cuda.

tests/integration/allocator_accessibility.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include "gtest/gtest.h"
+#include "umpire/Allocator.hpp"
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Umpire.hpp"
+#include "umpire/config.hpp"
+#include "umpire/util/MemoryResourceTraits.hpp"
+#include "umpire/strategy/QuickPool.hpp"
+
+//using cPlatform = camp::resources::Platform;
+//using myResource = umpire::MemoryResourceTraits::resource_type;
```
mcfadden8, line 16: Should these two comments be removed?

mcfadden8 (commented):
Looking good.  Some minor changes requested.  Also, the gitlab CI builds need to be fixed before merge.

tests/integration/allocator_accessibility.cpp
kab163: done

src/umpire/alloc/MallocAllocator.hpp
kab163: good idea, done!

examples/allocator_access.cpp
```diff
@@ -0,0 +1,72 @@
...
+//
+// SPDX-License-Identifier: (MIT)
+//////////////////////////////////////////////////////////////////////////////
+#include <iostream>
+#include <string>
+#include "umpire/Allocator.hpp"
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Umpire.hpp"
+
+bool testAccess(umpire::Allocator a)
```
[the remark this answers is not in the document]
kab163: very true, done.

examples/allocator_access.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+  ///////////////////////////////////////////////////
+  //Test accessibility
+  ///////////////////////////////////////////////////
+  std::cout << "Testing the available allocators for "
+            << "accessibility from the CAMP host platform:" 
+            << std::endl << std::endl;
+  const int size = 100;
+  for(auto a : alloc) {
+    if(testAccess(a)) {
```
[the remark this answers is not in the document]
kab163: done!

examples/allocator_access.cpp
```diff
@@ -0,0 +1,72 @@
...
+      int* data = static_cast<int*>(a.allocate(size*sizeof(int)));
+      for(int i = 0; i < size; i++) {
+        data[i] = i * i;
+      }
+      std::cout << "Testing: " << data[size-1] << " should be equal to " 
+                << (size-1)*(size-1) << std::endl << std::endl; 
+    }
```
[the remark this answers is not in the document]
kab163: good idea again, done!

kab163:
LGTM

mcfadden8 (approved):
Looks good to me.  Thanks!

davidbeckingsale (approved):

PR #446, opened 2020-09-25
Added ENABLE_POOL_SANITIZER guard to support more compiler options

mcfadden8:


mcfadden8:
LGTM

mcfadden8:
Opted to implement this a different way (using cmake).

PR #450, opened 2020-09-29
Fix/junit report

adrienbernede:
Fix 2 issues with junit reports:

- toss3 -> toss_3: allows junit to be generated on quartz.
- lxml not found: generate and use a specific python environment in the pipeline.

TODO:

Use `cache` to improve efficiency.


adrienbernede:
LGTM

adrienbernede:
LGTM

.gitlab/quartz-templates.yml
```diff
@@ -30,7 +30,6 @@ allocate_resources (on quartz):
   stage: q_allocate_resources
   script:
     - salloc -N 1 -c 36 -p pdebug -t 20 --no-shell --job-name=${ALLOC_NAME}
-  needs: []
```
davidbeckingsale, line 33: Is this required?

davidbeckingsale (changes requested):
Looks great! Just one quick question.

.gitlab/quartz-templates.yml
adrienbernede: Let me explain, (and I’m thinking I should find a way to document this that is not super verbose...):
the ordering is:

Quartz:`environment` -> `resource alloc` -> `build_and_test` -> `resource release`.
Lassen: `build_and_test`
Corona: `build_and_test`

Strictly speaking, `needs: []` in `resource alloc` was not necessary before. It was here to make all 3 machines look similar, but `resource alloc` was the first stage in the pipeline, so it started immediately anyway.

Now, `environment` needs to be executed before `build_and_test`. So strictly speaking, we could say that `build_and_test` `needs: [environment, resource alloc]`. However, `resource alloc` is time sensitive: once resource is allocated, the jobs using it should start right away, but if `environment` takes too much time, it could delay `build_and_test`, wasting allocation time.

In the end, the easiest solution is to leave the first 4 stages (quartz stages) run in the default "stage" mode, and only mark lassen and corona as starting right away (with `needs: []`).

adrienbernede (commented):
Pipeline sequence explained.

davidbeckingsale (approved):

adrienbernede:
LGTM

PR #452, opened 2020-09-29
Using cache to optimize virtual env management

adrienbernede:
I’ve been willing to use cache correctly for a while, and seized the opportunity.

Relevant points:
- cache prevents from creating "out of clone" directories, like "PYTHON_ENVIRONMENT_PATH" was before this PR.
- cache prevents from reinstalling stuff multiple times.
- cache involves some data movements, which should be optimized not to burden the pipeline: see use of `policy:pull` for an example.
- a cache are shared will all pipelines, and all jobs, unless keys are use to restrict that. See documentation for more.


adrienbernede:
LGTM

davidbeckingsale (approved):

PR #458, opened 2020-10-02
Add a build-only type of job for variant 'tests=none'

adrienbernede:
Here is the suggested patch:

- Parse the host-config in build_and_test.sh script to check if tests are enabled in the build.
- The job succeeds, with an error message to mention no files were found to upload as artifact.


davidbeckingsale:
LGTM

PR #469, opened 2020-10-07
revert -Xcudafe option to ignore nvcc warning

mcfadden8:
Instead of turning off this option entirely, we now specifically implement the workaround in the offending header file.


mcfadden8:
LGTM

davidbeckingsale (approved):

PR #475, opened 2020-10-16
Story/um 695 get parent

kab163:


kab163:
LGTM

src/umpire/strategy/AllocationStrategy.hpp
```diff
@@ -120,12 +120,14 @@ class AllocationStrategy {
   friend std::ostream& operator<<(std::ostream& os,
                                   const AllocationStrategy& strategy);
 
+  virtual AllocationStrategy* getParentResource() const noexcept;
```
davidbeckingsale, line 123: I think we should call this `getParent` or `getParentAllocator`, since it might not be a resource.

davidbeckingsale (changes requested):
All looks good, just think we should tweak the name. I would like @mcfadden8's input too.

kab163:
LGTM

kab163:
LGTM

tests/integration/allocator_integration_tests.cpp
```diff
@@ -39,7 +39,8 @@ TEST_P(AllocatorTest, AllocateDeallocateBig)
       static_cast<double*>(m_allocator->allocate(m_big * sizeof(double)));
 
   ASSERT_NE(nullptr, data);
-
+  ASSERT_EQ(nullptr, m_allocator->getParent());
+
```
mcfadden8, line 43: Is there any way we can extend this test to also confirm we got the parent we are expecting by examining the traits of the parent?

mcfadden8: Also, how do the zero-byte allocators and introspection allocators fit in to the parent/child scheme?  Are they ever visible to the Umpire user?

davidbeckingsale:
Not really, since the `resource` member of the trait already propagates all the way up the chain.

tests/integration/strategy_tests.cpp
```diff
@@ -160,7 +173,8 @@ TYPED_TEST(StrategyTest, AllocateDeallocateBig)
       this->m_allocator->allocate(this->m_big * sizeof(double)));
 
   ASSERT_NE(nullptr, data);
-
+  ASSERT_EQ(this->m_allocator->getParent()->getName(), this->m_parent_name);
```
mcfadden8, line 176: nice.

mcfadden8:
Looks great in general.  It would be cool to have an example that walks the getParent() tree until it gets to the root and prints out the name of each thing it finds along the way.

mcfadden8:
> Not really, since the `resource` member of the trait already propagates all the way up the chain.

If not traits, then how about with using type information?

davidbeckingsale:
I agree with @mcfadden8 that we could have an exmaple that walks the tree and prints some names out. It is fine for this to be host-only, and just create a chain of about ~5 allocators.

davidbeckingsale:
> > Not really, since the `resource` member of the trait already propagates all the way up the chain.
> 
> If not traits, then how about with using type information?

I think that will be best tested in the accessibility branch, where @kab163 is going to add a function that walks to the root allocator and casts it to a specific resource type.

examples/get_parent.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "umpire/strategy/DynamicPool.hpp"
+
+void check_parent(umpire::Allocator alloc)
+{
+  umpire::strategy::AllocationStrategy* root = alloc.getAllocationStrategy();
+  while ((root->getParent() != nullptr)) {
+    root = root->getParent();
+    std::cout << root->getName() << std::endl;
+  }
+  std::cout << "(NULL)" << std::endl;
```
mcfadden8, line 18: I cannot remember whether resources have names or not.  If they don't then just ignore my suggestion to print the name.  But I believe this could be updated to:
```suggestion
  //
  // At this point, root will be pointing to the resource being used.
  //
  std::cout << "(root) allocator name is " << root->getName() << std::endl;
```
Of course, if I am incorrect, please just disregard.  Overall, I think this looks good and I will approve it no whether you decide to take this change or not.

mcfadden8 (approved):

examples/get_parent.cpp
kab163: I think that's a great comment! Thanks!

kab163:
LGTM

kab163:
LGTM

davidbeckingsale (changes requested):
The test for getParent->getName should be in its own test, not added as an assertion to all the other tests.

kab163:
LGTM

davidbeckingsale (approved):

PR #477, opened 2020-10-21
Add deregisterAllocator method

davidbeckingsale:


mcfadden8:
I have some semantic questions concerning the lifetime of the allocator.

If I understand correctly, the allocator does not actually go away, but may still be obtained by its integer ID correct?

Perhaps the intention for this is to be able to manage names/aliases of an allocator that persists?

If this is the case, perhaps a an `createAlias` / `deleteAlias` API might be in order?

On the other hand, I think that the allocator for this name is an actual allocator that has its own introspection available right?  How will statistic work when gathered for the integer ID if the name no longer exists?

davidbeckingsale:
> I have some semantic questions concerning the lifetime of the allocator.
> 
> If I understand correctly, the allocator does not actually go away, but may still be obtained by its integer ID correct?

Correct.

> Perhaps the intention for this is to be able to manage names/aliases of an allocator that persists?
>
> If this is the case, perhaps a an `createAlias` / `deleteAlias` API might be in order?

I agree that is clearer, but would require an API break. I'd appreciate input from @MrBurmark
 
> On the other hand, I think that the allocator for this name is an actual allocator that has its own introspection available right? How will statistic work when gathered for the integer ID if the name no longer exists?

Everything works because the Allocator still exists with that ID (and name), but it cannot be accessed by name anymore.

MrBurmark:
I agree that using the word alias would be more clear.
Not breaking compatibility would let me use this much more easily, so maybe registerAllocator could stay around in a deprecated fashion for a while.
I would be fine with only allowing removal of aliases to avoid confusion. It would be fine if trying to remove the original allocator name throws an error.

mcfadden8:
I assume that one desired pattern might be for the application to be able to delete the alias and then reuse it for a different allocator.  Let's say that it does this several times and let's also say that a couple of those allocators had leaked allocations.  When the resource manager shuts down, it will find and report those leaked allocators and I believe we will see duplicate names.

I don't know if this is a good idea or not, but should we consider having Umpire alter the names of the allocator whose name (alias) is being unregistered?  Maybe appending a sequence number to the end of the original name or perhaps something else?

MrBurmark:
Right this is why I'm imagining something more along the lines of a main name that is set at creation and can't be removed. Then in addition the aliases which can be removed. That way any reports can always use the main name which must be unique.

davidbeckingsale:
This really useful, thanks @mcfadden8 and @MrBurmark. I think I will go ahead and implement add/removeAlias, and keep register for now but mark it deprecated. So the current pairing would be registerAllocator/removeAlias, with the idea to move on to it being addAlias/removeAlias.

I will make sure that the "main" name cannot be removed, and that the "main" name would always be used for the reports.

davidbeckingsale:
LGTM

src/umpire/ResourceManager.hpp
```diff
@@ -115,8 +115,39 @@ class ResourceManager {
    * \param name Name to register Allocator with.
    * \param allocator Allocator to register.
    */
+  UMPIRE_DEPRECATE("addAlias should be used instead")
   void registerAllocator(const std::string& name, Allocator allocator);
 
+  /*!
+   * \brief Remove an Allocator alias.
+   *
+   * After calling, allocator can no longer by accessed by calling
```
mcfadden8, line 124:
```suggestion
   * After calling, allocator can no longer be accessed by calling
```

src/umpire/ResourceManager.hpp
```diff
@@ -115,8 +115,39 @@ class ResourceManager {
...
+   *
+   * \param name Name to deregister Allocator with.
+   * \param allocator Allocator to deregister.
+   */
+  void addAlias(const std::string& name, Allocator allocator);
+
+  /*!
+   * \brief Remove an Allocator alias.
+   *
+   * After calling, allocator can no longer by accessed by calling
```
mcfadden8, line 139:
```suggestion
   * After calling, allocator can no longer be accessed by calling
```

src/umpire/util/Macros.hpp
```diff
@@ -132,4 +132,30 @@
...
+#if defined(__has_cpp_attribute)
+#if __has_cpp_attribute(deprecated)
+#define UMPIRE_HAS_CXX_ATTRIBUTE_DEPRECATED 1
+#endif
+#endif
+#endif
+
+#if defined(UMPIRE_HAS_CXX_ATTRIBUTE_DEPRECATED)
+#define UMPIRE_DEPRECATE(Msg) [[deprecated(Msg)]]
+#define UMPIRE_DEPRECATE_ALIAS(Msg) [[deprecated(Msg)]]
```
mcfadden8, line 145: Fancy! Very nice.

mcfadden8 (approved):
Looks great!  Minor typo changes that you may change if you choose.

davidbeckingsale:
LGTM

tests/unit/resource_manager_tests.cpp
```diff
@@ -77,3 +79,38 @@ TEST(ResourceManager, getAllocatorInvalidId)
...
+  EXPECT_NE(alloc_alias.getName(), "HOST_ALIAS");
+
+  EXPECT_THROW({
+    rm.removeAlias("BLAH", alloc);
+  }, umpire::util::Exception);
+  EXPECT_NO_THROW({
+    rm.removeAlias("HOST_ALIAS", alloc);
+  });
+
+  alloc = rm.getAllocator("HOST");
```
kab163, line 105: Is this not needed? (see line 87)

tests/unit/resource_manager_tests.cpp
```diff
@@ -77,3 +79,38 @@ TEST(ResourceManager, getAllocatorInvalidId)
...
+  EXPECT_THROW({
+    rm.removeAlias("HOST", alloc);
+  }, umpire::util::Exception);
+
+  auto named_alloc = rm.makeAllocator<umpire::strategy::NamedAllocationStrategy>(
+    "NAMED_ALLOCATOR", alloc);
+
+  EXPECT_THROW({
```
kab163, line 113: Is this redundant because of line 98 (trying to remove an alias that doesn't exist) or is it ok because this tests an allocator built off another allocator?

kab163: Oh wait no, this is a property of NamedAllocationStrategy, huh..

tests/unit/resource_manager_tests.cpp
kab163: Oh, I see: "After calling, allocator can no longer be accessed by calling getAllocator(name)." - but what if it really doesn't :O

kab163:
What about the "isAllocatorRegistered" function? Should that be renamed?

davidbeckingsale:
> What about the "isAllocatorRegistered" function? Should that be renamed?

Great catch! It's now marked as deprecated.

davidbeckingsale:
LGTM

kab163 (approved):

PR #481, opened 2020-10-26
Feature/uberenv env

adrienbernede:
This PR is demonstration of how Uberenv can use spack environments.
This changes nothing to Umpire workflow, the only changes are in Uberenv itself, and in spack-configs which is now a submodule.


adrienbernede:
LGTM

adrienbernede:
LGTM

scripts/gitlab/build_and_test.sh
```diff
@@ -108,7 +108,7 @@ then
     $cmake_exe \
       -C ${hostconfig_path} \
       ${project_dir}
-    if ! $cmake_exe --build . -j; then
+    if ! $cmake_exe --build . -j 12; then
```
mcfadden8, line 111: I am not sure I understand this.  Can you elaborate?  Why "12"?

scripts/uberenv/uberenv.py
```diff
@@ -380,7 +383,7 @@ def clone_repo(self):
             spack_url = self.project_opts.get("spack_url", "https://github.com/spack/spack.git")
             spack_branch = self.project_opts.get("spack_branch", "develop")
 
-            clone_cmd =  "git {0} clone --single-branch --depth=1 -b {1} {2}".format(clone_opts, spack_branch,spack_url)
+            clone_cmd =  "git {0} clone -b {1} {2}".format(clone_opts, spack_branch,spack_url)
```
mcfadden8, line 386: Won't this change result in bringing over a larger image?  What is the purpose of removing `--single-branch --depth=1`?

scripts/gitlab/build_and_test.sh
adrienbernede: First the main drawback: 12 is a value "optimized" for quartz.
"-j" maximize the number of threads to use. However, on quartz, we share a node for all builds. Throwing so many threads was counter productive, and could even have lead to shared memory issues. By reducing the maximum number of threads, I got the CI on quartz to run faster, and didn’t have to increase the allocation time.

adrienbernede: Quartz nodes have 2 sockets of 18 cores. Each cores supports 2 hyperthreads, and the 18 cores are organized by groups of 6 (sort of). Hence the 12 value. However, actually pining those threads to cores is again not optimal (per my tests). So I leave the thread orchestration to the OS.

scripts/uberenv/uberenv.py
adrienbernede: That may be a mistake of mine! Thanks.

adrienbernede:
LGTM

davidbeckingsale:
Hey @adrienbernede - is this something we should clean up and get merged?

adrienbernede:
@davidbeckingsale I used Umpire as a pilot project to test using Uberenv with spack environment.
The corresponding branch in uberenv has been waiting for review for a long time, and then it was decided to maintain also the old spack-patching mechanism.
Since then, I haven't had much time to clean and test this on Uberenv side.

Also, you can see this branch has only a few changes outside of the uberenv.py script. This script would now be provided by submodule, so updating this branch would mean basically emptying it.

I will close it, and we could recreate one from the appropriate branch in Uberenv when we are ready to dedicate some time to it.

davidbeckingsale:
Thanks for the explanation!

PR #484, opened 2020-11-03
Add isAllocator method for ids

davidbeckingsale:


davidbeckingsale:
cc @mclarsen

davidbeckingsale:
LGTM

kab163 (approved):
yay!

davidbeckingsale:
LGTM

src/umpire/interface/c_fortran/wrapResourceManager.h
```diff
@@ -148,12 +148,15 @@ umpire_allocator * umpire_resourcemanager_get_allocator_for_ptr(
     umpire_resourcemanager * self, void * ptr,
     umpire_allocator * SHC_rv);
 
-bool umpire_resourcemanager_is_allocator(umpire_resourcemanager * self,
-    const char * name);
+bool umpire_resourcemanager_is_allocator_name(
```
mcfadden8, line 151: Just a note for those that happen to be actively using `umpire_resourcemanager_is_allocator`, this is a signature change for this function which means that it is an API change at least for FORTRAN.  We may want to make sure and release note this.   Does this mean we need to also update the major release number in the next release?

mcfadden8 (approved):
Overall, this looks great.  I only had a question about whether additional release notes should be considered for the next release cycle.

davidbeckingsale:
LGTM

davidbeckingsale:
> Overall, this looks great. I only had a question about whether additional release notes should be considered for the next release cycle.

I updated the CHANGELOG so it notes that change, and we should put it in the release notes too

PR #490, opened 2020-11-10
Add NUMA build to CI

davidbeckingsale:


davidbeckingsale:
LGTM

davidbeckingsale:
LGTM

davidbeckingsale:
LGTM

kab163 (approved):
Pretty straight forward. Looks good

src/umpire/ResourceManager.cpp
```diff
@@ -20,6 +20,9 @@
 #include "umpire/strategy/DynamicPool.hpp"
 #include "umpire/strategy/FixedPool.hpp"
 #include "umpire/strategy/ZeroByteHandler.hpp"
+#if defined(UMPIRE_ENABLE_NUMA)
+#include "umpire/strategy/NumaPolicy.hpp"
+#endif
```
mcfadden8, line 25: It seems strange (to me) that the only change made to this file was to add a guarded include of NumaPolicy.hpp.  Is there someplace else further down in this file that should also be guarded?

davidbeckingsale: I think this got accidentally removed at some point. The `move` operation uses this class.

mcfadden8: Ah.  I guess we wern't building with NUMA very much.  Glad it has been added to CI!

mcfadden8 (approved):

PR #496, opened 2020-11-12
Use CMake 3.18.0 in CI

davidbeckingsale:


davidbeckingsale:
LGTM

davidbeckingsale:
LGTM

.gitlab/lassen-jobs.yml
```diff
@@ -23,7 +23,7 @@ clang_10_0_1_mpi (build and test on lassen):
 
 gcc_8_3_1:
   variables:
-    SPEC: "+fortran%gcc@8.3.1"
+    SPEC: "+fortran%gcc@8.3.1 ^cmake@3.14.5"
```
mcfadden8, line 26: How does this cause CMake 3.18.0 to be used?  Other than the CHANGELOG, I don't see any other changes being made.

kab163: I think it's a Spack spec syntax thing. It's hard to find the right documentation on it (imho) but it seems to be something that lets you handle dependencies (https://spack.readthedocs.io/en/latest/environments.html#speclists-as-constraints)

kab163: But I'm confused why this one change will make the CI then always use cmake 3.18 for Lassen - what about the other jobs in the file?

kab163: Lol I think Marty and I have the same confusion (it's late mkay)

davidbeckingsale:
@adrienbernede - I was expecting Spack to pick up the newer CMake by default, but that doesn't seem to be the case. Any thoughts?

.gitlab/lassen-jobs.yml
adrienbernede: I think @davidbeckingsale intent was not to use cmake@3.18.0 for this job specifically, but set it as a default for the other jobs. The config change happens in the Radiuss-CI submodule.

adrienbernede:
@davidbeckingsale can you be more specific?
In your CI run, cmake@3.18.0 was picked up: https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/127134#L122

davidbeckingsale:
> @davidbeckingsale can you be more specific?
> In your CI run, cmake@3.18.0 was picked up: https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/127134#L122

Okay - my change did have the intended effect. However, it's actually still using CMake 3.14.5: https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/127134#L152 I'll do some more investigating.

adrienbernede:
My bad! That’s because we explicitly set which CMake to use in the build_and_test.sh script instead of using the one defined in Spack.

adrienbernede:
Note: The host-config file should contain the path to the CMake binary defined by spack. We could retrieve it in the ci script.

davidbeckingsale:
LGTM

mcfadden8 (approved):
Looks great!

Thanks @davidbeckingsale

davidbeckingsale:
I just hope it works 😬

davidbeckingsale:
LGTM

kab163 (approved):

davidbeckingsale:
LGTM

PR #503, opened 2020-11-24
Feature/um 653 shared memory implementation

mcfadden8:
The MPI and Boost prototyping is complete.  Here is where the actual implementation work will be performed.


mcfadden8:
Being replaced by PR #553

PR #507, opened 2020-12-01
Remove internal tracking

davidbeckingsale:


davidbeckingsale:
LGTM

davidbeckingsale:
LGTM

src/umpire/strategy/AllocationStrategy.hpp
```diff
@@ -51,7 +52,11 @@ class AllocationStrategy {
    *
    * \param ptr Pointer to free.
    */
-  virtual void deallocate(void* ptr) = 0;
+  virtual void deallocate(void* ptr, std::size_t size=0) = 0;
+
+  void* allocate_internal(std::size_t bytes);
```
davidbeckingsale, line 57: Any thoughts on the naming of these internal/impl methods?

mcfadden8: Can they be hiddning behind pimpl?

src/umpire/Allocator.cpp
```diff
@@ -35,7 +38,7 @@ std::size_t Allocator::getSize(void* ptr) const
 
 std::size_t Allocator::getHighWatermark() const noexcept
 {
-  return m_allocator->getHighWatermark();
+ return m_allocator->getHighWatermark();
```
kab163, line 41:
```suggestion
  return m_allocator->getHighWatermark();
```

kab163: I'm being picky, but I want to add back the space here lol

src/umpire/Allocator.inl
```diff
@@ -46,7 +54,10 @@ inline void Allocator::deallocate(void* ptr)
     UMPIRE_LOG(Info, "Deallocating a null pointer (This behavior is intentionally allowed and ignored)");
     return;
   } else {
-    m_allocator->deallocate(ptr);
+    auto record = deregisterAllocation(ptr, m_allocator);
```
mcfadden8, line 57: Does this work when `m_tracking == false`?

src/umpire/ResourceManager.hpp
```diff
@@ -108,7 +112,7 @@ class ResourceManager {
   /*!
    * \brief Construct a new Allocator.
    */
-  template <typename Strategy, bool introspection = true, typename... Args>
+  template <typename Strategy, bool introspection = false, typename... Args>
```
mcfadden8, line 115: I didn't know we were changing the default for enabling introspection.  Was this intentional?

src/umpire/op/HostReallocateOperation.cpp
```diff
@@ -45,7 +45,7 @@ void HostReallocateOperation::transform(
     }
 
     ResourceManager::getInstance().registerAllocation(
-        *new_ptr, {*new_ptr, new_size, allocator});
+        *new_ptr, {*new_ptr, new_size, new_allocation->strategy});
```
mcfadden8, line 48: Shouldn't this be conditional upon whether introspection is enabled?

src/umpire/resource/CudaConstantMemoryResource.cu
```diff
@@ -52,33 +52,20 @@ void* CudaConstantMemoryResource::allocate(std::size_t bytes)
...
-  if (m_current_size > m_highwatermark)
-    m_highwatermark = m_current_size;
-
   UMPIRE_LOG(Debug, "(bytes=" << bytes << ") returning " << ret);
 
   return ret;
 }
 
-void CudaConstantMemoryResource::deallocate(void* ptr)
+void CudaConstantMemoryResource::deallocate(void* ptr, std::size_t size)
```
mcfadden8, line 60: How does `size` get set when/if introspection is disabled?

src/umpire/resource/DefaultMemoryResource.hpp
```diff
@@ -20,7 +20,9 @@ namespace resource {
  * allocate and deallocate memory.
  */
 template <typename _allocator>
-class DefaultMemoryResource : public MemoryResource {
+class DefaultMemoryResource : 
+  public MemoryResource
+{
```
kab163, line 25:
```suggestion
class DefaultMemoryResource : public MemoryResource {
```

kab163: (Again, being picky, but I think most of our other inheritance lines like this have been structured this way instead)

src/umpire/strategy/mixins/Inspector.cpp
```diff
@@ -12,41 +12,34 @@ namespace umpire {
...
 void
 Inspector::registerAllocation(
     void* ptr,
     std::size_t size,
-    strategy::AllocationStrategy* strategy)
+    strategy::AllocationStrategy* s) 
 {
-  m_current_size += size;
-  m_allocation_count++;
+  s->m_current_size += size;
```
mcfadden8, line 21: Directly changing state of another object seems sneaky to me.  I appreciate why it is being done this way, but is there perhaps a different way to accomplish the same thing without having to worry about the strategy object and inspector object from inadvertantly overwriting each other?

src/umpire/util/Macros.hpp
```diff
@@ -143,4 +143,12 @@
...
 #endif
 
+#if 0
+#define UMPIRE_INTERNAL_TRACK(p, s) registerAllocation(p, s, this);
+#define UMPIRE_INTERNAL_UNTRACK(p) deregisterAllocation(p, this);
+#else 
+#define UMPIRE_INTERNAL_TRACK(p, s)
+#define UMPIRE_INTERNAL_UNTRACK(p) umpire::util::AllocationRecord{};
+#endif
+
```
mcfadden8, line 153: Are these still needed?

tests/unit/resource/resource_tests.cpp
```diff
@@ -46,5 +46,5 @@ TEST(DefaultMemoryResource, AllocateDeallocate)
   double* pointer = (double*)alloc->allocate(10 * sizeof(double));
   ASSERT_NE(pointer, nullptr);
 
-  alloc->deallocate(pointer);
+  alloc->deallocate(pointer, 0);
```
mcfadden8, line 49: Why isn't this `alloc->deallocate(pointer);`?

mcfadden8 (changes requested):
Overall, this looks good and I really like how things seem to have simplified.  I'm still not sure I follow how things will work for Allocators that do not have tracking enabled which is the source of a lot of my questions.

tests/unit/resource/resource_tests.cpp
mcfadden8, line 49: What are the semantics of `size=0`?  Are they documented somewhere?

src/umpire/resource/MemoryResource.hpp
```diff
@@ -53,28 +53,7 @@ class MemoryResource : public strategy::AllocationStrategy {
...
-   * \brief Return the memory high watermark for this MemoryResource.
-   *
-   * This is the largest amount of memory allocated by this Allocator. Note
-   * that this may be larger than the largest value returned by
-   * getCurrentSize.
-   *
-   * \return Memory high watermark.
-   */
-  virtual std::size_t getHighWatermark() const noexcept override = 0;
+  virtual void deallocate(void* ptr, std::size_t size) override = 0;
```
mcfadden8, line 56: Please add documentation for `size` parameter above.  Specifically, the semantics for when `size == 0`

mcfadden8 (commented):
Looks great.  I've reviewed and added comments and some suggestions (mostly minor).

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -12,6 +12,7 @@
 #include "umpire/config.hpp"
 #include "umpire/ResourceManager.hpp"
 #include "umpire/Allocator.hpp"
+#include "umpire/strategy/FixedPool.hpp"
```
kab163, line 15: Why was this added?

mcfadden8 (approved):
Looks good.  Thanks.

PR #511, opened 2020-12-11
Bugfix/um 801 hip gitlab ci error (+UM-800-windows-ci-error)

kab163:
- [x] Need the radiusssss PR to go through (scripts are updated there too)

(**NOTE**: Needed to combine this with the windows error ticket so that they can both pass all the tests when fixed. JIRA ticket will be updated.)

### DIFF
--- a/src/umpire/util/io.cpp
+++ b/src/umpire/util/io.cpp
@@ -19,10 +19,6 @@
 #include "umpire/util/Macros.hpp"
 #include "umpire/util/OutputBuffer.hpp"
 
-#if defined(_MSC_VER)
-#define UMPIRE_ENABLE_FILESYSTEM
-#endif
-
 #if defined(UMPIRE_ENABLE_FILESYSTEM)
 #include <filesystem>
 #else
--- a/src/umpire/util/io.cpp
+++ b/src/umpire/util/io.cpp
@@ -19,6 +19,10 @@
 #include "umpire/util/Macros.hpp"
 #include "umpire/util/OutputBuffer.hpp"
 
+#if defined(_MSC_VER)
+#define UMPIRE_ENABLE_FILESYSTEM
+#endif
+
 #if defined(UMPIRE_ENABLE_FILESYSTEM)
 #include <filesystem>
 #else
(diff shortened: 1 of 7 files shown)


kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

src/umpire/util/io.cpp
```diff
@@ -19,6 +19,10 @@
 #include "umpire/util/Macros.hpp"
 #include "umpire/util/OutputBuffer.hpp"
 
+#if defined(_MSC_VER)
+#define UMPIRE_ENABLE_FILESYSTEM
```
mcfadden8, line 23: I'm not sure I understand this.  Is `UMPIRE_ENABLE_FILESYSTEM` one of the macros controlled via `cmake`?  If so, shouldn't this check be done there?

davidbeckingsale: Yes, it should. For some reason the check is not working and I think @kab163 just added this to test some things out.

kab163: That is a good question. When looking at this issue, the check here [https://github.com/LLNL/Umpire/blob/6068966b966de1560cb3783e47d6d6fcbc70cd62/cmake/SetupCompilerFlags.cmake#L14](url) is supposed to turn that flag on when needed. However, it suddenly stopped working for some reason. I tried to test that check on the dugout, but it would pass fine (I could get this check to work fine if cmake 3.19 was used and the policy set - see change in CMakeLists). That wouldn't work with the Windows CI. So, until that is sorted out, this forces that flag for Windows and the test can continue and hopefully pass. It's a temporary thing until I can figure out a better solution.

kab163: It's probably better explained here (https://rzlc.llnl.gov/jira/browse/UM-800)

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

mcfadden8 (approved):

davidbeckingsale (approved):

PR #517, opened 2020-12-18
Task/um 806 improve benchmarking

kab163:
PART1 - This is the benchmark for the first part (UM-806) of the JIRA ticket, finding Umpire overhead (UM-372)
- [ ] Document process (in JIRA) for updating cmake flag for ENABLE_DEVELOPER_BENCHMARKS (i.e. UMPIRE_ENABLE_BENCHMARKS may soon become UMPIRE_ENABLE_DEVELOPER_BENCHMARKS in a future ticket)
     ---Note: This will also affect how the No-Op memory resource files are included in the cmake and how the resource is registered in the RM
- [x] If No-Op stuff is NOT for users, then is it needed on CHANGELOG? Maybe add a note?

See [output](https://rzlc.llnl.gov/jira/secure/attachment/17936/no-op_stress_test_output.txt) on JIRA ticket.


kab163:
LGTM

examples/noop_malloc.cpp
```diff
@@ -0,0 +1,18 @@
+//////////////////////////////////////////////////////////////////////////////
```
mcfadden8, line 1: I like the idea of keeping the example.  But I would call it noop_allocator.cpp instead of noop_malloc.cpp.

src/umpire/resource/MemoryResourceTypes.hpp
```diff
@@ -30,7 +30,8 @@ enum MemoryResourceType {
   Pinned,
   Constant,
   File,
-  Unknown
+  Unknown,
```
mcfadden8, line 33: Personallhy, I kind of prefer to either make `Unknown` be the last entry or the first entry in the enumerations.  I really can't tell you why though :-).

src/umpire/resource/NoOpMemoryResource.cpp
```diff
@@ -0,0 +1,67 @@
...
+                                       const std::string& name, int id,
+                                       MemoryResourceTraits traits)
+    : MemoryResource{name, id, traits}, m_platform{platform}
+{
+  m_count = (UINT64_C(1)<<48);
+}
+
+NoOpMemoryResource::~NoOpMemoryResource()
+{
+  m_count = (UINT64_C(1)<<48);
```
mcfadden8, line 29: I think the entire destructor may be removed.  No allocations, so no leaks.  New instantiations will always call the constructor.

src/umpire/resource/NoOpMemoryResource.cpp
```diff
@@ -0,0 +1,67 @@
...
+namespace umpire {
+namespace resource {
+
+
+NoOpMemoryResource::NoOpMemoryResource(Platform platform,
+                                       const std::string& name, int id,
+                                       MemoryResourceTraits traits)
+    : MemoryResource{name, id, traits}, m_platform{platform}
+{
+  m_count = (UINT64_C(1)<<48);
```
mcfadden8, line 24: This may be done C++ 11 style and just moved to the place it was declared in the header file (see below).

```suggestion
```

src/umpire/resource/NoOpMemoryResource.hpp
```diff
@@ -0,0 +1,70 @@
...
+  /*!
+   * \brief Construct a new NoOpMemoryResource and initialize a counter
+   */
+  NoOpMemoryResource(Platform platform, const std::string& name, int id,
+                     MemoryResourceTraits traits);
+
+  /*!
+   * \brief Resets allocation counter.
+   */
+  ~NoOpMemoryResource();
```
mcfadden8, line 38: I would remove the desrtructor or leave its implementation empty.

src/umpire/resource/NoOpMemoryResource.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  bool isAccessibleFrom(Platform p) noexcept;
+  
+  Platform getPlatform() noexcept;
+
+ protected:
+  Platform m_platform;
+
+ private:
+  size_t m_count;
```
mcfadden8, line 64: Please qualify size_t and I suggest that the initialization may be done here instead of in the constructor.
```suggestion
  std::size_t m_count{(UINT64_C(1)<<48)};
```

mcfadden8 (commented):
Overall this looks great.  I made some minor comments.  I did not review the actual benchmark itself.  I think that you and @davidbeckingsale can tune that as we start evaluating test results.

src/umpire/resource/NoOpMemoryResource.cpp
kab163: AHHH you're totally right haha - Old habits :p

kab163:
> Overall this looks great. I made some minor comments. I did not review the actual benchmark itself. I think that you and @davidbeckingsale can tune that as we start evaluating test results.

Thanks, Marty!! These are great! Thanks for your input :)

src/umpire/resource/MemoryResourceTypes.hpp
kab163: Nah, I get it. I feel that is kinda better too lol

src/umpire/resource/NoOpMemoryResource.hpp
kab163: Great idea

src/umpire/resource/NoOpResourceFactory.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "umpire/util/make_unique.hpp"
+#include "umpire/util/detect_vendor.hpp"
+
+namespace umpire {
+namespace resource {
+
+bool NoOpResourceFactory::isValidMemoryResourceFor(
+    const std::string& name) noexcept
+{
+  if (name.find("No_Op") == std::string::npos)
```
kab163, line 20: This should actually be "NO_OP", right?

kab163: JK, keeping as No_Op

kab163:
LGTM

src/umpire/resource/NoOpMemoryResource.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+  bool isAccessibleFrom(Platform p) noexcept;
+  
+  Platform getPlatform() noexcept;
+
+ protected:
+  Platform m_platform;
+
+ private:
+  std::size_t m_count = (UINT64_C(1)<<48);
```
mcfadden8, line 65:
```suggestion
  std::size_t m_count{ (UINT64_C(1)<<48) };
```

src/umpire/resource/NoOpResourceFactory.cpp
mcfadden8, line 20: I have seen (I think) "NO_OP" and "No_Op", I think it needs to be one or the other.  I suggest either "NO_OP" or "NOOP".

examples/noop_allocator.cpp
```diff
@@ -0,0 +1,18 @@
...
+//
+// SPDX-License-Identifier: (MIT)
+//////////////////////////////////////////////////////////////////////////////
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Allocator.hpp"
+#include <iostream>
+
+int main() {
+  auto &rm = umpire::ResourceManager::getInstance();
+  auto alloc = rm.getAllocator("NO_OP");
```
mcfadden8, line 13: NoOpResorceFactory.cpp currently is using "No_Op".  This string needs to be the same.  I suggest either using "NO_OP" or potentially "NOOP".

src/umpire/resource/NoOpResourceFactory.cpp
kab163: I tried to use "NO_OP" but it gave me an exception with "getDefaultTraitsForResource MemoryResource NO_OP not found" - so I switched it back lol i thought that would have worked (?)

src/umpire/strategy/ZeroByteHandler.cpp
```diff
@@ -35,11 +35,7 @@ void* ZeroByteHandler::allocate(std::size_t bytes)
 
 void ZeroByteHandler::deallocate(void* ptr)
 {
-  if (!m_zero_byte_pool)
-    m_zero_byte_pool = static_cast<FixedPool*>(
-        ResourceManager::getInstance().getZeroByteAllocator());
-
-  if (m_zero_byte_pool->pointerIsFromPool(ptr)) {
+  if (m_zero_byte_pool && m_zero_byte_pool->pointerIsFromPool(ptr)) {
```
mcfadden8, line 38: @davidbeckingsale, while this fix works, I think further examination of the FixedBlockPool manager may be necessary.  It appeared to me that the FixedBlockPool manager may have missed some initialization of some of the `m_pool` elements which caused `m_zero_byte_pool->pointerIsFromPool(ptr)` to return true even when the `m_pool` was created but had no allocations present.

src/umpire/resource/NoOpResourceFactory.cpp
mcfadden8:
```suggestion
  if (name.find("NO_OP") != std::string::npos)
```

kab163: JK, fixing it now!

kab163:
LGTM

src/umpire/strategy/ZeroByteHandler.cpp
mcfadden8: This change was not necessary.  It was caused by something else.

mcfadden8:
LGTM

kab163:
LGTM

kab163:
LGTM

kab163:
[output.txt](https://github.com/LLNL/Umpire/files/5748805/output.txt)
Current output from the no-op_benchmark

benchmarks/allocator_no-op_benchmarks.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+  NoOpAllocatorBenchmark() {}
+
+  void SetUp(benchmark::State&) {
+    auto& rm = umpire::ResourceManager::getInstance();
+    alloc = rm.getAllocator("NO_OP");
+  }
+ 
+  void TearDown(benchmark::State&) {
+    alloc.getAllocationStrategy()->release();
```
mcfadden8, line 29: This line is no longer needed since you are dealing directly with the resource allocator.  `release()` is only meaningful for pools.

benchmarks/allocator_no-op_benchmarks.cpp
```diff
@@ -0,0 +1,68 @@
...
+  void NoOpAllocate(benchmark::State& st) {
+    size = static_cast<std::size_t>(st.range(0));
+    while (st.KeepRunning()) {
+      allocation = alloc.allocate(size);
+    }
+  }
+
+  void NoOpDeallocate(benchmark::State& st) {
+    while (st.KeepRunning()) {
+      alloc.deallocate(allocation);
```
mcfadden8, line 41: I'm not sure why this doesn't throw.  `allocation` is never assigned to something that was previously 'allocated'.  So I would assume that our allocation tracker should throw an exeption saying that you're trying to deallocate a pointer that is not currently allocated.

benchmarks/allocator_no-op_benchmarks.cpp
kab163: Yea I thought you might say that lol it feels weird to not have to release memory (but yea, I'll take that out, thanks)

benchmarks/allocator_no-op_benchmarks.cpp
mcfadden8: IIRC, calling deallocate with an address that was not previously allocated should trigger our tracker here: https://github.com/LLNL/Umpire/blob/6d06cbce8c75bb3d970af612a770f023629f9d09/src/umpire/strategy/mixins/Inspector.cpp#L49

kab163:
Most up-to-date benchmark output:
[output.txt](https://github.com/LLNL/Umpire/files/5755521/output.txt)

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

kab163:
[Output](https://rzlc.llnl.gov/jira/secure/attachment/17935/no-op_stress_test_output.txt) of the stress test is on JIRA

kab163:
LGTM

examples/noop_allocator.cpp
```diff
@@ -0,0 +1,18 @@
+//////////////////////////////////////////////////////////////////////////////
```
davidbeckingsale, line 1: I don't think we want this example file anymore do we?

kab163:
LGTM

kab163:
LGTM

benchmarks/no-op_stress_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  std::cout << "    SHUFFLE:" << std::endl; 
+  std::cout << "    alloc: " << (time[0]/double(NUM_ITER)*CONVERT) << "(us)" << std::endl;
+  std::cout << "    dealloc: " << (time[1]/double(NUM_ITER)*CONVERT) << "(us)" << std::endl << std::endl;
+}
+
+int main(int, char**) {
+  std::mt19937 gen(ALLOCATIONS);
+  std::uniform_int_distribution<std::size_t> dist(64, 4096);
+  std::size_t size = dist(gen);
```
davidbeckingsale, line 96: Size is unnecessary since it's not used by the NO_OP allocator.

davidbeckingsale, line 96: Can you change this to use a fixed size of 4096. I think we want to keep as many variables fixed as possible to try and isolate what this is benchmarking.

davidbeckingsale (changes requested):
One quick change!

benchmarks/no-op_stress_test.cpp
kab163: Done!

kab163:
LGTM

mcfadden8 (approved):
Thanks!

davidbeckingsale (approved):

PR #520, opened 2021-01-05
update documentation on nullptr deallocation behavior

kab163:


davidbeckingsale (approved):

kab163:
LGTM

kab163:
LGTM

src/umpire/Allocator.hpp
```diff
@@ -55,10 +55,12 @@ class Allocator {
    * \brief Free the memory at ptr.
    *
    * This method will throw an umpire::Exception if ptr was not allocated
-   * using this Allocator. If you need to deallocate memory allocated by an
-   * unknown object, use the ResourceManager::deallocate method.
+   * using this Allocator. If the pointer hasn't been used (a nullptr),
```
mcfadden8, line 58: Being a nullptr, does not really indicate whether the pointer has been used.  It simply means that the pointer was set to nullptr.  I suggest the following:

```suggestion
   * using this Allocator. If the value of the pointer is set to nullptr, 
```

src/umpire/Allocator.inl
```diff
@@ -43,7 +43,7 @@ inline void Allocator::deallocate(void* ptr)
   UMPIRE_LOG(Debug, "(" << ptr << ")");
 
   if (!ptr) {
-    UMPIRE_LOG(Info, "Deallocating a null pointer");
+    UMPIRE_LOG(Info, "Deallocating a null pointer (This behavior is purposefully allowed!)");
```
mcfadden8, line 46:
```suggestion
    UMPIRE_LOG(Info, "Deallocating a null pointer (This behavior is intentionally allowed and ignored)");
```

src/umpire/Allocator.hpp
kab163: Sounds good. I was trying to get the wording just right... words are hard

mcfadden8 (approved):
Thanks Kristi,

This looks good to me.

src/umpire/Allocator.inl
kab163: I think that is more clear, thanks!

kab163:
LGTM

PR #522, opened 2021-01-12
Task/um 807 benchmark size time

kab163:


benchmarks/allocation_size_stress_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Allocator.hpp"
+
+#define ALLOCATIONS 1000
+#define CONVERT 1000000 //convert sec (s) to microsec (us)
+#define NUM_ITER 3
+
+#pragma GCC push_options
+#pragma GCC optimize ("-O0")
```
davidbeckingsale, line 22: 😱

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

benchmarks/allocation_size_benchmarks.cpp
```diff
@@ -0,0 +1,206 @@
...
+class HipMallocManaged : public ResourceAllocator<umpire::alloc::HipMallocManagedAllocator> {};
+BENCHMARK_DEFINE_F(HipMallocManaged, hipMallocManaged)(benchmark::State& st) { allocation(st); }
+BENCHMARK_DEFINE_F(HipMallocManaged, hipFree)(benchmark::State& st)   { deallocation(st); }
+
+class HipPinned : public ResourceAllocator<umpire::alloc::HipPinnedAllocator> {};
+BENCHMARK_DEFINE_F(HipPinned, hipMallocHost)(benchmark::State& st) { allocation(st); }
+BENCHMARK_DEFINE_F(HipPinned, hipFreeHost)(benchmark::State& st)   { deallocation(st); }
+#endif
+
+template <umpire::resource::MemoryResourceType Resource>
```
davidbeckingsale, line 119: We don't need these `MemoryResourceAllocator` benchmarks in this file.

davidbeckingsale (changes requested):

benchmarks/allocation_size_benchmarks.cpp
kab163: Gasp! Then are we keeping those tests at all? (Because the 3rd benchmark is only for pools)

davidbeckingsale: I think the value has gone since they were used to compare overhead, but we now have the no-op benchmark for that.

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

benchmarks/vendor_allocator_benchmarks.cpp
```diff
@@ -0,0 +1,183 @@
...
+        for (std::size_t j{0}; j < Max_Allocations; j++)
+          deallocate(m_allocations[j]);
+        i = 0;
+        st.ResumeTiming();
+      }
+      m_allocations[i++] = allocate(size);
+    }
+
+    // This says that we process with the rate of state.range(0) bytes every iteration:
+    st.counters["BytesProcessed"] = benchmark::Counter(st.range(0), benchmark::Counter::kIsIterationInvariantRate, benchmark::Counter::OneK::kIs1024);
```
davidbeckingsale, line 62: Do you think we should convert the number to Kb or Mb here?

benchmarks/vendor_allocator_benchmarks.cpp
```diff
@@ -0,0 +1,183 @@
...
+static const int RangeLow{1048576}; //2^20 or ~1MB
+static const int RangeHi{268435456}; //2^28 or ~256MB
+
+/*
+ * Allocate either LARGE (about 12GB), MEDIUM (about 6GB)
+ * or SMALL (about 1GB) for benchmark measurements.
+ */
+//#define LARGE 12000000000
+//#define MEDIUM 6000000000
+#define SMALL 1000000000
```
davidbeckingsale, line 33: I think we want to default to using LARGE

benchmarks/vendor_allocator_benchmarks.cpp
kab163: We decided not to do any conversions here. @mcfadden8 any thoughts? (When I divided by 1000 to get MB, the units that google benchmark reported in were confusing anyways, so it was almost more clear to leave it as is)

kab163:
LGTM

kab163:
LGTM

benchmarks/vendor_allocator_benchmarks.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include "umpire/alloc/CudaMallocManagedAllocator.hpp"
+#include "umpire/alloc/CudaPinnedAllocator.hpp"
+#endif
+#if defined(UMPIRE_ENABLE_HIP)
+#include "umpire/alloc/HipMallocAllocator.hpp"
+#include "umpire/alloc/HipMallocManagedAllocator.hpp"
+#include "umpire/alloc/HipPinnedAllocator.hpp"
+#endif
+
+static const int RangeLow{1048576}; //2^20 or ~1MB
```
davidbeckingsale, line 24: Can the smallest allocation be smaller than 1 meg, maybe something like 1 kb.

davidbeckingsale (changes requested):

benchmarks/vendor_allocator_benchmarks.cpp
kab163: I could definitely do that, but as I test it on lassen, it takes longer than 2 hours to complete a run of the benchmark when I  have ENABLE_CUDA on. If it's just the host, I can do it within the timeframe of the default lalloc job. (For the job that I ran, it ranged from 1024 to 268435456 with a range multiplier of 4 as before) Is that ok?

kab163: [cuda-10-1_vendor_allocator_benchmarks_1kBto1GB.txt](https://github.com/LLNL/Umpire/files/5921037/cuda-10-1_vendor_allocator_benchmarks_1kBto1GB.txt)

(For reference)

mcfadden8 (approved):

kab163:
LGTM

davidbeckingsale (approved):

PR #526, opened 2021-01-20
python -> python3 binary name

adrienbernede:
Python binary has been removed in favor of explicit python3 or python2.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -48,6 +48,9 @@ stages:
   - c_build_and_test
 #  - multi_project
 
+# This jobs creates a python env and caches it.
+# Note: The cache needs to be removed manually in CI if the python environment
+# is changed, by clicking "CI/CD - Pipelines - Clear Runner Caches".
 configure_python:
   variables:
     GIT_STRATEGY: none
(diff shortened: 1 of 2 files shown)


adrienbernede:
LGTM

davidbeckingsale (approved):

davidbeckingsale:
Thanks for the quick fix @adrienbernede !

adrienbernede:
@davidbeckingsale  I had to clean the cache.
That’s because our cache key "venv" is the same for all branches. So in the rare event of a change in the python env, the cache has to be removed in order for the environment to be updated.
The alternative is to force one cache per branch, but then we lose the benefit of caching because we usually run Gitlab CI only once or twice per branch. Still, that’s more robust.
What do you think?

davidbeckingsale:
> @davidbeckingsale I had to clean the cache.
> That’s because our cache key "venv" is the same for all branches. So in the rare event of a change in the python env, the cache has to be removed in order for the environment to be updated.
> The alternative is to force one cache per branch, but then we lose the benefit of caching because we usually run Gitlab CI only once or twice per branch. Still, that’s more robust.
> What do you think?

I think using the same cache is a good idea. If we have to manually clean the cache occasionally I don't think that's a problem.

adrienbernede:
LGTM

PR #531, opened 2021-01-26
Story/um 690 thread sanitizer

kab163:


.gitlab/lassen-jobs.yml
```diff
@@ -109,6 +109,11 @@ clang_9_0_0_memleak (build and test on lassen):
     ASAN_OPTIONS: "detect_leaks=1"
   extends: .build_and_test_on_lassen
 
+clang_9_0_0_datarace (build and test on lassen):
+  variables:
+    SPEC: "%clang@9.0.0 cxxflags=-fsanitize=thread"
```
kab163, line 114: Adding the "+sanitizer_tests" caused it to fail the sanitizer tests during "make test" - is that ok?

kab163:
LGTM

.gitlab/lassen-jobs.yml
davidbeckingsale: Yup, that option should maybe be renamed - it's only for the memory sanitizer tests, and requires the asan compiler flag.

src/umpire/strategy/mixins/Inspector.cpp
```diff
@@ -29,7 +29,7 @@ Inspector::registerAllocation(
   m_allocation_count++;
 
   if (m_current_size > m_high_watermark) {
-    m_high_watermark = m_current_size;
+    m_high_watermark.store(m_current_size, std::memory_order_relaxed);
```
davidbeckingsale, line 32: I don't think we want to make these atomics - it's an unnecessary performance hit (but I understand it's required to shut up the analyzer). Is there an alternative?

davidbeckingsale (changes requested):

src/umpire/strategy/mixins/Inspector.cpp
kab163: Well we should talk about it this afternoon, because if it's a data race, then I should either fix it (and what that might be) or, if it's a known data race like the 'umpire_ver_5_found', then I can make the sanitizer ignore it explicitly.

kab163:
LGTM

kab163:
LGTM

davidbeckingsale (approved):

PR #535, opened 2021-02-08
Task/um 829 fixed pool release func

kab163:
PR includes:
- FixedPool release function 
- TYPED_TEST for testing release functions of different strategies
       
ReleaseTest description:
SizeLimiter is used to restrict the size of the memory for the test. Then, another allocator is created from the SizeLimiter to allocate pools of memory and max-out the available memory. That memory is then deallocated. If I try to use the SizeLimiter allocator without first calling release, it will fail. When I call the release function, the memory is properly cleaned up and I am able to use the SizeLimiter allocator again.

(Realized we needed this when developing for UM-808 and FixedPool was having issues.)


kab163:
LGTM

src/umpire/strategy/FixedPool.cpp
```diff
@@ -194,6 +194,17 @@ void FixedPool::deallocate(void* ptr)
...
 
+void FixedPool::release() 
+{ 
+  for (auto& p : m_pool) {
+    if (m_obj_per_pool == p.num_avail) {
+      p.strategy->deallocate(p.data);
+      std::free(p.avail);
+    } 
+  }
+  m_pool.clear();
```
davidbeckingsale, line 205: This shouldn't be `clear`, since you only want to remove the elements where `p.num_avail == m_obj_per_pool`. You could use `remove_if` to do this.

tests/integration/strategy_tests.cpp
```diff
@@ -276,6 +276,92 @@ TYPED_TEST(StrategyTest, getActualSize)
...
+
+TYPED_TEST_SUITE(ReleaseTest, ReleaseStrategies, );
+
+TYPED_TEST(ReleaseTest, ReleaseCheck)
+{
+  void* test[this->num_allocs];
+
+  for (int i = 0; i < this->num_allocs; i++) {
+    test[i] = this->m_allocator->allocate(this->max_alloc_size);
+  }
```
davidbeckingsale, line 353: I think it would be good to call `release` before the `deallocate` loop (which shouldn't do anything). This would have caught the issue I commented on above.

davidbeckingsale (changes requested):
Two quick changes, but this looks good in general.

src/umpire/strategy/FixedPool.cpp
kab163: Ah, makes sense. Thanks

mcfadden8 (approved):
With changes suggested by @davidbeckingsale, I think this is good to go.

Thanks!

kab163:
LGTM

kab163:
LGTM

kab163:
LGTM

davidbeckingsale (approved):

PR #539, opened 2021-02-19
Fix m_current_bytes accounting in DynamicPoolMap

davidbeckingsale:
Addresses an issue where the pool will never coalesce, since getCurrentSize will never return 0.

### DIFF
--- a/tests/integration/primary_pool_tests.cpp
+++ b/tests/integration/primary_pool_tests.cpp
@@ -590,7 +590,7 @@ TYPED_TEST(PrimaryPoolTest, heuristic_100_percent)
     ASSERT_EQ(dynamic_pool->getReleasableSize(), 0);
   }
 
-  // ASSERT_EQ(dynamic_pool->getBlocksInPool(), 4);
+  ASSERT_EQ(dynamic_pool->getBlocksInPool(), 4);
   ASSERT_NO_THROW({ alloc.deallocate(a[3]); }); // 25% releasable
   ASSERT_EQ(dynamic_pool->getBlocksInPool(), 4);
   ASSERT_NO_THROW({ alloc.deallocate(a[2]); }); // 50% releasable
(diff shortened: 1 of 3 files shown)


davidbeckingsale:
LGTM

kab163:
Did we test that this really solves the issue? I can't remember, but did we use the replay file and the rounded_bytes fix to check? Also, would it be worth it to add a gtest case that checks for this?

davidbeckingsale:
> Did we test that this really solves the issue? I can't remember, but did we use the replay file and the rounded_bytes fix to check? Also, would it be worth it to add a gtest case that checks for this?

Yup, I ran it against the replays and verified it fixed it. I'm not sure the best way to write a test. I think we eventually need a replay-based system for more complex issues like this.

kab163 (approved):
Looks good!

mcfadden8 (commented):
This looks good.  Does DynamicPoolList have the same problem?

We obviously don't seem to have a gtest for this particular failure since no tests were ever really failing.  Should we create one?

davidbeckingsale:
> This looks good. Does DynamicPoolList have the same problem?
> 
> We obviously don't seem to have a gtest for this particular failure since no tests were ever really failing. Should we create one?

I have added a test - verified that it failed for develop, and is fixed by this PR.

davidbeckingsale:
LGTM

tests/integration/primary_pool_tests.cpp
```diff
@@ -563,3 +563,40 @@ TYPED_TEST(PrimaryPoolTest, heuristic_75_percent)
...
+
+  void* a[4];
+  for (int i{0}; i < 4; ++i) {
+    ASSERT_NO_THROW({ a[i] = alloc.allocate(768); });
+    ASSERT_EQ(alloc.getActualSize(), (1024 * (i + 1)));
+    ASSERT_EQ(dynamic_pool->getBlocksInPool(), (i + 1));
+    ASSERT_EQ(dynamic_pool->getReleasableSize(), 0);
+  }
+
+  // ASSERT_EQ(dynamic_pool->getBlocksInPool(), 4);
```
kab163, line 593: Is this needed still?

davidbeckingsale: Whoops, removed the comment.

davidbeckingsale:
LGTM

mcfadden8 (approved):
Thanks!

PR #545, opened 2021-03-03
Minor cuda build fixes

gzagaris:
The code was attempting to use the host compiler b/c
the cuda dependency was not specified leading to
compilation errors. This commit fixes that by specifying
the "cuda" in the list of dependencies.


gzagaris:
NOTE: Linking CUDA here was required, b/c the `camp` guards are of the form `#ifdef __CUDA_ARCH__` . Linking `cuda` ensures that the code is fed into the NVCC compiler, which handles the `#ifdef` guards accordingly.

gzagaris:
Hi @davidbeckingsale, @mcfadden8   -- 

I think I would still need this change. I am not sure if there is a more appropriate fix. At your convenience, please, take a look. I can elaborate more if my earlier explanation of why I needed this was not clear.

src/umpire/op/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ if (ENABLE_CUDA)
 
   set (umpire_op_depends
     ${umpire_op_depends}
-    cuda_runtime)
+    cuda)
```
davidbeckingsale, line 70: I'm not sure this is required? The cuda dep (required by the DeviceAllocator) should be handled by the code added in the src/umpire/CMakeLists.txt

gzagaris: Hi @davidbeckingsale, I just commented this out and gave it a try. I am still getting errors like this:

```
Umpire/src/umpire/op/CudaCopyFromOperation.cpp:46:42: error: ‘Cuda’ is not a member of ‘camp::resources’
   auto device = ctx.get<camp::resources::Cuda>();
```

Again,  I think this is more of an issue with the `#ifdef __CUDA_ARCH__` guards in `camp`, which would only work if the code is fed into the NVCC compiler.

Also, note that the code in `src/umpire/CMakeLists.txt` adds the `cuda` dep to `umpire_depends`, which, is not used in this file. This file uses the `umpire_op_depends`.  

Instead of adding the `cuda` dep here, would you rather add `umpire_depends` to it? (I think that would work as well.)

Thanks again for taking a look. Let me know if you'd like me to make any changes or try anything else.

davidbeckingsale: I took bit more of a look into this, and we are handling the camp/cuda stuff here: https://github.com/LLNL/Umpire/blob/develop/src/umpire/tpl/CMakeLists.txt#L89-L92 Would a similar solution work for your project?

I agree we should have the cuda dep for the main library.

gzagaris: Thanks for looking into this. Yes, adding the definition `CAMP_HAVE_CUDA` to `camp` when building with CUDA works for me. I'll handle that on my end.

davidbeckingsale: Great, I'll remove this change from #581 and then get this merged.

gzagaris: Fantastic -- Thanks!

davidbeckingsale: Thank you for finding and fixing this!

PR #546, opened 2021-03-09
Task/um 844 fixed pool stress test

kab163:
**Output as of 3/22:**
  SAME_ORDER (FixedPool - 137438953472):
    alloc: 0.066874646(us)
    dealloc: 0.020496047(us)
    lifetime: 0.087370693(us)

  REVERSE_ORDER (FixedPool - 137438953472):
    alloc: 0.057279514(us)
    dealloc: 0.017566043(us)
    lifetime: 0.074845557(us)

  SHUFFLE_ORDER (FixedPool - 137438953472):
    alloc: 0.057184918(us)
    dealloc: 0.017525660(us)
    lifetime: 0.074710578(us)

  SAME_ORDER (FixedPool - 2147483648):
    alloc: 0.057128977(us)
    dealloc: 0.017348906(us)
    lifetime: 0.074477883(us)

  REVERSE_ORDER (FixedPool - 2147483648):
    alloc: 0.057150566(us)
    dealloc: 0.017399395(us)
    lifetime: 0.074549961(us)

  SHUFFLE_ORDER (FixedPool - 2147483648):
    alloc: 0.057364674(us)
    dealloc: 0.017485096(us)
    lifetime: 0.074849770(us)

  SAME_ORDER (FixedPool - 4194304):
    alloc: 0.057286170(us)
    dealloc: 0.017491672(us)
    lifetime: 0.074777842(us)

  REVERSE_ORDER (FixedPool - 4194304):
    alloc: 0.057318111(us)
    dealloc: 0.017618516(us)
    lifetime: 0.074936627(us)

  SHUFFLE_ORDER (FixedPool - 4194304):
    alloc: 0.057227529(us)
    dealloc: 0.017569389(us)
    lifetime: 0.074796918(us)


kab163:
LGTM

kab163:
LGTM

benchmarks/CMakeLists.txt
```diff
@@ -28,6 +28,11 @@ blt_add_executable(
   SOURCES copy_stress_test.cpp
   DEPENDS_ON ${stress_test_depends})
 
+blt_add_executable(
+  NAME fixed_pool_stress_test
+  SOURCES fixed_pool_stress_test.cpp
+  DEPENDS_ON ${stress_test_depends})
+
```
mcfadden8, line 35: I realize that this is simply using the current pattern "_stress_test" for the name, but I wonder whether the name would more accurately be called something like "benchmark_fixed_pool"?  I am thinking about what people might look for when the do an `ls ./bin` and see all of the executables there.  It might be nice to see all of the benchmarks, versus unit tests, versus stress tests, etc..

Thoughts?

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+#include "umpire/config.hpp"
+
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Allocator.hpp"
+#include "umpire/strategy/FixedPool.hpp"
+
+#define CONVERT 1000000 //convert sec (s) to microsec (us)
+
+const int NUM_RND = 1000; //number of rounds (used to average timing)
```
mcfadden8, line 14: I notice we the use of both `=` and `{}` throughout the code.  Could we perhaps only use `=` where it is required and otherwise always use `{}`?

```suggestion
const int NUM_RND{ 1000 }; //number of rounds (used to average timing)
```

benchmarks/CMakeLists.txt
kab163: I think we should definitely discuss a naming convention here because I realized that, after not looking at this since like last wednesday, I at first thought this was a "fixed" version of the pool_stress_test.cpp. So I do think it could be called something better that would give a bit more clarity as to what it is/measures/etc. Perhaps this is something to discuss for tomorrow!

benchmarks/fixed_pool_stress_test.cpp
kab163: Good idea

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+const int NUM_ALLOC = 512; //number of allocations used for testing
+const int FACTOR = 2048; //Default size used when creating FixedPool (64*sizeof(int)*8)
```
mcfadden8, line 16: What is `FACTOR` being used for in this program?  The only time I see it being used is for `cout`.

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+  //Set up formatting for output
+  std::cout << std::fixed << std::setprecision(9); 
+
+  auto& rm = umpire::ResourceManager::getInstance();
+  umpire::Allocator alloc = rm.getAllocator("HOST");
+  
+  //NUM_SIZES: Using two different sizes to show average times don't vary much.
+  //SIZE: Array of sizes used: 2^26, 2^11 (large vs. small)
+  const int NUM_SIZES = 2; 
+  static int SIZE[NUM_SIZES] = {67108864, 2048};
```
mcfadden8, line 136:
```suggestion
  static int SIZE[NUM_SIZES] = {2<<26, 2<<11};
```

kab163: Good catch lol

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+    time[0] += std::chrono::duration<double>(end_alloc - begin_alloc).count()/NUM_ALLOC;
+
+    auto begin_dealloc = std::chrono::system_clock::now();
+    for (int h = 0; h < NUM_ALLOC; h++)
+      alloc.deallocate(allocations[h]);
+    auto end_dealloc = std::chrono::system_clock::now();
+    time[1] += std::chrono::duration<double>(end_dealloc - begin_dealloc).count()/NUM_ALLOC;
+  }
+
+  alloc.release();
```
mcfadden8, line 46: Okay, I promise to pay (better) attention this time.  Why is this call needed again?  You've appear to have done a matching allocate/deallocate pair above.

mcfadden8:
This is looking good.  I noticed that 99% of the code is same between the order, reverse, and random functions.  It would be easier to review and debug if the code could be refactored just a little to where you could have one implementation function.  The deallocate loop could look something like:
```
for (int i = 0; i < allocations.size(); i++ )
{
   allocator.deallocate(allocation_index[i]);
}
```

The `allocation_index` could be vector of indexes to use that could get passed in a parameter to the function and would be one of (`order_allocation_index`, `reverse_allocation_index`, or `random_allocation_index`).  These allocation indexes could be calculated by the calling function.

benchmarks/fixed_pool_stress_test.cpp
kab163: When using the Google Benchmark stuff with FixedPool, we came across the problem where we would "deallocate" memory from a pool, but it was still marked as being 'used'. So even with deallocating, it wasn't "releasing" that memory to be used again by the next benchmark iterations. So the memory piled up and we kept running out of memory before the benchmark could finish even though the system memory info on Lassen says it should work fine. I just kind of rolled that logic over into the stress tests too (I don't think I've run the test without the release to see if it would be ok, I just assumed it was good to use for the stress tests too).

benchmarks/fixed_pool_stress_test.cpp
kab163: This could probably be done better, but it's my way of showing what the total amount of memory allocated will be. My largest size is 2^26, but it wouldn't be accurate to say 2^26 bytes is being allocated in total - it's really 2^26 * objects_per_pool (which in this case is 2048 or (64 * sizeof(int) * 8)). Factor just stores that default value for me. I've been debating modifying how I do that, but I haven't come up with anything I like better yet lol

kab163: And please correct me if I'm wrong on that... :o

benchmarks/fixed_pool_stress_test.cpp
mcfadden8: @davidbeckingsale, please feel free to correct me if I'm wrong, but I think this should work without having to call `release()`.

In general both our DynamicPools and FixedPool implementations work by performing large allocations, called `blocks`, from the resource that they are a pool for (DEVICE, HOST, etc.).  The Pool will then perform a 'sub-allocation' from these `blocks`.  When the pool cannot find enough available memory in its set of `blocks`, it will perform another large allocation from the resource and create a new block and then return a sub-allocation from it to the caller.

When the application gives memory back to the pool by calling deallocate, the pool returns that memory to a portion of its associated `block`.  Once a `block` gets all of its memory returned, it "may" be coallesced into a completely free (available) block.  Until this time, the `block` will only be partially available to the application.

There are two `block` administrative things that may be done by the application:

- `release()` - This will return all of the completely available blocks to the resource.
- `coalesce()` - This will release all available `blocks` and then allocate a single larger block that is equivalent in size 
to the sum of all of those released `block` sizes.

It would be interesting to me to see the reproducer for the problem @kab163 is describing.

benchmarks/fixed_pool_stress_test.cpp
mcfadden8: If I understand correctly, this is stating what the `objects_per_pool` parameter for the `FixedPool` constructor is looking for:

```
  /*!
   * \brief Constructs a FixedPool.
   *
   * \param name The allocator name for reference later in ResourceManager
   * \param id The allocator id for reference later in ResourceManager
   * \param allocator Used for data allocation. It uses std::malloc
   * for internal tracking of these allocations.
   * \param object_bytes The fixed size (in bytes) for each allocation
   * \param objects_per_pool Number of objects in each sub-pool
   * internally. Performance likely improves if this is large, at
   * the cost of memory usage. This does not have to be a multiple
   * of sizeof(int)*8, but it will also likely improve performance
   * if so.
   */
  FixedPool(const std::string& name, int id, Allocator allocator,
            const std::size_t object_bytes,
            const std::size_t objects_per_pool = 64 * sizeof(int) * 8) noexcept;
```

Perhaps it would be better to specify the `objects_per_pool` directly than using the default one.

```
const std::size_t OBJECTS_PER_BLOCK{ 1 << 11 };    // 2^11 = 2048 blocks of object_bytes size
```

kab163: Done, thanks :)

benchmarks/fixed_pool_stress_test.cpp
mcfadden8: Since this benchmark creates multiple pool allocators, it must `release` all of the memory resources used by the allocator when it is finished testing it before creating the next allocator to be tested.

benchmarks/CMakeLists.txt
mcfadden8: As discussed, we can probably do this at a later time.  I suggest that a separate Jira ticket be written to describe the naming that should be used for all of the executables placed in the ./bin directory.

kab163: Sounds good!

kab163:
LGTM

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -1,122 +1,42 @@
...
-  double dealloc_t{(time[1]/double(NUM_RND)*CONVERT)};
-
-  std::cout << "  REVERSE_ORDER (FixedPool - " << (long long int)SIZE*FACTOR << "):" << std::endl; 
-  std::cout << "    alloc: " << alloc_t << "(us)" << std::endl;
-  std::cout << "    dealloc: " << dealloc_t << "(us)" << std::endl;
-  std::cout << "    lifetime: " << alloc_t+dealloc_t << "(us)" << std::endl << std::endl;
-}
+const int NUM_RND {1000}; //number of rounds (used to average timing)
+const int NUM_ALLOC {512}; //number of allocations used for testing
+const int OBJECTS_PER_BLOCK {1<<11}; //2^11 == 2048 blocks of object_bytes size
```
mcfadden8, line 14: Lol - Thinking about comments from @davidbeckingsale yesterday, maybe the world isn't completely ready for bit shifting :-).  Either way, I think there is a typo and suggest that this might be fixed in one of two ways:
```suggestion
const int OBJECTS_PER_BLOCK {1<<11}; //1^11 == 2048 blocks of object_bytes size
```
or
```suggestion
const int OBJECTS_PER_BLOCK {2 * 1024}; // # of blocks of object_bytes size
```

benchmarks/CMakeLists.txt
kab163: See UM-849

benchmarks/fixed_pool_stress_test.cpp
kab163: 1^11 != 2048 though lol
Maybe I'll comment "1<<11 == 2^11 == 2048 ...."  haha

kab163:
LGTM

mcfadden8 (commented):
Looks great.  Thanks!

mcfadden8 (approved):

benchmarks/fixed_pool_stress_test.cpp
mcfadden8: lol - you're right.  But I don't know that 1<<11 is the same as 2 exclusive or'd with 11.

kab163:
LGTM

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+  std::vector<int> reverse_order_index(same_order_index);
+  std::reverse(reverse_order_index.begin(), reverse_order_index.end());
+
+  //create vector of indices for "shuffle_order" tests
+  std::vector<int> shuffle_order_index(same_order_index);
+  std::mt19937 gen(NUM_ALLOC);
+  std::shuffle(&shuffle_order_index[0], &shuffle_order_index[NUM_ALLOC], gen);
+ 
+  //create the FixedPool allocator and run stress tests for all sizes
+  for(int i = 0; i < NUM_SIZES; i++)
```
davidbeckingsale, line 80: Put the sizes in a vector and do `for (auto size : sizes)`

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+  {
+    umpire::Allocator pool_alloc = rm.makeAllocator<umpire::strategy::FixedPool, false>
+                                 ("fixed_pool" + std::to_string(i), alloc, SIZE[i], OBJECTS_PER_BLOCK);
+    run_test(pool_alloc, SIZE[i], same_order_index, "SAME_ORDER");
```
davidbeckingsale, line 84: `SIZE[i]` will go away with comment from above.

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <chrono>
+#include <string>
+#include <random>
+#include <numeric>
+
+#include "umpire/config.hpp"
+#include "umpire/ResourceManager.hpp"
+#include "umpire/Allocator.hpp"
+
+#define CONVERT 1000000 //convert sec (s) to microsec (us)
```
davidbeckingsale, line 16: This can be constexpr.

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+const int NUM_RND {1000}; //number of rounds (used to average timing)
```
davidbeckingsale, line 18: So can these...

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+const int NUM_ALLOC {512}; //number of allocations used for testing
+const int OBJECTS_PER_BLOCK {1<<11}; //number of blocks of object_bytes size (2048)
+
+void run_test(umpire::Allocator alloc, int SIZE, std::vector<int> &indices, std::string test_name)
```
davidbeckingsale, line 22: Give the method a more descriptive name, like `test_deallocation_performance`.

benchmarks/fixed_pool_stress_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+  
+  //Using a few different sizes to show average times don't vary much.
+  const int NUM_SIZES {3}; 
+
+  //Array of sizes used: 67108864, 1048576, 2048 (large vs. medium vs. small)
+  static int SIZE[NUM_SIZES] {1<<26, 1<<20, 1<<11};
+
+  //create vector of indices for "same_order" tests
+  std::vector<int> same_order_index(NUM_ALLOC);
+  for(int i = 0; i < NUM_ALLOC; i++)
```
davidbeckingsale, line 67: Add {} for the loop!

davidbeckingsale (changes requested):

benchmarks/fixed_pool_stress_test.cpp
kab163: Done

benchmarks/fixed_pool_stress_test.cpp
kab163: Done

benchmarks/fixed_pool_stress_test.cpp
kab163: Done

benchmarks/fixed_pool_stress_test.cpp
kab163: Done

benchmarks/fixed_pool_stress_test.cpp
kab163: :O gasp! Done lol

benchmarks/fixed_pool_stress_test.cpp
kab163: Done

davidbeckingsale (approved):

PR #549, opened 2021-03-30
Bugfix/um 851 zero byte allocations are sometimes not found

mcfadden8:
This turned out to be a difference between how the P9 architecture seems to be handling left-shift operations as compared to how x86 architectures do.  Before this fix, the code depended upon a shift operation wrapping back around if the number of bits were greated than the size of the word.  In this case, a << of 64 on an unsigned long long would effectively be the same thing as a << 0 on x86 architectures.  But on P9, the << operator would effectively shift the bits completely out of the word.


mcfadden8:
Closing this PR so we may do patch release first.

PR #553, opened 2021-04-07
Feature: POSIX Host Shared Memory Resource Added

mcfadden8:


examples/cookbook/recipe_shared_memory.cpp
```diff
@@ -0,0 +1,120 @@
...
+// in the debugger session windows, no MPI will be used and the debugger must
+// be used for synchronization (by setting breakpoints).
+//
+int main(
+  int
+  ac
+  ,
+  char**
+  av
+)
```
kab163, line 42: This hurts my brain!

mcfadden8: oops.  that was left over from when I had #ifdefs between those items (which I wasn't compiling with MPI)..  I fix...

examples/cookbook/recipe_shared_memory.cpp
```diff
@@ -0,0 +1,120 @@
...
+    MPI_Init(&ac, &av);
+  }
+
+  auto& rm = umpire::ResourceManager::getInstance();
+
+  //
+  // Set up the traits for the allocator
+  //
+  auto traits{umpire::get_default_resource_traits("SHARED")};
+  traits.size = 1*1024*1024;  // Maximum size of this Allocator
```
kab163, line 57: For this whole example, it seems like it goes back and forth between initializing like "int var = 0;" and "int var {0};" - should we keep this more consistent?

mcfadden8: Not sure I follow.  Do you have a suggestion for this line?

kab163: Sorry, Marty, I misread this - it's fine! Whoopsies.

src/umpire/ResourceManager.hpp
```diff
@@ -317,6 +317,7 @@ class ResourceManager {
   int getNextId() noexcept;
 
   std::string getAllocatorInformation() const noexcept;
+  std::string getResourceInformation() const noexcept;
```
kab163, line 320: Where is this function implemented?

mcfadden8: good question!  I have no idea.  I must have inadvertantly picked it up during a merge from somewhere?  I honestly don't remember adding it, but `git blame` doesn't lie.  I'll remove.

mcfadden8: Fixed.

cmake/SetupCompilerFlags.cmake
```diff
@@ -57,7 +57,7 @@ if (ENABLE_PEDANTIC_WARNINGS)
     DEFAULT  ""
     GNU "-Wpedantic"
     CLANG "-Wpedantic"
-    INTEL "-Wall -Wcheck -wd2259"
+    INTEL "-Wall -Wcheck -wd2259 -wd654"
```
davidbeckingsale, line 60: I think it might be worth adding a comment explaining what warnings we are disabling.

examples/cookbook/recipe_shared_memory.cpp
davidbeckingsale: This isn't fixed.

scripts/spack_packages/umpire/package.py
```diff
@@ -112,6 +113,8 @@ class Umpire(CMakePackage, CudaPackage):
     conflicts('+openmp', when='+hip')
     conflicts('+openmp_target', when='+hip')
     conflicts('+deviceconst', when='~hip~cuda')
+    conflicts('~mpi', when='+posix_shmem', msg='Shared Memory Allocator requires MPI')
```
davidbeckingsale, line 116: This is saying posix_shmem requires MPI, but this is not enfornced at CMake time.

src/umpire/Umpire.cpp
```diff
@@ -168,4 +170,56 @@ std::vector<util::AllocationRecord> get_leaked_allocations(Allocator allocator)
...
 
+umpire::MemoryResourceTraits get_default_resource_traits(const std::string name)
+{
+  umpire::resource::MemoryResourceRegistry&
+    registry{ umpire::resource::MemoryResourceRegistry::getInstance() };
+  umpire::MemoryResourceTraits traits{ registry.getDefaultTraitsForResource(name) };
+  return traits;
+}
+
+void* find_pointer_from_name(Allocator allocator, const std::string name)
```
davidbeckingsale, line 181: Do you think we need to make this work for non-shared Allocators since they support named allocation

src/umpire/strategy/AllocationStrategy.hpp
```diff
@@ -44,8 +44,8 @@ class AllocationStrategy
 
   virtual ~AllocationStrategy() = default;
 
-
   void* allocate_internal(std::size_t bytes);
+  void* allocate_internal(const std::string& name, std::size_t bytes);
```
davidbeckingsale, line 48: This is unused.

davidbeckingsale (changes requested):

cmake/SetupCompilerFlags.cmake
mcfadden8: Me too!  Especially since I cannot remember now :-)

src/umpire/strategy/AllocationStrategy.hpp
mcfadden8: Agreed.

cmake/SetupCompilerFlags.cmake
mcfadden8: Turns out, this extra option was not needed so I took it back out.  Will make sure and add a comment if/when it comes back.

src/umpire/Umpire.cpp
mcfadden8: I'm not sure I follow?

scripts/spack_packages/umpire/package.py
mcfadden8: @davidbeckingsale, I have attempted to addres this with https://github.com/LLNL/Umpire/pull/553/commits/650df6707640250f5ee9dce873dbf3d650194a5b

Please let me know if you think that is sufficient.  It seems to work.  The only thing I don't like is that it does not really offer any feedback to the user other than "Shared Memory is Disabled" if the user does not also specify ENABLE_MPI.

examples/cookbook/recipe_shared_memory.cpp
mcfadden8: Fixed.

src/umpire/Umpire.cpp
davidbeckingsale: This throws an error, but when calling `allocate(name, size)` on a non-shared allocator, the name is just ignored. I think it's a strange mismatch.

mcfadden8: But it's not.  Allocate(name,size) throws an error too: https://github.com/LLNL/Umpire/blob/cc10fea3cd7a9cc37ded9f58d8fb57aff2edaa57/src/umpire/strategy/AllocationStrategy.cpp#L31

davidbeckingsale: Ah okay. I don't think it used to, did it? We can resolve this then.

mcfadden8: Oh.  oops.  I intended for it to throw an error.

src/umpire/strategy/AllocationStrategy.cpp
```diff
@@ -28,6 +28,13 @@ void* AllocationStrategy::allocate_internal(std::size_t bytes)
   return allocate(bytes);
 }
 
+void* AllocationStrategy::allocate_named(const std::string& UMPIRE_UNUSED_ARG(name), std::size_t UMPIRE_UNUSED_ARG(bytes))
+{
+  UMPIRE_LOG(Error, "AllocationStrategy::allocate_named() not implemented");
```
davidbeckingsale, line 33: I think this error message should hint at the fact allocate_named() is only supported for shared resources.

mcfadden8: Okay.  I think I have fixed this now.  If you agree, let's mark this as resolved.

davidbeckingsale: Yes, but the message just tells me that `allocate_named` is not implemented. As a user, it would be more helpful if it explained that allocate_named only exists for shared resources.

If you don't think that's appropriate, we need to catch this incorrect usage higher up the chain (i.e. in the Allocator)

davidbeckingsale (approved):

tests/debug/allocate_deallocate.cpp
```diff
@@ -10,9 +10,9 @@
 
 int main() {
   auto &rm = umpire::ResourceManager::getInstance();
-  auto alloc = rm.getAllocator("NO_OP");
+  auto alloc = rm.getAllocator("HOST");
```
kab163, line 13: Why are we changing this? I thought No_Op was needed here?

src/umpire/resource/HostSharedMemoryResourceFactory.cpp
```diff
@@ -0,0 +1,60 @@
...
+  return create(name, id, getDefaultTraits());
+}
+
+std::unique_ptr<resource::MemoryResource> HostSharedMemoryResourceFactory::create(
+    const std::string& name, int id, MemoryResourceTraits traits)
+{
+  if (traits.scope != MemoryResourceTraits::shared_scope::node) {
+    UMPIRE_ERROR("HostSharedMemoryResource only supports shared_scope::node");
+  }
+  return util::make_unique<HostSharedMemoryResource>(Platform::undefined, name, id,
```
kab163, line 38: In this instance, and in general, I am confused by the use of undefined as a platform type (like in the traits file too). What is happening with this?

tests/debug/allocate_deallocate.cpp
mcfadden8: This was accidental.  However, I thought that the stuff under "debug" was changeable?  I didn't think they were actually tests that ran.

src/umpire/resource/HostSharedMemoryResourceFactory.cpp
mcfadden8: What should the platform be?

mcfadden8: It probably should be Platform::host.

davidbeckingsale: Host.

tests/debug/allocate_deallocate.cpp
davidbeckingsale: This can be anything. Like Marty said, it's not tests that run, it's just little code snippets useful for debugging.

kab163: Ah, ok. It's been so long since this was created that I forgot! Ha :p

src/umpire/strategy/AllocationStrategy.cpp
mcfadden8: Done

src/umpire/resource/HostSharedMemoryResourceFactory.cpp
mcfadden8: Done!  Nice catch!

kab163: But! ... It's not done lol what about the isAccessibleFrom function? Hold on, I'll link to it...

src/umpire/resource/HostSharedMemoryResource.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void HostSharedMemoryResource::deallocate(void* ptr, std::size_t)
+{
+  return pimpl->deallocate(ptr);
+}
+
+bool HostSharedMemoryResource::isAccessibleFrom(Platform p) noexcept
+{
+  // Todo: Need to determine how to update tests which require
```
kab163, line 50: Here!

kab163:
```suggestion
   if(p == Platform::host)
    return true;
  else
    return false;
```

kab163: oh I messed that up...

src/umpire/resource/HostSharedMemoryResource.cpp
```diff
@@ -0,0 +1,72 @@
...
+  // unnamed alloc to test.
+  //
+  UMPIRE_USE_VAR(p);
+  return false;
```
kab163, line 54:
```suggestion
  if(p == Platform::host)
    return true;
  else
    return false;
```

kab163: Sorry for the barrage of email notifications....

src/umpire/resource/HostSharedMemoryResource.cpp
mcfadden8: ??

kab163 (approved):
(Note: Accessibility ticket! Hurray!)

PR #555, opened 2021-04-13
Bugfix/um 856 hip rangelo

kab163:


benchmarks/vendor_allocator_benchmarks.cpp
```diff
@@ -21,16 +21,26 @@
...
 
-static const int RangeLow{1<<10}; //1kB
+/*
+ * Note: HIP runs need their own RangeLow since hipMalloc 
+ * makes allocations aligned along 4k pages. HIP UM and 
+ * PINNED memory currently have a bug which may require
+ * further modification of the ranges used.
+ */
+#if defined(UMPIRE_ENABLE_HIP)
+  static const int RangeLow{1<<12}; //1kB
```
davidbeckingsale, line 31: Comment `//1kB` is wrong here.

kab163: ah good catch

mcfadden8 (commented):
I wonder if we should simply change the rangelo to 4k for all of the device types in order to keep the ranges used the same when showing comparisons?

kab163:
> I wonder if we should simply change the rangelo to 4k for all of the device types in order to keep the ranges used the same when showing comparisons?

Yea good point. I think the key question is whether or not any valuable info is gained by doing an additional range value for CUDA. The cudaMalloc has the ability to do 1024 byte allocations until exhausting max memory, while hipMalloc can't - but is that necessarily important to know? (Keeping it all the same rangelo will definitely make it look nicer hehe)

davidbeckingsale (approved):

mcfadden8 (approved):

PR #560, opened 2021-04-20
SYCL-2020 requires using -std=c++17

victor-anisimov:
Need to switch from -std=c++11 to -std=c++17 to adopt SYCL-2020 standard when SYCL is enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,11 @@ if(WIN32 OR APPLE)
 endif()
 option(ENABLE_FILE_RESOURCE "Enable File Resource" On)
 
-set(BLT_CXX_STD "c++11" CACHE STRING "Version of C++ standard")
+if (ENABLE_SYCL)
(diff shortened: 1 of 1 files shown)


davidbeckingsale:
Thanks @victor-anisimov - do you have a working build like this? I noticed that the json library we use is causing problems in our CI.

victor-anisimov:
@davidbeckingsale I do everything manually. I work on SYCL / Latest oneAPI SDK / Intel GPU platform. After updating CMakeLists.txt in the way I suggested, `cmake -H. -Bbuild -DENABLE_SYCL=On -DUMPIRE_ENABLE_FILESYSTEM=OFF -DENABLE_TESTS=On -DENABLE_EXAMPLES=On -DCMAKE_INSTALL_PREFIX=$HOME/Umpire/install -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=dpcpp .` All tests `make test` pass except test 21, which fails presumably due to the USM pointer being created and deleted in different contexts. I'm working on it, but that is a separate issue.

PR #561, opened 2021-04-21
Add registerAllocation to C & Fortran interfaces

davidbeckingsale:


tests/integration/interface/allocator_c_tests.cpp
```diff
@@ -498,3 +498,21 @@ TEST(Allocators, GetInvalidId)
...
+
+  double d = 1.0;
+
+  ASSERT_NO_THROW(
+    umpire_resourcemanager_register_allocation(&rm, &d, sizeof(double), allocator)
+  );
+
+  ASSERT_NO_THROW(
+    umpire_resourcemanager_deregister_allocation(&rm, &d)
+  );
```
kab163, line 517: My only question is whether or not we want to test this more than "can it register something" and "can it deregister something" - like, I think the c++ test also asks "given that it's registered something, can it do a simple query on it"... you know what i mean? should that kind of thing be added here? https://github.com/LLNL/Umpire/blob/a11edc3d90321990a3175a42b9932d73834f96b0/tests/integration/allocator_integration_tests.cpp#L185

mcfadden8 (approved):

tests/integration/interface/allocator_c_tests.cpp
davidbeckingsale: Fixed in latest commit.

PR #562, opened 2021-04-21
Task/um 864 hpctoolkit docs

kab163:
Updated:
https://umpire.readthedocs.io/en/task-um-864-hpctoolkit-docs/developer/hpctoolkit.html


docs/sphinx/developer/hpctoolkit.rst
```diff
@@ -0,0 +1,192 @@
+.. _hpctoolkit:
+
+==========
+HPCToolKit
+==========
+
+This page will describes the process and series of steps to analyze Umpire specific 
+applications with HPCToolKit.
```
davidbeckingsale, line 8: Change this to link to HPCToolkit page.

docs/sphinx/developer/hpctoolkit.rst
```diff
@@ -0,0 +1,192 @@
...
+.. note::
+   Depending on what's available on your local machine, you may have to download or update Java
+   in order to run hpcviewer. There are instructions `here <http://hpctoolkit.org/download/hpcviewer/>`_
+   for hpcviewer. You can get Java 8 from `here <https://www.oracle.com/java/technologies/javase/javase-jdk8-downloads.html>`_.
+
+
+Running with Hatchet
+^^^^^^^^^^^^^^^^^^^^
+
+`Hatchet <https://hatchet.readthedocs.io/en/latest/index.html>`_ is a tool that can better analyze performance metrics given from a variety of tools,
```
davidbeckingsale, line 101: I think it would be better to link to the hatchet github page, rather than straight to the docs.

docs/sphinx/developer/hpctoolkit.rst
```diff
@@ -0,0 +1,192 @@
...
+* Grab the example program's database from a different branch I want to compare against
+* Create a graphframe for each database
+* Create a filtered graphframe for each that focuses on the specific part of the program I want to measure against
+* Compare the inclusive time for each filtered graphframe (or whatever metric I want to analyze)
+* If the metric (e.g., inclusive time) of the new branch's filtered graphframe is more than ``threshold`` more than that of develop's, then fail the test!
+
+This is just one example of how analyzing results from HPCToolKit and Hatchet can be used to improve performance of Umpire.
+There are many other ways to do this - once we have more interesting use cases, this page will be updated!
+
+.. note::
```
davidbeckingsale, line 189: This should be removed.

davidbeckingsale (changes requested):

docs/sphinx/developer/hpctoolkit.rst
kab163: done

docs/sphinx/developer/hpctoolkit.rst
kab163: Done

docs/sphinx/developer/hpctoolkit.rst
kab163: Done

mcfadden8 (approved):
This looks great @kab163, thanks for the write-up.  I also like the referenced Jira ticket. Although, that will only be visible to people with permission to get to the link which I think should be okay since it is HPC-specific.

docs/sphinx/developer/hpctoolkit.rst
davidbeckingsale: I meant this whole note should be removed. I don't think linking to JIRA from readthedocs is necessary or appropriate

davidbeckingsale (changes requested):

davidbeckingsale (approved):

kab163:
> This looks great @kab163, thanks for the write-up. I also like the referenced Jira ticket. Although, that will only be visible to people with permission to get to the link which I think should be okay since it is HPC-specific.

Yea I like the link too, but I think it would be better placed elsewhere. Like the ReadTheDocs probably isn't the best spot for it. But I'll keep note of that ticket # elsewhere, so we can always go back and reference it!

PR #565, opened 2021-04-23
UM-866 Added guard for ASAN

mcfadden8:


mcfadden8:
This is ready for review.  I've confirmed that the sanitizer tests are still being succefully compiled and run: https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/225642#L4705

adayton1 (approved):
LGTM

cmake/SetupCompilerFlags.cmake
```diff
@@ -32,21 +32,6 @@ else ()
...
-  "#include <sanitizer/asan_interface.h>
-
-  int main(int, char**)
-  {
-    return 0;
-  }"
-  UMPIRE_HAS_ASAN)
-
-if (UMPIRE_HAS_ASAN)
-  message(STATUS "Umpire may be built with ASAN support")
```
kab163, line 45: Are you sure we want to get rid of this? It seems like it could be good to have the option for automatic detection.. what if we modified to so that it will turn `UMPIRE_HAS_ASAN` to ON unless `ENABLE_ASAN` has been manually turned off?

.gitlab/lassen-jobs.yml
```diff
@@ -105,7 +105,7 @@ clang_9_0_0_libcpp (build and test on lassen):
 
 clang_9_0_0_memleak (build and test on lassen):
   variables:
-    SPEC: "+sanitizer_tests %clang@9.0.0 cxxflags=-fsanitize=address"
+    SPEC: "+asan +sanitizer_tests %clang@9.0.0 cxxflags=-fsanitize=address"
```
kab163, line 108: Does the datarace test also need this? (https://github.com/LLNL/Umpire/blob/7e6acc84abad6041070554c9279643a0c20db42f/.gitlab/lassen-jobs.yml#L114)

mcfadden8: I don't think so.  This is address sanitizer (ASAN) specific (`-fsanitize=address`)

kab163: Mkay, just checking.

cmake/SetupCompilerFlags.cmake
mcfadden8: It's a good question. My inclination is to remove it as it is not reliable.  We have found (too many) systems where the header file exists, but the runtime support is not there.  Or the runtime advertises support, but the header file is missing.

.gitlab/lassen-jobs.yml
kab163: Oh and I just looked and the datarace test runs in the CI (it fails where we already know it will), so that's a good sign lol

cmake/SetupCompilerFlags.cmake
kab163: Oh I see. Well that makes sense. So I guess this means that ASAN support will be up to the user to figure out - which should be fine, right?

docs/sphinx/advanced_configuration.rst
```diff
@@ -91,3 +92,7 @@ These arguments are explained in more detail below:
 * ``ENABLE_PERFORMANCE_TESTS``
   Build and run performance tests
 
+* ``ENABLE_ASAN``
+  This option enables address sanitization checks within Umpire by compilers
+  that support options like -fsanitize=address
```
kab163, line 97: I'm being picky, I know, but I get OCD sometimes :)
```suggestion
  that support options like ``-fsanitize=address``
```

kab163 (approved):
Looks good! Thanks!

davidbeckingsale (approved):

PR #566, opened 2021-04-30
Um 869 add accessibility check 2 host shared memory resource

mcfadden8:


davidbeckingsale (approved):

src/umpire/resource/HostSharedMemoryResource.cpp
```diff
@@ -45,15 +49,35 @@ void HostSharedMemoryResource::deallocate(void* ptr, std::size_t)
...
+  cudaGetDevice(&cdev);
+
+  //Device supports coherently accessing pageable memory
+  //without calling cudaHostRegister on it
+  cudaDeviceGetAttribute(&pageableMem,
+     cudaDevAttrPageableMemoryAccess, cdev);
+  if(pageableMem)
+    return true;
+#endif
+  // Note: Regarding omp_target, we pick a default of false here
```
kab163, line 66: Is this comment located in the right spot? Like, if we're on cuda platform, would we actually get here if we were also dealing with omp_target?

kab163 (approved):

PR #569, opened 2021-05-05
Added block-based heuristic and mechanism to obtain pool block stats

mcfadden8:


src/umpire/strategy/DynamicPoolList.cpp
```diff
@@ -55,6 +55,16 @@ void DynamicPoolList::release()
   dpa.release();
 }
 
+std::size_t DynamicPoolList::numReleasableBlocks() const noexcept
```
davidbeckingsale, line 58: These methods should be prefixed with `get`.

src/umpire/strategy/DynamicPoolList.cpp
```diff
@@ -113,8 +123,16 @@ void DynamicPoolList::coalesce() noexcept
   dpa.coalesce();
 }
 
-DynamicPoolList::CoalesceHeuristic DynamicPoolList::percent_releasable(
-    int percentage)
+DynamicPoolList::CoalesceHeuristic DynamicPoolList::default_heuristic{percent_releasable(100)};
+
+DynamicPoolList::CoalesceHeuristic DynamicPoolList::releasable_blocks(std::size_t nblocks)
```
davidbeckingsale, line 128: I think this should be `blocks_releasable` or `num_blocks_releasable` to better match up with `percent_releasable`

src/umpire/strategy/QuickPool.hpp
```diff
@@ -35,6 +35,12 @@ class QuickPool : public AllocationStrategy, private mixins::AlignedAllocation {
   using CoalesceHeuristic = std::function<bool(const strategy::QuickPool&)>;
 
   static CoalesceHeuristic percent_releasable(int percentage);
+  static CoalesceHeuristic releasable_blocks(std::size_t nblocks);
+
+  static constexpr std::size_t default_first_block_size{512 * 1024 * 1024};
+  static constexpr std::size_t default_next_block_size{1 * 1024 * 1024};
+  static constexpr std::size_t default_alignment{16};
+  static CoalesceHeuristic default_heuristic; // percent_releasable(100)
```
davidbeckingsale, line 43: `const`?

davidbeckingsale (changes requested):

src/umpire/strategy/DynamicPoolList.cpp
mcfadden8: done.

src/umpire/strategy/DynamicPoolList.cpp
mcfadden8: agreed.  done

src/umpire/strategy/QuickPool.hpp
mcfadden8: done

davidbeckingsale (approved):

PR #573, opened 2021-05-06
Fix for how memory resource is set

kab163:
- Checked the other benchmarks for similar problems, but it seems to really only affect this benchmark
- Is it ok to leave the FixedPool benchmark alone? (Currenly uses HOST resource only)


benchmarks/pool_stress_test.cpp
```diff
@@ -72,7 +72,13 @@ template <class T>
 void do_test(std::string pool_name, std::map<const std::string, const std::vector<std::size_t>&> indexing_pairs)
 {
   auto& rm {umpire::ResourceManager::getInstance()};
+
+#if defined (UMPIRE_ENABLE_DEVICE)
```
davidbeckingsale, line 76: I think always using HOST would be fine here.

kab163: Oh wow, even easier haha - okay!

davidbeckingsale (approved):

PR #575, opened 2021-05-10
Create benchmark for DeviceAllocator

kab163:
Notes from the JIRA ticket:

- [ ] Currently allocating doubles - should we allocate floats instead?
- [ ] Is the way I set THREADS_PER_BLOCK, BLOCKS_PER_SM, and NUM_SM ok?
- [ ] Currently, I ask whether or not concurrent kernels is supported. If so, I use the multiplication of the 3 values mentioned above to set the total number of threads. If not, I do something else. Maybe I should just set the number of threads a single way, but leave a warning if concurrent kernels aren't supported (benchmark should still work either way).


kab163:
https://rzlc.llnl.gov/jira/browse/UM-831?focusedCommentId=378667&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-378667

kab163:
Benchmark results from RzAnsel:
Running on device: Tesla V100-SXM2-16GB
Number of threads: 2621440

Kernel: Warm-up
Total time: 263.648us
Time per allocation: 2.63648us
Retrieved value: 512

Kernel: First thread per block
Total time: 1327.2us
Time per allocation: 0.00518438us
Retrieved value: 1024

Kernel: Only thread idx 0
Total time: 240.768us
Time per allocation: 2.40768us
Retrieved value: 512

Kernel: Each thread
Total time: 431779us
Time per allocation: 0.00164711us
Retrieved value: 256

(Ran with gcc/8.3.1)
(See JIRA ticket for more results)

PR #583, opened 2021-05-19
Removed DynamicPool and DynamicPoolMap from replay_test

mcfadden8:


tests/integration/allocator_accessibility.cpp
```diff
@@ -116,7 +116,7 @@ struct allocate_and_use<omp_target_platform>
 
 #pragma omp target is_device_ptr(d_data) device(dev)
 #pragma omp teams distribute parallel for schedule(static, 1)
-    for (auto i = 0; i < size; ++i) {
```
mcfadden8, line 119: I fixed a build warning that I was getting here when OpenMP was enabled.

PR #587, opened 2021-05-26
Return MPI_COMM_NULL when scope is not node

davidbeckingsale:


mcfadden8 (commented):
Looks great.  Would you add a simple test for this to test:
1. Return of cached communicator
2. Return of null type for non-Node
3. Return of new communicator

mcfadden8 (approved):

PR #589, opened 2021-05-26
umpire::mark_event() implementation added

mcfadden8:


src/umpire/Umpire.cpp
```diff
@@ -143,6 +143,11 @@ std::size_t get_process_memory_usage()
 #endif
 }
 
+void mark_event(const std::string event)
```
davidbeckingsale, line 146:
```suggestion
void mark_event(const std::string& event)
```

src/umpire/Umpire.hpp
```diff
@@ -150,6 +150,11 @@ std::string get_backtrace(void* ptr);
  */
 std::size_t get_process_memory_usage();
 
+/*!
+ * \brief Mark an application-specific event string within Umpire life cycle.
+ */
+void mark_event(const std::string event);
```
davidbeckingsale, line 156:
```suggestion
void mark_event(const std::string& event);
```

mcfadden8: - [ ]

davidbeckingsale (approved):

PR #600, opened 2021-07-07
Switch uberenv to use CachedCMakePackage

davidbeckingsale:


.gitlab/lassen-jobs.yml
```diff
@@ -105,7 +105,7 @@ xl_16_1_1_7_gcc_8_3_1_cuda_11:
 
 clang_9_0_0_libcpp (build and test on lassen):
   variables:
-    SPEC: "%clang@9.0.0+libcpp"
+    SPEC: "%clang@9.0.0+libcpp cxxflags='-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0' cflags='-DGTEST_HAS_CXXABI_H_=0'"
```
adrienbernede, line 108: I guess you don't want those flags in the compiler definition.
How about leaving a comment to explain that?

scripts/gitlab/build_and_test.sh
```diff
@@ -65,18 +65,18 @@ if [[ -z ${hostconfig} ]]
 then
     # If no host config file was provided, we assume it was generated.
     # This means we are looking of a unique one in project dir.
-    hostconfigs=( $( ls "${project_dir}/"hc-*.cmake ) )
```
adrienbernede, line 68: I liked the prefix because it help making the filename unique. (the script accept only one result).
I agree that this should be safe though. There is no reason for umpire to have a `.cmake` file in the project root dir.

scripts/spack_packages/umpire/package.py
```diff
@@ -118,290 +83,149 @@ class Umpire(CMakePackage, CudaPackage):
...
-            cfg.write("#------------------{0}\n\n".format("-" * 60))
-
-            cfg.write(cmake_cache_option("ENABLE_HIP", True))
-
-#            -DHIP_ROOT_DIR=/opt/rocm-3.6.0/hip -DHIP_CLANG_PATH=/opt/rocm-3.6.0/llvm/bin
-
             hip_root = spec['hip'].prefix
             rocm_root = hip_root + "/.."
-            cfg.write(cmake_cache_entry("HIP_ROOT_DIR",
+            entries.append(cmake_cache_string("HIP_ROOT_DIR",
```
adrienbernede, line 173: Is there a `cmake_cache_entry` to treat the variable as a PATH? I won't change the behavior, I think, but... well it seems more appropriate.

scripts/spack_packages/umpire/package.py
```diff
@@ -118,290 +83,149 @@ class Umpire(CMakePackage, CudaPackage):
...
                                         hip_root))
-            cfg.write(cmake_cache_entry("HIP_CLANG_PATH",
+            entries.append(cmake_cache_string("HIP_CLANG_PATH",
```
adrienbernede, line 175: Is there a cmake_cache_entry to treat the variable as a PATH? I won't change the behavior, I think, but... well it seems more appropriate.

scripts/spack_packages/umpire/package.py
```diff
@@ -118,290 +83,149 @@ class Umpire(CMakePackage, CudaPackage):
...
-                                            description))
+                entries.append(cmake_cache_string("BLT_EXE_LINKER_FLAGS", flags, description))
 
 
-        gcc_toolchain_regex = re.compile(".*gcc-toolchain.*")
-        gcc_name_regex = re.compile(".*gcc-name.*")
-
-        using_toolchain = list(filter(gcc_toolchain_regex.match, spec.compiler_flags['cxxflags']))
-        using_gcc_name = list(filter(gcc_name_regex.match, spec.compiler_flags['cxxflags']))
         compilers_using_toolchain = ["pgi", "xl", "icpc"]
```
adrienbernede, line 143: I see icpc, so I wonder: pgc++ ? xlc++ ?

scripts/spack_packages/umpire/package.py
```diff
@@ -118,290 +83,149 @@ class Umpire(CMakePackage, CudaPackage):
...
+        spec = self.spec
 
-        cfg.write(cmake_cache_option("ENABLE_BENCHMARKS", 'tests=benchmarks' in spec))
-        cfg.write(cmake_cache_option("ENABLE_TESTS", not 'tests=none' in spec))
-        cfg.write(cmake_cache_option("ENABLE_TOOLS", '+tools' in spec))
-        cfg.write(cmake_cache_option("ENABLE_WARNINGS_AS_ERRORS", '+werror' in spec))
-        cfg.write(cmake_cache_option("ENABLE_ASAN", '+asan' in spec))
-        cfg.write(cmake_cache_option("ENABLE_SANITIZER_TESTS", '+sanitizer_tests' in spec))
+        entries = super(Umpire, self).initconfig_mpi_entries()
+        entries.append(cmake_cache_option("ENABLE_MPI", '+mpi' in spec))
```
adrienbernede, line 204: I’ll need a little more time for a serious review :)

adrienbernede (commented):
Looks nice and more concise. I would like to have a closer look tomorrow.

scripts/spack_packages/umpire/package.py
```diff
@@ -118,290 +83,149 @@ class Umpire(CMakePackage, CudaPackage):
...
                 gcc_prefix = join_path(gcc_bin, '..')
-                cfg.write(cmake_cache_entry("HIP_CLANG_FLAGS",
-                "--gcc-toolchain={0}".format(gcc_prefix))) 
-                cfg.write(cmake_cache_entry("CMAKE_EXE_LINKER_FLAGS",
-                hip_link_flags + " -Wl,-rpath {}/lib64".format(gcc_prefix)))
+                entries.append(cmake_cache_string("HIP_CLANG_FLAGS", "--gcc-toolchain={0}".format(gcc_prefix))) 
+                entries.append(cmake_cache_string("CMAKE_EXE_LINKER_FLAGS", hip_link_flags + " -Wl,-rpath {}/lib64".format(gcc_prefix)))
             else:
-                cfg.write(cmake_cache_entry("CMAKE_EXE_LINKER_FLAGS", hip_link_flags))
+                entries.append(cmake_cache_string("CMAKE_EXE_LINKER_FLAGS", hip_link_flags))
```
adrienbernede, line 188: So, if I understand correctly, HIP is be a compiler not recognized as such is Spack (yet). So when we ask for `umpire%gcc +hip` under the hood we do `umpire%hip +gcc`. Correct?

adrienbernede (approved):
I like you clean code ;)

scripts/spack_packages/umpire/package.py
davidbeckingsale: Yup, that's right. It's a bit like CUDA combo - and especially in Umpire since we mostly use the hip runtime library, we are still using a regular compiler to build most things.

scripts/spack_packages/umpire/package.py
davidbeckingsale: I think the prefixes will catch that when I try and do the matching.

PR #602, opened 2021-07-15
Refactored pool heuristic to return 0 or minimum pool size

mcfadden8:


src/umpire/strategy/DynamicPoolMap.cpp
```diff
@@ -253,25 +255,26 @@ void DynamicPoolMap::coalesce()
...
+    std::size_t size_post{getActualSize()};
 
     // If this removed anything from the map, re-allocate a single large chunk
     // and insert to free map
-    if (released_bytes > 0) {
-      UMPIRE_LOG(Debug, "coalescing " << released_bytes << " bytes.");
-      Pointer ptr{allocateBlock(released_bytes)};
-      insertFree(ptr, released_bytes, true, released_bytes);
+    if (size_post < suggested_size) {
+      std::size_t alloc_size{suggested_size - size_post};
```
davidbeckingsale, line 274: Do you think we should check if alloc_size is reasonable?

mcfadden8: Good question. What is reasonable?

davidbeckingsale: I think it should be _at least_ the minimum block size (I think that will happen anyway, but good to check)

mcfadden8: @davidbeckingsale, I am assuming we resolved this in our discussion yesterday.  Please let me know if you are still looking for changes.  Thx

davidbeckingsale: I think the decision was made. I don't think we need to enforce anything, but can you please add a log message here that prints the alloc_size to Debug?

mcfadden8: Do you mean like on line 275?

davidbeckingsale (approved):

tools/replay/ReplayOperationManager.cpp
```diff
@@ -14,7 +14,9 @@
 #include "umpire/Allocator.hpp"
 #include "umpire/strategy/AllocationAdvisor.hpp"
 #include "umpire/strategy/AllocationPrefetcher.hpp"
+#include "umpire/strategy/PoolCoalesceHeuristic.hpp"
 #include "umpire/strategy/SizeLimiter.hpp"
+#include "umpire/strategy/QuickPool.hpp"
```
kab163, line 19: Seeing as you didn't also need to include DynamicPoolList here, is adding QuickPool necessary?

src/umpire/strategy/DynamicSizePool.hpp
```diff
@@ -432,13 +432,17 @@ class DynamicSizePool : private umpire::strategy::mixins::AlignedAllocation {
     return nb;
   }
 
-  void coalesce()
+  void coalesce(std::size_t suggested_size)
   {
     if (getFreeBlocks() > 1) {
-      std::size_t size_to_coalesce = freeReleasedBlocks();
+      freeReleasedBlocks();
```
kab163, line 438: Maybe I'm missing something, but I'm confused how before this function returned some size_t value and now there is no return value (or it is being ignored). Do you see what I mean?

kab163 (commented):
Looks good, just had a few questions. This is exciting though :)

src/umpire/strategy/DynamicSizePool.hpp
mcfadden8: Good question.  `freeReleasedBlocks()` returns the number of bytes that were freed.  The coalesce method is now concerned about what the total size of the pool should be.  So, we call `getActualSize()` after `freeReleasedBlocks()` to determine the new size of the pool.  If that size is smaller than `suggested_size`, we then allocate a block big enough to increase the pools size.

kab163: Hmm I see. So it is purposely ignored.. could the number of bytes that were freed be useful at all (like in the logging statement that follows)? But ya, that makes sense. Thanks for clarifying!

PR #605, opened 2021-07-20
Add async memset and reallocate for CUDA & HIP

davidbeckingsale:


src/umpire/ResourceManager.cpp
```diff
@@ -444,6 +444,36 @@ void ResourceManager::memset(void* ptr, int value, std::size_t length)
...
+  auto alloc_record = m_allocations.find(ptr);
+
+  std::ptrdiff_t offset = static_cast<char*>(ptr) - static_cast<char*>(alloc_record->ptr);
+  std::size_t size = alloc_record->size - offset;
+
+  if (length == 0) {
+    length = size;
+  }
+
+  UMPIRE_REPLAY(R"( "event": "memset", "payload": { )"
```
mcfadden8, line 462: Side node: While recording these events is perhaps important, we should have a review about whether replay of these kinds of operations are appropriate.  The replay tool currently does not fully support operations on memory that is already allocated (other than `realloc`).

src/umpire/ResourceManager.cpp
```diff
@@ -444,6 +444,36 @@ void ResourceManager::memset(void* ptr, int value, std::size_t length)
...
+                << R"( "ptr": ")" << ptr << R"(")"
+                << R"(, "value": )" << value << R"(, "size": )" << size << R"(, "allocator_ref": ")"
```
mcfadden8, line 464: It we do decide to replay these, we will need to add a `ctx_ref` here to make that possible.

src/umpire/ResourceManager.cpp
```diff
@@ -469,6 +499,31 @@ void* ResourceManager::reallocate(void* current_ptr, std::size_t new_size)
...
+  UMPIRE_REPLAY(R"( "event": "reallocate", "payload": {)"
+                << R"( "ptr": ")" << current_ptr << R"(")"
+                << R"(, "size": )" << new_size << R"(, "allocator_ref": ")" << strategy << R"(" } )");
+
+  void* new_ptr{reallocate_impl(current_ptr, new_size, Allocator(strategy), ctx)};
+
+  UMPIRE_REPLAY(R"( "event": "reallocate", "payload": {)"
+                << R"( "ptr": ")" << current_ptr << R"(")"
+                << R"(, "size": )" << new_size << R"(, "allocator_ref": ")" << strategy << R"(" } )"
+                << R"(, "result": { "memory_ptr": ")" << new_ptr << R"(" } )");
```
mcfadden8, line 522: Similarly, need to record the `ctx_ref`

mcfadden8 (approved):

PR #627, opened 2021-08-23
Add get_device_allocator function

kab163:


examples/cookbook/recipe_device_allocator.cpp
```diff
@@ -0,0 +1,98 @@
...
+__global__ void my_kernel(int* result)
+{
+  unsigned int i{threadIdx.x+blockIdx.x*blockDim.x};
+
+  if (i == 0) {
+    umpire::DeviceAllocator alloc = umpire::util::getDeviceAllocator(0);
+    data = static_cast<int*>(alloc.allocate(BLOCKS * sizeof(int)));
+  }
+
+  data[blockIdx.x] = blockIdx.x;
```
mcfadden8: What happens when threads with non-zero ids (`i`) write to `data` before data is allocated by the thread with id of zero?

examples/cookbook/recipe_device_allocator.cpp
```diff
@@ -0,0 +1,98 @@
...
+/*
+ * Kernel that performs simple arithmetic and stores the
+ * result in the array of ints passed to the kernel.
+ *
+ * Note: Benign data race that writes to the device allocated
+ * array, data.
+ */
+__global__ void my_kernel(int* result)
+{
+  unsigned int i{threadIdx.x+blockIdx.x*blockDim.x};
```
mcfadden8:
```suggestion
  unsigned int my_id{threadIdx.x+blockIdx.x*blockDim.x};
```

examples/cookbook/recipe_device_allocator.cpp
```diff
@@ -0,0 +1,98 @@
...
+  //create an array of integers which will be used by the gpu to do some simple arithmetic
+  int* results = static_cast<int*>(allocator.allocate(BLOCKS*THREADS_PER_BLOCK * sizeof(int)));
+  memset(results, 0, BLOCKS*THREADS_PER_BLOCK);
+
+  //Sync up the device and host side global pointers that keep track of DeviceAllocator objects
+  UMPIRE_SYNC_DEVICE_ALLOCATORS();
+
+  //call kernel and sync
+  my_kernel<<<BLOCKS, THREADS_PER_BLOCK>>>(results);
+  cudaDeviceSynchronize();
```
mcfadden8: Shouldn't these lines be performed by the Umpire library so that the application does not need to perform vendor-specific calls like `cudaDeviceSynchronize()`?

examples/device-allocator.cpp
```diff
@@ -4,29 +4,66 @@
...
+  }
+}
+
 int main(int argc, char const* argv[])
 {
   auto& rm = umpire::ResourceManager::getInstance();
+
+  //Checking to make sure a DeviceAllocator doesn't yet exist
+  if(!umpire::util::existsDeviceAllocator()) {
+    std::cout << "Before I create a DeviceAllocator, it doesn't exist!" << std::endl;
```
mcfadden8: Is this an error or is this the expected case?  I'm not sure what the message means.  It appears that this is information-only, but I am not sure.  If it is information-only, then perhaps an else-block below with a different output might also be useful.

src/umpire/ResourceManager.cpp
```diff
@@ -327,6 +328,23 @@ void ResourceManager::registerAllocation(void* ptr, util::AllocationRecord recor
   m_allocations.insert(ptr, record);
 }
 
+#if defined(UMPIRE_ENABLE_CUDA)
+DeviceAllocator ResourceManager::makeDeviceAllocator(Allocator allocator, size_t size, const char* name)
+{
+  static size_t i{0};
```
mcfadden8:
```suggestion
  static size_t allocator_id{0};
```

src/umpire/util/device_allocator_helper.cu
```diff
@@ -0,0 +1,32 @@
...
+int UMPIRE_TOTAL_DEV_ALLOCS_h{10};
+int UMPIRE_DEV_ALLOCS_COUNTER_h{0};
+DeviceAllocator* UMPIRE_DEV_ALLOCS_h{nullptr};
+__device__ DeviceAllocator* UMPIRE_DEV_ALLOCS{nullptr};
+
+__device__ DeviceAllocator getDeviceAllocator(const char* name)
+{
+  for (int i = 0; i < 10; i++)
+  { 
+    if(umpire::util::UMPIRE_DEV_ALLOCS[i].getName() == name) {
```
mcfadden8: Just a nit, please treat so :-).  But, how come I sometimes see spaces between a key work and paren (`for (...`) and sometimes not (`if(...`))?

src/umpire/DeviceAllocator.hpp
```diff
@@ -38,14 +29,41 @@ class DeviceAllocator {
...
+   */
+  __host__ __device__ size_t getID();
+
+  /*
+   * \brief Get the name associated with the
+   *  current DeviceAllocator object.
+   *
+   * \return Returns the name. 
+   */
+  __device__ const char* getName();
```
davidbeckingsale: Make host/device.

src/umpire/util/device_allocator_helper.cu
kab163: Good point. That's probably why the style check is failing lol

src/umpire/ResourceManager.cpp
kab163: Done!

examples/device-allocator.cpp
kab163: I reworked this function to take a id or name so that it can be used on a specific DeviceAllocator object. That changed the way that I queried it in the example too. (I only check for it after I've created a DA, then query it.)

examples/cookbook/recipe_device_allocator.cpp
kab163: I've removed this recipe example all together.

examples/cookbook/recipe_device_allocator.cpp
kab163: I removed this recipe file, though I will make this change to the example I have instead.

src/umpire/DeviceAllocator.hpp
kab163: Done

examples/cookbook/recipe_device_allocator.cpp
kab163: (Update: No change, since this doesn't apply for the example file - I am only using one block and I use threadIdx.x)

examples/cookbook/recipe_device_allocator.cpp
kab163: I think I put in something that fixes this problem. It at least hides the backend stuff. We can work on polishing up the new stuff next (basically, i added a function that will call cuda(hip)DeviceSynchronize depending on what's enabled)

src/umpire/util/device_allocator_helper.cu
kab163: I ran clang-format and added the changes it generated, hopefully that fixed all the style stuff

kab163:
There were a few things I think I still haven't included from our meeting yesterday. From what I remember there was:
- [x] Using the camp resource to distinguish between cuda vs hip (etc.) 
- [x] Making sure destructor works as expected
- [x] Making the `umpire_device` object library in CMake
- [x] Make sure it passes the style check

src/umpire/CMakeLists.txt
```diff
@@ -86,15 +86,21 @@ endif ()
...
   set(umpire_sources
     ${umpire_sources}
-    DeviceAllocator.cpp)
+    DeviceAllocator.cpp
+    device_allocator_helper.cpp)
 
   set_source_files_properties(DeviceAllocator.cpp
                               PROPERTIES
                               LANGUAGE CUDA)
+  set_source_files_properties(device_allocator_helper.cpp
```
davidbeckingsale, line 49: These two files should go in a new `umpire_device` object library (like what you did with `umpire_interface`)

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace umpire {
+
+const int UMPIRE_TOTAL_DEV_ALLOCS_h{10};
+__device__ const int UMPIRE_TOTAL_DEV_ALLOCS{10};
+
+DeviceAllocator* UMPIRE_DEV_ALLOCS_h{nullptr};
+__device__ DeviceAllocator* UMPIRE_DEV_ALLOCS{nullptr};
+
+__device__ DeviceAllocator getDeviceAllocator(const char* name)
```
davidbeckingsale: This can be a host/device function, since there's no reason it shouldn't work on the host.

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+  for (int i = 0; i < UMPIRE_TOTAL_DEV_ALLOCS; i++) {
+    if (UMPIRE_DEV_ALLOCS[i].getName() == name) {
+      return UMPIRE_DEV_ALLOCS[i];
+    }
+  }
+  // For now, if we can't find the specified DeviceAllocator,
+  // return the first DeviceAllocator.
+  return UMPIRE_DEV_ALLOCS[0];
```
davidbeckingsale: This should crash. You can use the UMPIRE_ERROR macro.

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+__device__ DeviceAllocator getDeviceAllocator(int id)
```
davidbeckingsale: Needs error checking (make sure id not bigger than max), and can be host/device.

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+  return UMPIRE_DEV_ALLOCS[id];
+}
+
+DeviceAllocator makeDeviceAllocator(Allocator allocator, size_t size, const char* name)
```
davidbeckingsale: This should be marked host.

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+  static size_t allocator_id{0};
+  auto dev_alloc = DeviceAllocator(allocator, size, name, allocator_id);
+
+  if (allocator_id == 0) {
+#if defined(UMPIRE_ENABLE_CUDA)
+    auto um_alloc = umpire::alloc::CudaMallocManagedAllocator();
```
davidbeckingsale:
```suggestion
    auto& rm = umpire::ResourceManager::getInstance();
    auto alloc = rm.getAllocator("UM");
```

davidbeckingsale: You can then remove the ifdefs.

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,89 @@
...
+void destroyDeviceAllocator()
+{
+  for (int i = 0; i < 10; i++) {
+    if (UMPIRE_DEV_ALLOCS_h[i].isInitialized()) {
+      UMPIRE_DEV_ALLOCS_h[i].destroy();
+    }
+  }
+}
+
+void synchronizeDeviceAllocator()
```
davidbeckingsale: I think we can remove this and use `camp::resources` instead.

src/umpire/device_allocator_helper.hpp
```diff
@@ -0,0 +1,84 @@
...
+#define UMPIRE_SET_UP_DEVICE_ALLOCATORS()                                                                          \
+  {                                                                                                                \
+    cudaMemcpyToSymbol(umpire::UMPIRE_DEV_ALLOCS, &umpire::UMPIRE_DEV_ALLOCS_h, sizeof(umpire::DeviceAllocator*)); \
+  }
+#elif defined(UMPIRE_ENABLE_HIP)
+#define UMPIRE_SET_UP_DEVICE_ALLOCATORS()                                                                         \
+  {                                                                                                               \
+    hipMemcpyToSymbol(umpire::UMPIRE_DEV_ALLOCS, &umpire::UMPIRE_DEV_ALLOCS_h, sizeof(umpire::DeviceAllocator*)); \
+  }
+#endif
```
davidbeckingsale, line 74: Add an empty macro for else, so people can leave this in their code:
```suggestion
#else
#define UMPIRE_SET_UP_DEVICE_ALLOCATORS()
#endif
```

davidbeckingsale (changes requested):

src/umpire/device_allocator_helper.cpp
kab163: done

src/umpire/device_allocator_helper.cpp
kab163: done

src/umpire/device_allocator_helper.cpp
kab163: done

src/umpire/device_allocator_helper.cpp
kab163: Done

src/umpire/device_allocator_helper.cpp
kab163: Are you sure? Once I change it to host/device, I'll have to make other changes so that it doesn't call a host function in a device func. etc. So like making a separate getDeviceAllocator that is just for the host so that I query all the host variables instead

src/umpire/device_allocator_helper.cpp
kab163: I see how to make use of the `camp::resources` stuff, but I don't see where there's a cudaDeviceSynchronize() handler there. I want to do something like:
`#if defined(UMPIRE_ENABLE_CUDA)`
  `using resource_type = camp::resources::Cuda;`
`#elif defined(UMPIRE_ENABLE_HIP)`
  `using resource_type = camp::resources::Hip;`
`#endif`

`auto resource = camp::resources::Resource{resource_type};`
`resource.synchronize();`

src/umpire/CMakeLists.txt
kab163: Ok - I put this off because I was starting to do that in another ticket which handled umpire_interface. I will instead remove all the DeviceAllocator stuff from that other ticket and instead do it here.

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,95 @@
...
+__device__ DeviceAllocator* UMPIRE_DEV_ALLOCS{nullptr};
+
+__device__ DeviceAllocator getDeviceAllocator(const char* name)
+{
+  for (int i = 0; i < UMPIRE_TOTAL_DEV_ALLOCS; i++) {
+    if (UMPIRE_DEV_ALLOCS[i].getName() == name) {
+      return UMPIRE_DEV_ALLOCS[i];
+    }
+  }
+  UMPIRE_ERROR("No DeviceAllocator by the name " << name << " was found.");
```
kab163: This creates a `missing return statement at end of non-void function` warning. I remember we handled this some way before, but I forget what exactly. @mcfadden8, didn't you find a way to avoid this?

mcfadden8: https://github.com/LLNL/Umpire/blob/develop/src/umpire/resource/MemoryResourceTypes.hpp#L48-L58

src/umpire/device_allocator_helper.cpp
kab163: Done

src/umpire/device_allocator_helper.cpp
kab163: Done - i think lol

src/umpire/CMakeLists.txt
kab163: I put in the first attempt at attacking that. Hopefully I did it right lol

kab163:
Things I might still need to add:

- [x] Test to check functionality and error checking for getDeviceAllocator
- [x] Documentation! - This will be for a new ticket!
- [x] JIRA ticket to set up HIP functionality?

src/umpire/CMakeLists.txt
```diff
@@ -217,6 +240,10 @@ install(FILES
   ${umpire_headers}
   DESTINATION include/umpire)
 
+install(FILES
```
kab163: When doing the umpire_interface cmake changes (see PR #631 ), I had to add a `OBJECTS DESTINATION` line for the install line (included in something like this https://github.com/LLNL/Umpire/blob/020372220fd6e59c1e085f62301b296d7170e344/src/umpire/CMakeLists.txt#L232-L237) - is that needed here too, or is this good enough?

src/umpire/CMakeLists.txt
```diff
@@ -114,6 +136,7 @@ endif ()
 
 blt_list_append( TO umpire_depends ELEMENTS umpire_interface IF ENABLE_C )
 blt_list_append( TO umpire_depends ELEMENTS cuda IF ENABLE_CUDA )
```
davidbeckingsale: I think this line can be removed, since only `umpire_device` needs to depend on CUDA directly.

src/umpire/CMakeLists.txt
davidbeckingsale: As discussed, no object install is required.

src/umpire/DeviceAllocator.cpp
```diff
@@ -6,40 +6,63 @@
...
 #include "umpire/ResourceManager.hpp"
+#include "umpire/device_allocator_helper.hpp"
 #include "umpire/resource/MemoryResourceTypes.hpp"
 #include "umpire/util/Macros.hpp"
 
 namespace umpire {
 
-__host__ DeviceAllocator::DeviceAllocator(Allocator allocator, size_t size)
-    : m_allocator(allocator), m_ptr(static_cast<char*>(m_allocator.allocate(size))), m_size(size), m_child(false)
+__host__ DeviceAllocator::DeviceAllocator(Allocator allocator, size_t size, const char* name, size_t id)
```
davidbeckingsale: I would prefer if we could use std::string in the API and handle the conversion to char* internally.

src/umpire/DeviceAllocator.hpp
```diff
@@ -17,15 +19,18 @@ namespace umpire {
...
+   * data.
    *
    * \param allocator Allocator to use for allocating memory.
+   * \param size Total amount of memory that the DeviceAllocator will have.
+   * \param name Name of the DeviceAllocator object.
+   * \param id ID associated with this DeviceAllocator object. ID will be used
+   *   to reference UMPIRE_DEV_ALLOCS array if necessary.
    */
-  __host__ DeviceAllocator(Allocator allocator, size_t size);
+  __host__ DeviceAllocator(Allocator allocator, size_t size, const char* name, size_t id);
```
davidbeckingsale: Use std::string

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,119 @@
...
+DeviceAllocator* UMPIRE_DEV_ALLOCS_h{nullptr};
+__device__ DeviceAllocator* UMPIRE_DEV_ALLOCS{nullptr};
+
+//////////////////////////////////////////////////////////////////////////
+// host/device functions
+//////////////////////////////////////////////////////////////////////////
+__host__ __device__ DeviceAllocator getDeviceAllocator(const char* name)
+{
+  int index = findDeviceAllocatorID(name);
+  if (index == -1)
```
davidbeckingsale: { } around the statement after the conditional.

src/umpire/device_allocator_helper.cpp
```diff
@@ -0,0 +1,119 @@
...
+        tally++;
+    } while (name[index++] != 0);
+    if (tally == 0)
+      return i;
+  }
+  return -1;
+#endif
+}
+
+__host__ __device__ bool deviceAllocatorExists(int id)
```
davidbeckingsale: I think this should be called `isDeviceAllocator` to match our other functions.

examples/device-allocator.cpp
```diff
@@ -4,29 +4,81 @@
...
+  my_kernel<<<1, 16>>>(ptr_to_data);
+  resource.get_event().wait();
+  std::cout << "After first kernel, found value: " << (*ptr_to_data)[7] << std::endl;
+
+  my_other_kernel<<<1, 16>>>(ptr_to_data);
+  resource.get_event().wait();
+  std::cout << "After second kernel, found value: " << (*ptr_to_data)[0] << std::endl;
+
+  // Tear down and deallocate memory.
+  umpire::destroyDeviceAllocator();
```
davidbeckingsale: This function should be called by the ResourceManager destructor, not manually by the user.

src/umpire/device_allocator_helper.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * DeviceAllocator objects created.
+ */
+extern DeviceAllocator* UMPIRE_DEV_ALLOCS_h;
+__device__ extern DeviceAllocator* UMPIRE_DEV_ALLOCS;
+
+/*
+ * Get the DeviceAllocator object specified by either the given
+ * name or id.
+ */
+__host__ __device__ DeviceAllocator getDeviceAllocator(const char* name);
```
davidbeckingsale: Can we make these snake_case?

src/umpire/device_allocator_helper.hpp
```diff
@@ -0,0 +1,83 @@
...
+/*
+ * Check if the DeviceAllocator object specified by the
+ * given id currently exists.
+ */
+__host__ __device__ bool deviceAllocatorExists(int id);
+
+/*
+ * Given a name, find the corresponding ID, if it exists.
+ */
+__host__ __device__ int findDeviceAllocatorID(const char* name);
```
davidbeckingsale: I don't think this function needs to be exposed, you can just have the implementation inside the getDeviceAllocator(name) function.

davidbeckingsale (changes requested):

src/umpire/CMakeLists.txt
kab163: Update: not needed

src/umpire/device_allocator_helper.cpp
kab163: Good idea

examples/device-allocator.cpp
kab163: There was a discussion at one point that we wanted to eventually move away from the ResourceManager. This will make it go from being completely independent of the RM to not (so it will be created by the user manually not using the RM, but destroyed automatically with the RM.) Is that ok?

davidbeckingsale: Good point. If we remove the resource manager from the user interface we will still have some kind of "shutdown" mechanism to do stuff like this, so I think moving it for now is a good idea so we don't need the user to call this function.

kab163: Mkay!

src/umpire/device_allocator_helper.hpp
kab163: Done for the device_allocator_helper functions, the DeviceAllocator member functions stayed the same

src/umpire/device_allocator_helper.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "umpire/DeviceAllocator.hpp"
+
+namespace umpire {
+
+/*
+ * Const variable for the limit of unique DeviceAllocator
+ * objects available at once.
+ */
+extern __managed__ int UMPIRE_TOTAL_DEV_ALLOCS;
```
davidbeckingsale:
```suggestion
constexpr int UMPIRE_TOTAL_DEV_ALLOCS{64};
```

src/umpire/DeviceAllocator.cpp
```diff
@@ -6,40 +6,66 @@
...
 #include "umpire/ResourceManager.hpp"
+#include "umpire/device_allocator_helper.hpp"
 #include "umpire/resource/MemoryResourceTypes.hpp"
 #include "umpire/util/Macros.hpp"
 
 namespace umpire {
 
-__host__ DeviceAllocator::DeviceAllocator(Allocator allocator, size_t size)
-    : m_allocator(allocator), m_ptr(static_cast<char*>(m_allocator.allocate(size))), m_size(size), m_child(false)
+__host__ DeviceAllocator::DeviceAllocator(Allocator allocator, size_t size, const std::string old_name, size_t id)
```
davidbeckingsale: I think this should be `const std::string&`

davidbeckingsale (changes requested):
Just those 2 small edits and this should be good to go!

src/umpire/device_allocator_helper.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * DeviceAllocator objects created.
+ */
+extern DeviceAllocator* UMPIRE_DEV_ALLOCS_h;
+__device__ extern DeviceAllocator* UMPIRE_DEV_ALLOCS;
+
+/*
+ * Get the DeviceAllocator object specified by either the given
+ * name or id.
+ */
+__host__ __device__ DeviceAllocator get_device_allocator(const char* name);
```
davidbeckingsale, line 33: This should be a `const std::string&`

davidbeckingsale (changes requested):
Found one last thing! 😄

davidbeckingsale (approved):

PR #630, opened 2021-08-25
Replace umpire::replay with umpire::event

mcfadden8:
This update replaces umpire::replay with a new and improved umpire::event mechanism.

The umpire::event mechanism can record events to JSON files (the default) or SQLITE (enabled with `+sqlite` spack variant).

During development, other updates were added including:

1. Update nlohmann/json to 3.10.4
2. Benchmark scripts were added to compare cost of millions of allocations between branches
3. Allocate and makeAllocate events were changed to record only one event after successful allocation/creation instead of two events for before and after attempt to perform request


src/umpire/Allocator.inl
```diff
@@ -42,8 +40,9 @@ inline void* Allocator::allocate(std::size_t bytes)
     registerAllocation(ret, bytes, m_allocator);
   }
 
-  UMPIRE_REPLAY("\"event\": \"allocate\", \"payload\": { \"allocator_ref\": \""
-                << m_allocator << "\", \"size\": " << bytes << " }, \"result\": { \"memory_ptr\": \"" << ret << "\" }");
+  umpire::event::record<umpire::event::allocate>(
+      [&](auto& event) { event.size(bytes).ref((void*)m_allocator).ptr(ret); });
+
```
kab163, line 45:
```suggestion
  umpire::event::record<umpire::event::allocate>([&](auto& event) { 
    event.size(bytes)
             .ref((void*)m_allocator)
             .ptr(ret); 
   });
```

kab163: This is over the top annoying to try and do in the github suggestion syntax... this is what I'm proposing for all the `umpire::event::record` calls to be structured like.

kab163: (But without the weird spacing... the `.size` and `.ref` etc. should all be at the same indentation)

mcfadden8: I think I addressed these (there were others too).  Please check and mark resolved if you agree.

kab163:
Is there any documentation added with this PR?

kab163:
Is the way you build and run with this the same as running with replay? (Besides turning on the `UMPIRE_ENABLE_SQLITE=On` flag)

mcfadden8:
> Is there any documentation added with this PR?

Not really no.  From the end-user perspecitive umpire will still generate a replay file and the replay tool will still perform replays from it.  The default replay file is still in json format.

mcfadden8:
> Is the way you build and run with this the same as running with replay? (Besides turning on the `UMPIRE_ENABLE_SQLITE=On` flag)

Yes, everything would look and feel the same (except MUCH SLOWER).  The only difference will be that Umpire will be generating a single sqlite file instead of a json file and neither are intented to be read or understood by the end user.

kab163 (dismissed):

davidbeckingsale (dismissed):

mcfadden8:
@kab163 and @davidbeckingsale, we might want to change `UMPIRE_ENABLE_SQLITE` to a different name that indicates that it is experimental.  The reason SQLITE support is important is because it proves that the storage subsystem of event can read/write to multiple types of storage.  However, sqlite really should not be used by users as it is significantly slower.  I hadn't worried about this because most of our users leave options disabled unless they absolutely need something.

Perhaps I could update the read the docs page for SQLITE to say that it is EXPERIMENTAL?

kab163:
You could also have it print something out in the cmake that warns it is experimental like you do with the shared memory stuff

mcfadden8:
> You could also have it print something out in the cmake that warns it is experimental like you do with the shared memory stuff

Really?  Is shared memory still marked as experimental?  I beleive there are active users of it now

kab163:
> > You could also have it print something out in the cmake that warns it is experimental like you do with the shared memory stuff
> 
> Really? Is shared memory still marked as experimental? I beleive there are active users of it now

No, I mean the cmake outputs something each time.. sorry lol that's all I meant. Like `-- Host Shared Memory Disabled` is something that's output, and if it's enabled, it outputs something else - but it's a message to the user. So I guess my suggestion was "if sqlite is enabled, then output warning message in cmake to user that it is `EXPERIMENTAL`".

mcfadden8:
> > > You could also have it print something out in the cmake that warns it is experimental like you do with the shared memory stuff
> > 
> > 
> > Really? Is shared memory still marked as experimental? I beleive there are active users of it now
> 
> No, I mean the cmake outputs something each time.. sorry lol that's all I meant. Like `-- Host Shared Memory Disabled` is something that's output, and if it's enabled, it outputs something else - but it's a message to the user. So I guess my suggestion was "if sqlite is enabled, then output warning message in cmake to user that it is `EXPERIMENTAL`".

But will users even see it?  Most of our users are spack users and I think `+sqlite` will just hide all of that.

kab163:
> > > > You could also have it print something out in the cmake that warns it is experimental like you do with the shared memory stuff
> > > 
> > > 
> > > Really? Is shared memory still marked as experimental? I beleive there are active users of it now
> > 
> > 
> > No, I mean the cmake outputs something each time.. sorry lol that's all I meant. Like `-- Host Shared Memory Disabled` is something that's output, and if it's enabled, it outputs something else - but it's a message to the user. So I guess my suggestion was "if sqlite is enabled, then output warning message in cmake to user that it is `EXPERIMENTAL`".
> 
> But will users even see it? Most of our users are spack users and I think `+sqlite` will just hide all of that.

Good point...

kab163 (dismissed):

davidbeckingsale (approved):

PR #631, opened 2021-09-02
Move umpire_interface CMake to the umpire/interface subdirectory

kab163:
This reorganizes the cmake includes for the c and fortran related code so that there is an object library that gets compiled in. It also creates a separate CMakeLists for the interface sub-directory.


davidbeckingsale (changes requested):
Please make a note in the CHANGELOG about the include paths having to change - this will be a breaking change for users.

src/umpire/interface/c_fortran/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+  OBJECT TRUE)
+
+target_include_directories(
+  umpire_interface
+  PUBLIC
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
+  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
+  $<INSTALL_INTERFACE:include>)
+
+install(TARGETS
```
davidbeckingsale, line 55: I don't think you need this, since we are linking this into the uber libumpire. If you remove this you can also remove the `OBJECTS DESTINATION lib` line below.

src/umpire/interface/c_fortran/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+# Copyright (c) 2016-20, Lawrence Livermore National Security, LLC and Umpire
+# project contributors. See the COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (MIT)
+##############################################################################
+set (umpire_interface_c_fortran_headers
+  wrapAllocator.h
+  wrapResourceManager.h
+  wrapUmpire.h
+  umpire.h
```
davidbeckingsale, line 11: Was there a particular reason you moved this file?

kab163: Hmm off the top of my head, I'm not sure, I'll have to remind myself! :O

davidbeckingsale: Not a problem - it's more consistent this way.

src/umpire/interface/c_fortran/CMakeLists.txt
kab163: I got rid of line 53 above and the `OBJECTS DESTINATION lib` line.

kab163: Update: that still caused errors only in the gcc8 docker build. Not sure why, but I reverted it back to what it was. Not sure how I should proceed here.

davidbeckingsale (approved):

PR #636, opened 2021-09-03
Remove deprecated isAllocatorRegistered/registerAllocator methods

davidbeckingsale:


kab163:
Does this PR do a bit more than just remove the deprecated functions? It seems like it also updates the comment headers and a few function names. Is that ok to keep in this PR or should it be separate?

davidbeckingsale:
> Does this PR do a bit more than just remove the deprecated functions? It seems like it also updates the comment headers and a few function names. Is that ok to keep in this PR or should it be separate?

Hey @kab163 - we discussed the change (caused by updating Shroud) yesterday. It doesn't change any user APIs, apart from the deprecated functions being explicitly removed. Can you please approve?

kab163 (approved):

PR #637, opened 2021-09-03
Set CUDA_RESOLVE_DEVICE_SYMBOLS on Umpire libraries

davidbeckingsale:

PR #639, opened 2021-09-07
nvcc compile flags for use with CUDA_RESOLVE_DEVICE_SYMBOLS 

PhilipDeegan:
test #637 with rdc=true

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 
 ### Changed
 
+- Set property CUDA_RESOLVE_DEVICE_SYMBOLS on Umpire library.
+
 ### Removed
 
 - Removed unneeded hip dependency in the tests/debug/ CMake file.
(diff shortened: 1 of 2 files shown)


PhilipDeegan:
The error is related to CMake INTERFACE link properties, which I don't see with CMake 3.21.2

but this modification was required to not have link errors later, or when building shared libaries (which seems like the best thing to do with this IMO)

I tried
```
target_compile_options(umpire PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-rdc=true>)
```

but it didn't appear to add the flag

white238:
I'm not sure why that generator expression did not work @PhilipDeegan .  Can you try the following macro to figure out if for some reason the target properties are causing you a problem?

```
blt_print_target_properties(TARGET umpire)
```

This will print all target properties when you run cmake.

davidbeckingsale:
Base on discussion with NVIDIA this is not the correct approach, so I'm closing this PR and #637

PhilipDeegan:
I think I only needed to build shared libs for umpire anyway

PR #650, opened 2021-09-16
Add a DeviceAllocator reset function

kab163:


davidbeckingsale (dismissed):

davidbeckingsale (changes requested):
I think we ought to have a test or example that uses this function, both on the host, and on the device.

kab163:
> I think we ought to have a test or example that uses this function, both on the host, and on the device.

I added an example of calling the reset function on the host. I think it'd be better to hold off on the device until I come up with a more useful example. (Right now, it would look the exact same as the host call since I'm only using one thread to do the DA stuff on the GPU.) What do you think?

davidbeckingsale (approved):

PR #657, opened 2021-09-22
Add 0 value for rtype in enum

davidbeckingsale:
This allows us to use memset to initialize AllocatorTableEntry to all 0.

Prior to this change, this warning would be triggered in GCC 10.2.1:
`error: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type`

The warning is valid, since prior to this change, 0 was not a valid value
for the rtype enum.


mcfadden8 (approved):
Thanks!

PR #662, opened 2021-10-04
Add documentation for the device allocator

kab163:
Documented the important aspects of the DeviceAllocator (all except the Hip support functionality which is still in progress).

https://umpire.readthedocs.io/en/task-um-944-document-device-allocator/sphinx/features/device_allocators.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 
 - Added a reset function to the DeviceAllocator so that old data can be rewritten.
 
+- Added documentation of the DeviceAllocator.
+
 ### Changed
 
 - Reorganized cmake object library for c/fortran interface. NOTE: This is a breaking
--- a/docs/sphinx/features/device_allocators.rst
+++ b/docs/sphinx/features/device_allocators.rst
@@ -1,0 +1,79 @@
+.. _device_allocators:
+
+=================
+Device Allocators
+=================
+
+The DeviceAllocator is designed for memory allocations on the GPU. 
+Currently there is only support for CUDA, although HIP support is coming soon.
+
+Creating a Device Allocator
+--------------------------
+
+To create a DeviceAllocator, users can call the :class:`umpire::make_device_allocator` host function.
+This function takes an allocator, the total amount of memory the DeviceAllocator will have, and a name
+for the new DeviceAllocator object, as shown below. Currently, a maximum of 64 unique DeviceAllocators can be
+created at a time.
+
+.. literalinclude:: ../../../examples/device-allocator.cpp
+   :start-after: _sphinx_tag_make_dev_allocator_start
+   :end-before: _sphinx_tag_make_dev_allocator_end
+   :language: C++
+
+When the DeviceAllocator is created, the ``size`` parameter that is passed to the :class:`umpire::make_device_allocator`
+function is the total memory, in bytes, available to that allocator. Whenever the ``allocate`` function is
+called on the GPU, it is simply atomically incrementing a counter which offsets a pointer to the start of that
+memory.
+
+Retrieving a DeviceAllocator Object
+-----------------------------------
+
+The :class:`umpire::get_device_allocator` host/device function returns the DeviceAllocator object that corresponds 
+to the ID or name given. The DeviceAllocator class also includes a helper function, :class:`umpire::is_device_allocator`,
+to query whether or not a given ID corresponds to an existing DeviceAllocator. Below is an example of using the **ID** to 
+obtain the DeviceAllocator object:
+
+.. literalinclude:: ../../../examples/device-allocator.cpp
+   :start-after: _sphinx_tag_get_dev_allocator_id_start
+   :end-before: _sphinx_tag_get_dev_allocator_id_end
+   :language: C++
+
+And next is an example of using the **name** instead:
+
+.. literalinclude:: ../../../examples/device-allocator.cpp
+   :start-after: _sphinx_tag_get_dev_allocator_name_start
+   :end-before: _sphinx_tag_get_dev_allocator_name_end
+   :language: C++
+
+With the :class:`umpire::get_device_allocator` function, there is no need to keep track of a DeviceAllocator, since function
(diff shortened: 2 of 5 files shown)


davidbeckingsale (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 
 - Added a reset function to the DeviceAllocator so that old data can be rewritten.
 
+- Added documentation of the DeviceAllocator.
+
```
mcfadden8, line 19: @davidbeckingsale, @kab163, do we still have a CHANGELOG?

kab163: 💩

docs/sphinx/features/device_allocators.rst
```diff
@@ -0,0 +1,79 @@
+.. _device_allocators:
+
+=================
+Device Allocators
+=================
+
+The DeviceAllocator is designed for memory allocations on the GPU. 
+Currently there is only support for CUDA, although HIP support is coming soon.
```
mcfadden8, line 8: I would be wary of stating what is coming in the future versus what is here now.  These types of sentences can get lost/forgotten and then they can inadvertently become stale.  I suggest we remove this sentence.

docs/sphinx/features/device_allocators.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Creating a Device Allocator
+--------------------------
+
+To create a DeviceAllocator, users can call the :class:`umpire::make_device_allocator` host function.
+This function takes an allocator, the total amount of memory the DeviceAllocator will have, and a name
+for the new DeviceAllocator object, as shown below. Currently, a maximum of 64 unique DeviceAllocators can be
```
mcfadden8, line 15:
```suggestion
for the new DeviceAllocator object, as shown below. A maximum of 64 unique DeviceAllocators can be
```

docs/sphinx/features/device_allocators.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. literalinclude:: ../../../examples/device-allocator.cpp
+   :start-after: _sphinx_tag_make_dev_allocator_start
+   :end-before: _sphinx_tag_make_dev_allocator_end
+   :language: C++
+
+When the DeviceAllocator is created, the ``size`` parameter that is passed to the :class:`umpire::make_device_allocator`
+function is the total memory, in bytes, available to that allocator. Whenever the ``allocate`` function is
+called on the GPU, it is simply atomically incrementing a counter which offsets a pointer to the start of that
+memory.
```
mcfadden8, line 26: I'm not sure that discussing "how" the allocation mechanism works is necessary.  It might be better to simply say:
"The total size from all of the allocates performed on the device may not exceed the size used when making the device allocator".

docs/sphinx/features/device_allocators.rst
```diff
@@ -0,0 +1,79 @@
...
+With the :class:`umpire::get_device_allocator` function, there is no need to keep track of a DeviceAllocator, since function
+call stacks can become quite complex. Users can instead use this function to obtain it inside whichever host or device
+function they need.
+
+Under the hood, the :class:`umpire::get_device_allocator` uses global arrays which can be accessed by both 
+the host and device. The global array is indexed by DeviceAllocator ID, which is returned by :class:`DeviceAllocator::getID()`. 
+Because we are using global arrays on host and device, the arrays need to be "set up" after at least one DeviceAllocator
+has been created, but before any kernels which use a DeviceAllocator are called. This process is done by calling 
+the ``UMPIRE_SET_UP_DEVICE_ALLOCS()`` macro. This just ensures that the host and device global arrays are updated and 
+pointing at the same memory.
```
mcfadden8, line 57: Question: Can't this all be done within the umpire implementation?  Why can't UMPIRE_SET_UP_DEVICE_ALLOCS() be done by make_device_allocator"

docs/sphinx/features/device_allocators.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. note::
+   In order to use the full capabilities of the DeviceAllocator, relocatable device code must be enabled.
```
mcfadden8, line 60: What is the difference?  What does one get if they do not have relocatable device code?  In what way will the device allocator behave diferently?

docs/sphinx/features/device_allocators.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Resetting Memory on the DeviceAllocator
+---------------------------------------
+
+The memory that has been used with the DeviceAllocator is only freed at the end of a program when the 
+ResourceManager is torn down. However, there is a way to overwrite old or outdated data. Users can call
+the :class:`DeviceAllocator::reset()` method which will allows old data to be overwritten. Below is an example:
```
mcfadden8, line 67: I don't understand this concept.  Is the allocated memory read-only?  Why does it require a reset to write to the memory with new data?  Isn't data read and written from memory and device memory all the time?

docs/sphinx/features/device_allocators.rst
kab163: This will be addressed with https://rzlc.llnl.gov/jira/browse/UM-969

docs/sphinx/features/device_allocators.rst
kab163: This will also be addressed with UM-969

docs/sphinx/features/device_allocators.rst
kab163: If you look at my benchmark (still under construction) 
https://github.com/LLNL/Umpire/blob/2d2e1be032dc303babd602d4d4112d8ae2ee4fae/benchmarks/device_allocator_stress_test.cpp 
you'll see an example of the reset() function. I have one DeviceAllocator with size "total_num_threads * sizeof(double)". So when I use the same DeviceAllocator in each of my kernel calls, that memory space is being filled. If I don't call reset, it will run out of memory. My other option is creating a larger DA, but I don't want to do that since I don't need to keep that data around (overwriting it is totally ok).

PR #664, opened 2021-10-06
Add test to check pools work with DEVICE_CONST memory

davidbeckingsale:


src/umpire/resource/CudaConstantMemoryResource.cu
```diff
@@ -47,7 +47,7 @@ void* CudaConstantMemoryResource::allocate(std::size_t bytes)
 
   void* ret{static_cast<void*>(ptr)};
 
-  if (m_offset > 1024 * 64)
+  if (m_offset > (1024 * 64))
```
mcfadden8, line 50: Are the added parenthesis for clarity or was precendence not being following correctly?

mcfadden8 (dismissed):

src/umpire/resource/CudaConstantMemoryResource.cu
davidbeckingsale: Clarity.

mcfadden8 (approved):

PR #665, opened 2021-10-06
Make ResourceManager::isAllocator work for resources

davidbeckingsale:
MemoryResources like HOST are created lazily, and constructed the first
time they are requested through ResourceManager::getAllocator

The isAllocator function previously only considered created allocators,
this change makes it check the list of (potentially not yet created)
memory resources.


mcfadden8 (dismissed):
Looks great!  Thanks!!

kab163:
I'm scared. Can we just double check the boolean logic lol it confuses me everytime (or I guess the test you've added will do this for me and ensure the boolean logic is right?)

mcfadden8 (approved):

PR #666, opened 2021-10-06
Add ResourceManager::prefetch method

davidbeckingsale:
Add apply_async to the CudaMemPrefetchOperation that takes a
camp::resource to determine which stream to perform the prefetch on.

Expose the "PREFETCH" operations registered with the
MemoryOperationRegistry with a new ResourceManager::prefetch method.
Previously, these operations were only accessible manually, or through
the AllocationPrefetcher strategy.

### DIFF
--- a/src/umpire/ResourceManager.hpp
+++ b/src/umpire/ResourceManager.hpp
@@ -281,6 +281,16 @@ class ResourceManager {
   void deallocate(void* ptr);
 
   /*!
+   * \brief Asynchronously prefetch memory ptr to device.
+   *
+   * \param ptr Pointer to prefech
+   * \param device Device to prefetch data to
+   * \param ctx Resource to use for asynchronous operation
+   */
+  camp::resources::EventProxy<camp::resources::Resource> prefetch(void* ptr, int device,
+                                                                  camp::resources::Resource& ctx);
+
+  /*!
    * \brief Get the size in bytes of the allocation for the given pointer.
    *
    * \param ptr Pointer to find size of.
--- a/src/umpire/op/CudaMemPrefetchOperation.cpp
+++ b/src/umpire/op/CudaMemPrefetchOperation.cpp
@@ -42,5 +42,39 @@ void CudaMemPrefetchOperation::apply(void* src_ptr, util::AllocationRecord* UMPI
   }
 }
 
+camp::resources::EventProxy<camp::resources::Resource> CudaMemPrefetchOperation::apply_async(
+    void* src_ptr, util::AllocationRecord* UMPIRE_UNUSED_ARG(allocation), int value, std::size_t length,
+    camp::resources::Resource& ctx)
+{
+  int device{value};
+  cudaError_t error;
+
+  // Use current device for properties if device is CPU
+  int current_device;
+  cudaGetDevice(&current_device);
+  int gpu = (device != cudaCpuDeviceId) ? device : current_device;
+
+  cudaDeviceProp properties;
+  error = ::cudaGetDeviceProperties(&properties, gpu);
+
+  auto stream = ctx.get<camp::resources::Cuda>().get_stream();
+
+  if (error != cudaSuccess) {
+    UMPIRE_ERROR("cudaGetDeviceProperties( device = " << device << "),"
+                                                      << " failed with error: " << cudaGetErrorString(error));
+  }
+
+  if (properties.managedMemory == 1 && properties.concurrentManagedAccess == 1) {
+    error = ::cudaMemPrefetchAsync(src_ptr, length, device, stream);
+
+    if (error != cudaSuccess) {
+      UMPIRE_ERROR("cudaMemPrefetchAsync( src_ptr = " << src_ptr << ", length = " << length << ", device = " << device
+                                                      << ") failed with error: " << cudaGetErrorString(error));
+    }
+  }
+
+  return camp::resources::EventProxy<camp::resources::Resource>{ctx};
+}
+
 } // end of namespace op
 } // end of namespace umpire
--- a/tests/integration/operation_tests.cpp
+++ b/tests/integration/operation_tests.cpp
@@ -807,3 +807,24 @@ TEST(AsyncTest, Reallocate)
   device_alloc.deallocate(source_array);
 }
 #endif
+
+#if defined(UMPIRE_ENABLE_CUDA)
+TEST(AsyncTest, Prefetch)
+{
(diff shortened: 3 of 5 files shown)


davidbeckingsale:
cc @gzagaris

src/umpire/ResourceManager.hpp
```diff
@@ -280,6 +280,16 @@ class ResourceManager {
...
 
+  /*!
+   * \brief Asynchronously prefetch memory ptr to device.
+   *
+   * \param ptr Pointer to prefech
+   * \param device Device to prefetch data to
+   * \param ctx Resource to use for asynchronous operation
+   */
+  camp::resources::EventProxy<camp::resources::Resource> prefetch(void* ptr, int device,
+                                                                  camp::resources::Resource& ctx);
```
gzagaris, line 291: It looks like the `streamID` is a member of the `ctx` object that is passed as the 3rd argument. 

Would it make sense to provide another version of the `prefetch` method that takes an integer instead (which internally constructs the `camp::resources::Resource` object and then calls this method)?  

That way the caller would not be required to construct a `camp::Resources::Resource` object in order to call `prefetch`, which may simplify its usage in a host application.

src/umpire/ResourceManager.hpp
```diff
@@ -280,6 +280,16 @@ class ResourceManager {
    */
   void deallocate(void* ptr);
 
+  /*!
+   * \brief Asynchronously prefetch memory ptr to device.
+   *
+   * \param ptr Pointer to prefech
```
gzagaris, line 286: Is this going to work with any memory type other than UM? 

If not, it may be worthwhile noting in the doxygen that this operation is only applicable to UM memory and perhaps checking that `ptr` does indeed point to a UM buffer (?)

tests/integration/operation_tests.cpp
```diff
@@ -807,3 +807,24 @@ TEST(AsyncTest, Reallocate)
...
+
+  constexpr std::size_t size = 1024;
+
+  auto alloc = rm.getAllocator("UM");
+  float* array = static_cast<float*>(alloc.allocate(size * sizeof(float)));
+
+  ASSERT_NE(array, nullptr);
+
+  // Prefetch to device 0
+  camp::resources::Event event = rm.prefetch(array, 0, resource);
```
gzagaris, line 825: How would I `prefetch` to the CPU from the device? Would I just pass `cudaCpuDeviceId` as the 2nd argument?

tests/integration/operation_tests.cpp
```diff
@@ -807,3 +807,24 @@ TEST(AsyncTest, Reallocate)
...
+  event.wait();
```
gzagaris, line 826: I can ignore the `event` argument if I am just launching kernels afterwards on the same stream. Correct? (I eventually synchronize  at a later stage using `cudaDeviceSynchronize()`

gzagaris (commented):
Thanks for adding this @davidbeckingsale!

I added some high-level comments/questions for your consideration.

tests/integration/operation_tests.cpp
davidbeckingsale: Yes, and the method actually returns an EventProxy object which means an event won't actually be created if the object is never explicitly stored as an Event, avoiding the overhead.

tests/integration/operation_tests.cpp
davidbeckingsale: Yup.

src/umpire/ResourceManager.hpp
davidbeckingsale: Great suggestions - I will add that.

mcfadden8 (dismissed):

src/umpire/ResourceManager.hpp
davidbeckingsale: We are trying to use Event/Resource everywhere to keep things decoupled from a particular programming model. A Resource can be created from an existing CUDA stream, and I think you should be able to create it inline.

gzagaris: I see. That seems like a reasonable approach in order to be able to future-proof the API.

src/umpire/op/CudaMemPrefetchOperation.cpp
```diff
@@ -42,5 +42,39 @@ void CudaMemPrefetchOperation::apply(void* src_ptr, util::AllocationRecord* UMPI
...
+
+  // Use current device for properties if device is CPU
+  int current_device;
+  cudaGetDevice(&current_device);
+  int gpu = (device != cudaCpuDeviceId) ? device : current_device;
+
+  cudaDeviceProp properties;
+  error = ::cudaGetDeviceProperties(&properties, gpu);
+
+  auto stream = ctx.get<camp::resources::Cuda>().get_stream();
```
gzagaris, line 60: So, at this point the expectation is that Resource was constructed from stream correct? 

- What is the expected behavior if someone calls this with ``Resource ctx{Host()};``?
- Should this be checked with a `try_get` ?

src/umpire/ResourceManager.hpp
```diff
@@ -280,6 +280,16 @@ class ResourceManager {
...
+   * \param device Device to prefetch data to
+   * \param ctx Resource to use for asynchronous operation
```
gzagaris, line 288: - How would I create a `Resource` for the default cuda stream?

davidbeckingsale: Use [CudaFromStream](https://github.com/LLNL/camp/blob/main/include/camp/resource/cuda.hpp#L139) with a stream argument of 0

src/umpire/op/CudaMemPrefetchOperation.cpp
davidbeckingsale: Either case would have to throw - the `try` will throw directly from camp. We have used `get` in all our async operations so far so I won't change this here, but will add a ticket to change this to use `try_get` and throw a more detailed Umpire exception everywhere.

mcfadden8 (approved):

src/umpire/op/CudaMemPrefetchOperation.cpp
gzagaris: Sounds good. Thanks.

src/umpire/ResourceManager.hpp
gzagaris: Thank you for clarifying.

PR #668, opened 2021-10-07
Fixed link to Resources Page in Docs

kab163:
Before, the Allocator Accessibility page had a hard link to the Resources page. When the docs got reorganized, the link broke. To make the link more robust to changes to the docs, I changed it to use the :ref: keyword which references another page in the docs. It should withstand reorg changes.

Note: Tried for a long time to figure out how to use the :doc: keyword instead, but nothing seemed to work right. Will stick to using :ref: since it seems to work well.


davidbeckingsale (changes requested):
Can you please amend the commit message to be imperative, eg. "Fix link to resources on allocator_accessibility page"

mcfadden8 (approved):
Thanks!

davidbeckingsale (approved):

PR #671, opened 2021-10-19
Add GitHub action to generate C/FORTRAN API

davidbeckingsale:


src/umpire/interface/umpire_shroud.yaml
```diff
@@ -26,6 +26,7 @@ cxx_header: umpire/Umpire.hpp
 options:
   F_module_per_class: False
   C_API_case: lower
+  wrap_python: False
```
mcfadden8, line 29: What does this do?  Is it required by github actions in some way?

davidbeckingsale: Nope - it controls whether or not Shroud generates a Python wrapper. I was trying to avoid having to create the empty `python` directory (it didn't help though).

I left it in since it's technically correct and somewhat clearer than leaving it out.

mcfadden8 (approved):
Thanks @davidbeckingsale

PR #672, opened 2021-10-19
Add replay --introspection-off option

mcfadden8:
This option will allow for replays to be run with introspection
turned off for all allocators being replayed.


davidbeckingsale (commented):
Do we also want to create the resources with no introspection?

davidbeckingsale (approved):

PR #676, opened 2021-11-03
Fix return value from get_process_highwater_mark

mcfadden8:
This function obtains the high watermark from the VmHWM line in
/proc/PID/status.  The value associated with this field is expressed in
kB.  The previous implementation treated this value as the number of
pages causing the return value to be off by a factor of PAGESIZE/1024.

This fix corrects this by treating VmHWM as kB instead.

### DIFF
--- a/src/umpire/Umpire.cpp
+++ b/src/umpire/Umpire.cpp
@@ -136,9 +136,12 @@ std::size_t get_process_memory_usage_hwm()
 
     if (key == "VmHWM:") {
       std::size_t resident_hwm;
-      long page_size{::sysconf(_SC_PAGE_SIZE)};
       ss >> resident_hwm;
-      rval = std::size_t{resident_hwm * page_size};
+
+      //
+      // "VmHWM" returns the number of kB in use. Convert this to number of bytes
+      //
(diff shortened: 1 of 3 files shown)


src/umpire/Umpire.cpp
```diff
@@ -136,9 +136,8 @@ std::size_t get_process_memory_usage_hwm()
 
     if (key == "VmHWM:") {
       std::size_t resident_hwm;
-      long page_size{::sysconf(_SC_PAGE_SIZE)};
       ss >> resident_hwm;
-      rval = std::size_t{resident_hwm * page_size};
+      rval = std::size_t{resident_hwm * 1024};
```
davidbeckingsale, line 144: Can you add a comment here about the magical 1024?

davidbeckingsale (changes requested):

src/umpire/Umpire.cpp
mcfadden8: @davidbeckingsale, this is done (I need your review again)

davidbeckingsale (dismissed):

mcfadden8:
LGTM

davidbeckingsale (approved):

PR #680, opened 2021-11-10
Fix host and device Allocator ID overlap

kab163:
Modify the way DeviceAllocator IDs are created so they don't conflict with host allocator IDs.
Create test suite to double check functionality of DeviceAllocator and helper functions.
Add error/logging messages and documentation for helper functions, etc.


benchmarks/device_allocator_stress_test.cpp
```diff
@@ -95,9 +95,8 @@ int main(int, char**)
                  "Timing won't be as accurate. Continuing anyways... **" << std::endl;
   }
 
-  // Create device allocator and set up memory
+  // Create device allocator
   auto dev_alloc = umpire::make_device_allocator(allocator, N * sizeof(double), "dev_alloc");
-  UMPIRE_SET_UP_DEVICE_ALLOCATORS();
```
davidbeckingsale, line 100: This is still required (in case this is a different translation unit)

docs/sphinx/features/device_allocators.rst
```diff
@@ -49,15 +50,11 @@ With the :class:`umpire::get_device_allocator` function, there is no need to kee
...
-Under the hood, the :class:`umpire::get_device_allocator` uses global arrays which can be accessed by both 
-the host and device. The global array is indexed by DeviceAllocator ID, which is returned by :class:`DeviceAllocator::getID()`. 
-Because we are using global arrays on host and device, the arrays need to be "set up" after at least one DeviceAllocator
-has been created, but before any kernels which use a DeviceAllocator are called. This process is done by calling 
-the ``UMPIRE_SET_UP_DEVICE_ALLOCS()`` macro. This just ensures that the host and device global arrays are updated and 
-pointing at the same memory.
-
 .. note::
-   In order to use the full capabilities of the DeviceAllocator, relocatable device code must be enabled.
+   In order to use the full capabilities of the DeviceAllocator, relocatable device code (RDC) must be enabled. If RDC is
```
davidbeckingsale: 👍

src/umpire/device_allocator_helper.hpp
```diff
@@ -62,11 +62,13 @@ __host__ void destroy_device_allocator();
...
 #define UMPIRE_SET_UP_DEVICE_ALLOCATORS()                                                                          \
   {                                                                                                                \
+    UMPIRE_LOG(Debug, "Calling cudaMemcpyToSymbol macro.");                                                        \
     cudaMemcpyToSymbol(umpire::UMPIRE_DEV_ALLOCS, &umpire::UMPIRE_DEV_ALLOCS_h, sizeof(umpire::DeviceAllocator*)); \
   }
 #elif defined(UMPIRE_ENABLE_HIP)
 #define UMPIRE_SET_UP_DEVICE_ALLOCATORS()                                                                         \
   {                                                                                                               \
+    UMPIRE_LOG(Debug, "Calling hipMemcpyToSymbol macro.");                                                        \
     hipMemcpyToSymbol(umpire::UMPIRE_DEV_ALLOCS, &umpire::UMPIRE_DEV_ALLOCS_h, sizeof(umpire::DeviceAllocator*)); \
```
kab163: check for null ptr (see how desul does this) - don't want to have too many duplicate copy calls

kab163:
Things to check:
- [x] Is my strategy for how to save the allocator IDs within the DA class ok? **No! Fix this next...**
- [x] Should the way the benchmark vs the example handles the macro be changed to be more consistent? **Yes. Call the macro manually in both the example and the benchmark.**
- [x] Should the name of the macro be changed? **No, leaving it alone is fine.**
- [x] Should we make RDC enablement required for this to work properly? Does the documentation need to change? **I don't think we need to make RDC enablement required, but there will be documentation about what happens when RDC is there and when it's not. Most of our LLNL users (all?) use RDC anyways.**

kab163:
Device Allocator IDs...
- From user's perspective, DAs will start at 1 and increment upwards with each new DA created (64 being max).
- From DA class perspective, DA ids will be negative, starting at -1 (and -64 is the "lowest")
- When indexing into the global array... (1) a user given ID will be decremented by 1 and then used... or (2) a value from DeviceAllocator's m_id will be multiplied by (-1) and then decremented by 1

docs/sphinx/features/device_allocators.rst
```diff
@@ -28,17 +28,19 @@ exceed the size that was used when making the device allocator.
...
-obtain the DeviceAllocator object:
+obtain the first DeviceAllocator object:
 
 .. literalinclude:: ../../../examples/device-allocator.cpp
    :start-after: _sphinx_tag_get_dev_allocator_id_start
    :end-before: _sphinx_tag_get_dev_allocator_id_end
    :language: C++
 
-And next is an example of using the **name** instead:
+Note that the IDs start at one. The range of valid DeviceAllocator IDs is between 1 and 64, inclusively.
```
mcfadden8: I do not think that this should be exposed to the user.  They should always use is_device_allocator to determine whether it is a device allocator or not.  ID value semantics should not be exposed so that applications aren't tempted to make use of them.

kab163: Even the fact that they start at one and not zero?

docs/sphinx/features/device_allocators.rst
```diff
@@ -28,17 +28,18 @@ exceed the size that was used when making the device allocator.
 Retrieving a DeviceAllocator Object
 -----------------------------------
 
-The :class:`umpire::get_device_allocator` host/device function returns the DeviceAllocator object that corresponds 
+After creating a DeviceAllocator, we can immediately start using that allocator to allocate device memory. To do this, we
+have the :class:`umpire::get_device_allocator` host/device function which returns the DeviceAllocator object corresponding
```
mcfadden8, line 32: I think that this example should be replaced with just an example that uses the name as this will likely be the most common case.  If an example with IDs is necessary, the ID should be passed in (from somewhere) as a parameter and not be a integer literal.

examples/device-allocator.cpp
```diff
@@ -56,17 +56,15 @@ int main(int argc, char const* argv[])
   auto device_allocator = umpire::make_device_allocator(allocator, sizeof(double), "my_device_alloc");
   // _sphinx_tag_make_dev_allocator_end
 
-  // Checking that a DeviceAllocator exists...
-  if (umpire::is_device_allocator(0)) {
+  // Checking that the DeviceAllocator just created can be found...
+  if (umpire::is_device_allocator(1)) {
```
mcfadden8: Please replace this with a call to get the allocator by name instead of an integer literal.

Also, the integer literal above needs to be negative.

src/umpire/DeviceAllocator.cpp
```diff
@@ -79,9 +79,9 @@ __device__ void* DeviceAllocator::allocate(size_t size)
   return static_cast<void*>(m_ptr + counter);
 }
 
-__host__ __device__ size_t DeviceAllocator::getID()
+__host__ __device__ int DeviceAllocator::getID()
 {
-  return m_id;
+  return (m_id * (-1));
```
mcfadden8: May I presume that `m_id` starts at 1?  If it starts at 0, this will be problematic.

src/umpire/device_allocator_helper.cpp
```diff
@@ -64,13 +65,15 @@ __host__ __device__ DeviceAllocator get_device_allocator(const char* name)
 
 __host__ __device__ DeviceAllocator get_device_allocator(int id)
 {
-  if (id < 0 || id > UMPIRE_TOTAL_DEV_ALLOCS) {
+  if (id < 1 || id > UMPIRE_TOTAL_DEV_ALLOCS) {
```
mcfadden8: This does not seem to be correct.  The user should be given a negative number as an ID.  So, when they call get_device_allocator() it will be a negative number that they pass in right?

src/umpire/device_allocator_helper.cpp
```diff
@@ -96,17 +106,25 @@ __host__ __device__ bool is_device_allocator(int id)
 //////////////////////////////////////////////////////////////////////////
 __host__ DeviceAllocator make_device_allocator(Allocator allocator, size_t size, const std::string& name)
 {
-  static size_t allocator_id{0};
-  auto dev_alloc = DeviceAllocator(allocator, size, name, allocator_id);
+  static int allocator_id{0};
+  int da_id = ((-1) * (allocator_id)) - 1;
+
+  // The DA ID should not conflict with other allocator IDs,
```
mcfadden8: Please change "DA" to "Device Allocator" or "DeviceAllocator" in documentation and comments.

mcfadden8 (changes requested):
IIRC, the interface discussed was to have Umpire return an ID to the user that is negative so that we may avoid overlapping with IDs created for allocators managed by the Host.  I see some inconsistencies in get_device_allocator's implementation which appears to be assuming that the integer it is receiving is positive.  I think it needs to assume it is negative right?

docs/sphinx/features/device_allocators.rst
davidbeckingsale: Agree, name is best here.

docs/sphinx/features/device_allocators.rst
davidbeckingsale: I think this note can be simplified, to just something like - "When compiling without relocatable device code, the UMPIRE_SET_UP_DEVICE_ALLOCATORS macro must be called in every translation unit that will use the get_device_allocator function"

src/umpire/DeviceAllocator.cpp
davidbeckingsale: This should return m_id, the conversion from the negative id to an array index should only happen internally to the make_device_allocator/get_device_allocator functions.

src/umpire/ResourceManager.cpp
```diff
@@ -220,6 +220,10 @@ Allocator ResourceManager::getAllocator(int id)
 {
   UMPIRE_LOG(Debug, "(\"" << id << "\")");
 
+  if (id < 0) {
+    UMPIRE_ERROR("Passed an invalid ID - is this a DeviceAllocator instead?");
```
davidbeckingsale: This error message should include the id.

src/umpire/device_allocator_helper.cpp
davidbeckingsale: The id needs to be converted here, at the very top of the function.

src/umpire/device_allocator_helper.cpp
```diff
@@ -96,17 +106,25 @@ __host__ __device__ bool is_device_allocator(int id)
 //////////////////////////////////////////////////////////////////////////
 __host__ DeviceAllocator make_device_allocator(Allocator allocator, size_t size, const std::string& name)
 {
-  static size_t allocator_id{0};
-  auto dev_alloc = DeviceAllocator(allocator, size, name, allocator_id);
+  static int allocator_id{0};
```
davidbeckingsale: Should be -1, and the conversion should only be used to index the array.

davidbeckingsale (changes requested):
Main thing is that the IDs exposed to the user should always be the actual value, and conversion from the negative ID to an array index should happen internally. I would recommend adding an inline function to do this.

src/umpire/DeviceAllocator.cpp
mcfadden8: And m_id should start at -1

mcfadden8:
> Device Allocator IDs...
> 
> * From user's perspective, DAs will start at 1 and increment upwards with each new DA created (64 being max).
> * From DA class perspective, DA ids will be negative, starting at -1 (and -64 is the "lowest")

Just the opposite right?  From the user's perspective, they should see -1 to -64

kab163:
Ok I think I've applied all the changes we needed. One question though: Should I overload the "is_device_allocator" function so that it could also be given a name? Or just change it from checking an ID to instead checking a name?

davidbeckingsale:
> Ok I think I've applied all the changes we needed. One question though: Should I overload the "is_device_allocator" function so that it could also be given a name? Or just change it from checking an ID to instead checking a name?

Good idea - I think `is_device_allocator` should work with both a name and an id.

mcfadden8:
Do we CI unit tests for these?  I think those would be good if possible.

kab163:
> Do we CI unit tests for these? I think those would be good if possible.

I have a separate ticket for the DeviceAllocator tests
[UPDATE: Edited that ticket so it has a "has to be finished together with" link to this PR's ticket.]

kab163:
DA Tests questions:
- [x] Should I check to see if a DeviceAllocator can be created with size 0? I think this could technically be done, but it would be useless.
- [x] Should resetting the UMPIRE_DEV_ALLOCS array back to nullptr be added to the "destroy_device_allocator" function?
- [x] Could the structure of DeviceAllocator tests be improved? Other tests added?

davidbeckingsale:
> DA Tests questions:
> 
>     * [ ]  Should I check to see if a DeviceAllocator can be created with size 0? I think this could technically be done, but it would be useless.
> 
>     * [ ]  Should resetting the UMPIRE_DEV_ALLOCS array back to nullptr be added to the "destroy_device_allocator" function?
> 
>     * [ ]  Could the structure of DeviceAllocator tests be improved? Other tests added?

I think the size 0 allocator should throw an error.
Yup, resetting to a nullptr is a good idea.
Will check the tests...

src/umpire/device_allocator_helper.cpp
davidbeckingsale: Minor nitpick - this is really the index, and should be named as such.

davidbeckingsale (commented):
We should make the DeviceAllocator constructor private and mark the `make_device_allocator` method a friend. This will stop users from creating allocators that will be "untracked".

mcfadden8 (dismissed):
Nice work

mcfadden8 (approved):

davidbeckingsale (approved):

PR #681, opened 2021-11-12
Add cmake install command for DeviceAllocator headers

kab163:
When users "make install", the DeviceAllocator headers will now be included in the install dir, if enabled

### DIFF
--- a/src/umpire/CMakeLists.txt
+++ b/src/umpire/CMakeLists.txt
@@ -66,6 +66,10 @@ if (UMPIRE_ENABLE_CUDA)
     $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
     $<INSTALL_INTERFACE:include>)
 
+  install(FILES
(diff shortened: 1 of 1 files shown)


src/umpire/CMakeLists.txt
```diff
@@ -170,6 +170,7 @@ install(TARGETS
 
 install(FILES
   ${umpire_headers}
+  ${umpire_device_headers}
```
davidbeckingsale: There should be a separate `install` command with the rest of the `umpire_device` library stuff (inside the `if (ENABLE_CUDA)` block)

davidbeckingsale (changes requested):

src/umpire/CMakeLists.txt
kab163: Oh duh, yea you're right

davidbeckingsale (approved):

PR #682, opened 2021-11-23
Fix shroud workflow

mcfadden8:


mcfadden8:
@davidbeckingsale and @kab163, I tested these changes as follows:

1. I modified `umpire_shroud.yaml` and confirmed that the workflow ran.
2. I cherry-picked this change into my `event` branch which did not have a modified `umpire_shroud.yaml` and saw that the workflow was indeed not triggered.
3. I reverted my modification done in step one, rebased and squashed and confirmed that the workflow is no longer triggered here since this pull request no longer includes a changed `umpire_shroud.yaml` file.

This change was inpsired by: https://github.community/t/is-it-possible-to-run-the-job-only-when-a-specific-file-changes/115484

davidbeckingsale (approved):

PR #686, opened 2021-12-21
Fix comparison operators for TypedAllocators

MrBurmark:
Move the operators into the same namespace as TypedAllocator for ADL instead of leaving them in the global namespace.


davidbeckingsale (approved):

mcfadden8 (approved):

MrBurmark:
I can't press merge, but this is ready to go.

PR #692, opened 2022-01-18
Use blt namespace for hip targets

mcfadden8:
The cmake configuration system of Umpire was changed to use `blt::blt_runtime` and `blt::blt` for HOST runtime and DEVICE code respectively.  Additionally, the docker CI instance for HIP was updated to use ROCM 4.3.1.

Finally, for the time being, the CI for hip was modified to stop treating compiler warnings as errors since the ROCM 4.3.1 cmake configuration file is adding a command line option to the compiler which is not supported yet causing the compiler to generate a warning that it is ignoring the option.


examples/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ if (UMPIRE_ENABLE_HIP)
   blt_add_executable(
     NAME multi_device
     SOURCES multi_device.cpp
-    DEPENDS_ON umpire hip)
+    DEPENDS_ON umpire blt::hip)
```
white238, line 36: You shouldn't have to list `blt::hip` and `blt::hip_runtime` everywhere that depends on `umpire` because it will be inherited properly.

This is probably a hold-over from the old BLT register libraries that did not get inserted into the CMake DAG.

mcfadden8: So you would expect for ALL DEPEND_ON hip and hip_runtime to go away in Umpire?

white238: Not all but probably most if not all but one.

src/umpire/CMakeLists.txt
```diff
@@ -84,13 +84,22 @@ set (umpire_depends
   umpire_util
   umpire_tpl_judy)
 
+if (UMPIRE_ENABLE_HIP)
+  set (umpire_depends
+    ${umpire_depends}
+    blt::hip)
```
davidbeckingsale:
```suggestion
    blt::hip_runtime)
```

src/umpire/CMakeLists.txt
```diff
@@ -84,13 +84,22 @@ set (umpire_depends
...
+endif()
+
 if (UMPIRE_ENABLE_OPENMP_TARGET)
   set (umpire_depends
     ${umpire_depends}
     openmp)
 endif ()
 
 blt_list_append( TO umpire_depends ELEMENTS umpire_interface IF UMPIRE_ENABLE_C )
+blt_list_append( TO umpire_depends ELEMENTS umpire_device IF UMPIRE_ENABLE_CUDA)
```
davidbeckingsale: These two lines should be removed.

src/umpire/interface/c_fortran/CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ set_source_files_properties(
 
 set(umpire_interface_c_fortran_depends camp)
 blt_list_append( TO umpire_interface_c_fortran_depends ELEMENTS cuda_runtime IF UMPIRE_ENABLE_CUDA )
-blt_list_append( TO umpire_interface_c_fortran_depends ELEMENTS hip_runtime IF UMPIRE_ENABLE_HIP )
+blt_list_append( TO umpire_interface_c_fortran_depends ELEMENTS blt_hip_runtime IF UMPIRE_ENABLE_HIP )
```
davidbeckingsale:
```suggestion
blt_list_append( TO umpire_interface_c_fortran_depends ELEMENTS blt::hip_runtime IF UMPIRE_ENABLE_HIP )
```

src/umpire/CMakeLists.txt
mcfadden8: @davidbeckingsale Do you mean lines 99 and 100 or 100 and 101?

PR #700, opened 2022-01-26
Adding Third-party UMap dependency

asarkar-parsys:
Creates a UMPIRE_ENABLE_UMAP cmake build flag which is off by default
Integrates File Resource to use UMap APIs
Resolved UMPIRE unit test issues with UMap: 
UMap requires a valid file descriptor to access file after mapping and so added UMap related changes to retain the fd until the resource is released unmap gets called. 
Also, UMap does not support MAP_SHARED
For zero-byte file test, UMap changes are in zero_file_map branch (https://github.com/LLNL/umap/tree/zero_file_map)


mcfadden8:
Thanks @asarkar-parsys, if I understand correctly, `umap` has been integrated with Umpire in such a way that it modifies the File Resource to internally use `umap` instead of `mmap`.  Is the the correct intention?  What are the anticipated performance and usage differences on the File Resource when it has been compiled in `umap` mode instead of `mmap`?  Also, what are the reasons to use `umap` instead of `mmap` within the File Resource implementation?

After our discussion yesterday afternoon, I now understand that `umap` may be configured with environment variables and I agree that this is an acceptable level of integration at this point as it will provide what is needed for further use case exploration.

mcfadden8:
It appears that there have been changes in the `blt` and `radiuss-spack-configs` packages?  Are these intentional?

cmake/SetupUmpireThirdParty.cmake
```diff
@@ -15,6 +15,24 @@ if (EXISTS ${SHROUD_EXECUTABLE})
   include(${CMAKE_CURRENT_BINARY_DIR}/SetupShroud.cmake)
 endif ()
 
+if (UMPIRE_ENABLE_UMAP)
+  if(NOT DEFINED ENV{UMAP_ROOT})
```
mcfadden8: I'm not sure that environment variables are the best way to do this.  @davidbeckingsale, to you have any thoughts on this?

asarkar-parsys:
> It appears that there have been changes in the `blt` and `radiuss-spack-configs` packages? Are these intentional?

Rebased submodule changes

asarkar-parsys:
force pushed new changes

asarkar-parsys:
> It appears that there have been changes in the `blt` and `radiuss-spack-configs` packages? Are these intentional?

cmake/SetupUmpireThirdParty.cmake
davidbeckingsale: I think it is okay to use both an ENV variable, and a regular CMake variable for UMAP_DIR. However, it would be best to still try the find_path, and only issue an error if the library cannot be found (rather than erroring just because a variable is undefined).

davidbeckingsale:
Merged in #711

PR #704, opened 2022-02-07
Add profile compare benchmark

mcfadden8:
An existing benchmark was extended to add metrics for time in addition
to space.  A script was also added to build and create an hatchet
profile for the current branch and the develop branch for comparison on
HPC deployments.

### DIFF
--- a/scripts/benchmark/profile_million_allocations
+++ b/scripts/benchmark/profile_million_allocations
@@ -1,0 +1,90 @@
+#!/bin/bash
+#
+# This is intended to be run from within the scripts/benchmark directory of the umpire source.
+# It will take a profile of the develop branch and compare it to a problem of this branch (using hatchet and hpctoolkit)
+#
+
+#
+# Treat undefined variable usage as an error
+#
+set -u
+
+rootdir=`pwd`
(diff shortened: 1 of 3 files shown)


scripts/benchmark/profile_million_allocations
```diff
@@ -0,0 +1,90 @@
...
+  if ! eval $1
+  then
+    echo "FAIL"
+    echo $1
+    echo "FAIL"
+    exit 1
+  fi
+}
+
+# module -q load cmake git gcc/10.2.1 hpctoolkit/2021.05
```
davidbeckingsale: Can we make the modules SYS_TYPE dependent or anything?

mcfadden8: Sure.  Maybe v2?  What we have here will work for TOSS3, TOSS4 (I think), and BLUE systems.

davidbeckingsale: Okay, then this looks good, but please remove all the commented lines that look like they were from testing stuff out.

davidbeckingsale (approved):

PR #708, opened 2022-02-24
Resolve cuda device symbols in CMakeLists.txt

td-mpcdf:
Linking against the Umpire library led to unresolved symbols, which can be avoided by this PR.
When using the Fortran interface with `use umpire_mod` I had to add the `include/umpire` path to the exported include paths of the package.


davidbeckingsale:
Umpire requires a device link unless you build with UMPIRE_ENABLE_DEVICE_ALLOCATOR=Off 

Resolving the device symbols this way "works", but for any user that tries to leverage the device-side functionality.

PR #713, opened 2022-03-14
Add umpire::coalesce method that acceps all Allocators

davidbeckingsale:
umpire::coalesce will throw an exception if the provided allocator doesn't support coalesce. Added umpire::try_coalesce method that returns a boolean indicating if coalescing could be performed.


src/umpire/Umpire.cpp
```diff
@@ -253,4 +255,32 @@ MPI_Comm get_communicator_for_allocator(Allocator a, MPI_Comm comm)
...
+  auto s = a.getAllocationStrategy();
+  bool coalesced{false};
+
+  strategy::QuickPool* qp{dynamic_cast<strategy::QuickPool*>(s)};
+  if (qp) {
+    qp->coalesce();
+    coalesced = true;
+  }
+
+  strategy::DynamicPoolList* dpl{dynamic_cast<strategy::DynamicPoolList*>(s)};
```
mcfadden8, line 269: Should this be placed within an `else {}` block for the case when `qp` is a `nullptr`?

mcfadden8 (approved):
Looks good! You can take or leave my minor comment.

PR #715, opened 2022-03-14
Bugfix/um 995 update corona ci jobs

kab163:
Adding back in corona CI jobs, testing latest hip versions


.gitlab/corona-jobs.yml
```diff
@@ -4,12 +4,12 @@
 #
 # SPDX-License-Identifier: (MIT)
 ##############################################################################
-rocm_4_1_fortran_gcc_8_1_0 (build and test on corona):
+rocm_4_5_2_fortran_clang_13_0_0 (build and test on corona):
```
mcfadden8, line 7: Why did we switch from gcc_8_1_0 to clang_13_0_0?

examples/cookbook/CMakeLists.txt
```diff
@@ -80,10 +80,17 @@ if (UMPIRE_ENABLE_CUDA)
...
+    if (UMPIRE_ENABLE_HIP)
+      set_source_files_properties(
+        recipe_pinned_pool.F
+        PROPERTIES
+        COMPILE_FLAGS -Mfree)
+    else()
+      set_source_files_properties(
+        recipe_pinned_pool.F
+        PROPERTIES
+        Fortran_FORMAT FREE)
```
mcfadden8, line 92: Interesting.  I wonder why this broke?  Wasn't this previously working with HIP?  (just curious)

scripts/spack_packages/umpire/package.py
```diff
@@ -189,20 +189,26 @@ def initconfig_hardware_entries(self):
...
+            #entries.append(cmake_cache_path("HIP_CLANG_PATH",
+            #                            rocm_root + '/llvm/bin'))
             entries.append(cmake_cache_string("HIP_HIPCC_FLAGS",
                                         '--amdgpu-target=gfx906'))
-            entries.append(cmake_cache_string("HIP_RUNTIME_INCLUDE_DIRS",
-                                        "{0}/include;{0}/../hsa/include".format(hip_root)))
-            hip_link_flags = "-Wl,--disable-new-dtags -L{0}/lib -L{0}/../lib64 -L{0}/../lib -Wl,-rpath,{0}/lib:{0}/../lib:{0}/../lib64 -lamdhip64 -lhsakmt -lhsa-runtime64".format(hip_root)
+            #entries.append(cmake_cache_string("HIP_RUNTIME_INCLUDE_DIRS",
+            #                            "{0}/include;{0}/../hsa/include".format(hip_root)))
+            #hip_link_flags = "-Wl,--disable-new-dtags -L{0}/lib -L{0}/../lib64 -L{0}/../lib -Wl,-rpath,{0}/lib:{0}/../lib:{0}/../lib64 -lamdhip64 -lhsakmt -lhsa-runtime64".format(hip_root)
```
mcfadden8: Is there a reason to keep these commented lines?  I suggest we remove them or add another comment explaining why they are being left here.

scripts/spack_packages/umpire/package.py
```diff
@@ -189,20 +189,26 @@ def initconfig_hardware_entries(self):
...
+            entries.append(cmake_cache_option("UMPIRE_ENABLE_TOOLS", False))
+            hip_link_flags = ""
             if '%gcc' in spec:
                 gcc_bin = os.path.dirname(self.compiler.cxx)
                 gcc_prefix = join_path(gcc_bin, '..')
                 entries.append(cmake_cache_string("HIP_CLANG_FLAGS", "--gcc-toolchain={0}".format(gcc_prefix))) 
                 entries.append(cmake_cache_string("CMAKE_EXE_LINKER_FLAGS", hip_link_flags + " -Wl,-rpath {}/lib64".format(gcc_prefix)))
             else:
-                entries.append(cmake_cache_string("CMAKE_EXE_LINKER_FLAGS", hip_link_flags))
+                entries.append(cmake_cache_string("CMAKE_EXE_LINKER_FLAGS", "-Wl,-rpath={0}/llvm/lib/".format(rocm_root)))
```
mcfadden8, line 206: Should we await feedback from LC before making this particular change?

examples/cookbook/CMakeLists.txt
kab163: I'm not sure. Until I added the -Mfree flag, I got errors (https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/586524), but then I realized that the -Mfree flag would not work for some compilers. So then I made the different flag depending on if hip was enabled or not and that seemed to do the trick.

scripts/spack_packages/umpire/package.py
kab163: Ah yes. I think I planned to remove them once everything started working again. I think it should be removed (we can always look back to previous Umpire versions to reference what the flags used to be if needed, but I don't think it should be)

scripts/spack_packages/umpire/package.py
kab163: Indeed we should!

kab163: Documenting what we talked about in dev meeting: We will keep this change for now and if LC develops a better fix later on, then we'll go from there.

examples/cookbook/CMakeLists.txt
kab163: Used Fortran_COMPILER_FAMILY_IS_CLANG to make this flag more exact to amdclang... let's see if it works

.gitlab/corona-jobs.yml
kab163: Documenting what we talked about in dev meeting: GCC on corona only works for hip_runtime, not hip, so for now we have to use amdclang (which is clang 13) and hopefully at some point that will extend to cce

mcfadden8 (approved):
Looks good to me.  Thanks!

kab163:
John G. commented on ticket from LC - he will double check whether or not the bug has been reported or not and ping the right people, but for rocm 4.5.2. it won't be fixed. For now at least, it seems we need the `-rpath` flag, with the hope that for some upcoming rocm release, it will no longer be needed.

davidbeckingsale (approved):

PR #719, opened 2022-03-21
Add DeviceAllocator CurrentSize/TotalSize functions

kab163:
Adding `getCurrentSize` and `getTotalSize` functions to DeviceAllocator for better monitoring of DeviceAllocator size. Also adding more to the DeviceAllocator test to check that the `reset` function works.


src/umpire/DeviceAllocator.hpp
```diff
@@ -64,6 +64,9 @@ class DeviceAllocator {
    */
   __host__ __device__ bool isInitialized();
 
+  __host__ __device__ unsigned int getCurrentSize();
```
mcfadden8, line 73: Please add doxygen comment here similar to the rest of the header file

src/umpire/DeviceAllocator.hpp
```diff
@@ -64,6 +64,9 @@ class DeviceAllocator {
...
+  __host__ __device__ size_t getTotalSize();
```
mcfadden8, line 81: Ditto (Doxygen comment needed)

mcfadden8 (approved):

PR #720, opened 2022-03-28
Fix find_first_set() on Windows

degrbg:
I encountered a bug when running the tutorials on dynamic pools (`tut_dynamic_pool_1` and `tut_dynamic_pool_2`). Both tutorials would crash on my machine (Windows 10, Visual Studio 2019) after encountering a stack overflow in `FixedSizePool<T, MA, IA, NP>::allocate()`.

I traced the bug to where `allocate()` calls `allocInPool()`. The "search" (`for` loop) in `allocInPool()` would always fail, and a null pointer was returned. `allocate()` would then continue, passing the call to `newPool()`, and then recursing into `allocate()`.

The "search" in `allocInPool()` was always failing because the Windows implementation of `find_first_set()` was incorrect. This appeared in two places, so I refactored the code (and added a unit test) before I fixed it.

### DIFF
--- a/src/umpire/strategy/FixedPool.cpp
+++ b/src/umpire/strategy/FixedPool.cpp
@@ -12,8 +12,8 @@
 #include <cstring>
 #include <sstream>
 
-#include "umpire/util/find_first_set.hpp"
 #include "umpire/util/Macros.hpp"
+#include "umpire/util/find_first_set.hpp"
 
 #if !defined(_MSC_VER)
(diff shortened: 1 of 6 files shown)


davidbeckingsale:
Thanks @degrbg! I have opened a "local" PR so that we can run our internal testing. See #722

degrbg:
Thank you @davidbeckingsale!

degrbg:
Reopening; I'll close this when #722 is merged.

PR #721, opened 2022-04-01
Adding HIP/CUDA jobs in CI

kab163:
For a while, we have made hip+openmp jobs a conflict within the spack. I don't think that restriction holds anymore. This ticket will add a hip+openmp job, but keep the conflict for hip+openmp_target.

Along the same lines, we have disabled cuda+shared jobs because there was a restriction at one point between cuda jobs and building with shared libs. That no longer applies, so we will also add '+shared' to an existing cuda job to test.


adrienbernede:
@kab163 I am currently working on revamp the CI. The change here is pretty minimal, so I don’t think I need to worry too much about having this merged in before I start.

.gitlab/corona-jobs.yml
```diff
@@ -11,5 +11,5 @@ rocm_4_5_2_fortran_clang_13_0_0 (build and test on corona):
 
 rocm_4_5_2_fortran_clang_13_0_0(build and test on corona):
```
adrienbernede: I think it will be wise to rename one of those job so that they are easier to differentiate.

kab163: I was thinking the same thing!

davidbeckingsale (approved):
👍

mcfadden8 (approved):

adrienbernede:
Oops, so this was merged. I’ll need to adapt #723

adrienbernede:
Done.

PR #730, opened 2022-04-11
Re-enable corona and use pdebug queue

mcfadden8:


davidbeckingsale (approved):

kab163:
Yea, exact same failures on same tests. There is some larger bug happening here, but I can't figure it out. Marty, have you tried running this locally? When I was testing this, it would work/succeed locally, but still fail on Gitlab. :(

Update: RAJA's CI just passed and they are using pdebug. I just restarted those tests to see if it passes now... and they still failed lol

PR #732, opened 2022-04-13
Remove blt_print_target_properties from umpire_tpl_fmt

davidbeckingsale:


kab163:
I restarted some of the Ruby jobs that seemed to time out or be some kind of error as the fault of Ruby.. hopefully those pass, though Corona jobs are as good as it gets right now..

kab163 (approved):
Only tests that fail the CI are the known Corona ones!

PR #733, opened 2022-04-13
Add option UMPIRE_DISABLE_ALLOCATIONMAP_DEBUG

### LINKED ISSUE #718: Better guard debug output in AllocationMap::find()

In debug, I see a lot of verbose output from Umpire due to the following:
https://github.com/LLNL/Umpire/blob/8f61165c0bdcf7b3a09cad69b2a938120a54d94e/src/umpire/util/AllocationMap.cpp#L192-L195

This happens b/c I do have buffers that may be allocated on the stack, but, the code may still try to find an umpire allocator in the process of determining whether the memory is on the host or device. 

The output can be quite verbose and it is preferable if it can be disabled on demand for debug builds as well (on Release builds it goes away). 

Perhaps, one way to do this would be to provide an umpire cmake option, e.g., `UMPIRE_DISABLE_ALLOCATION_MAP_NO_DEBUG` or something like that.

I tried setting `target_source_properties` and `blt_target_definitions` to define `NDEBUG`, but, that didn't seem to work...

davidbeckingsale:
Closes #718


davidbeckingsale:
CC @gzagaris

mcfadden8 (dismissed):

mcfadden8 (approved):

gzagaris:
Thanks, @davidbeckingsale -- I'll check this out after this gets merged. Do you plan on making a patch release any time soon?

davidbeckingsale:
> Thanks, @davidbeckingsale -- I'll check this out after this gets merged. Do you plan on making a patch release any time soon?

We won't include this as a patch, but we are going to start releasing RAJA/CHAI/Umpire on a more regular schedule! So shouldn't be too long until next release.

kab163 (commented):
My only concern here is the sort of double negative this is using. If you turn OFF a "disable" then it's enabled.. so by default it is enabled. If you turn this ON then it is disabled.. it's just a funky wording. Do we want it this way?

PR #735, opened 2022-04-14
Adding umpire_util dependency to umpire_device lib

kab163:
In order to allow error handling for the DeviceAllocator, we need to add the umpire_util dependency to the umpire_device lib.  There were some other additional edits that were made so that error handling in the DeviceAllocator files were correct. We also added "+device_alloc" to a lassen CI job so that the DeviceAllocator is tested in the CI and these issues can be found sooner.


davidbeckingsale (changes requested):
Can you add a GitLab job that tests this? (Or update an existing job)

kab163:
> Can you add a GitLab job that tests this? (Or update an existing job)

Good idea. Let's hope I did that right! Lol
Toda!! https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/622319

davidbeckingsale (approved):

src/umpire/CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ if (UMPIRE_ENABLE_CUDA AND UMPIRE_ENABLE_DEVICE_ALLOCATOR)
                               PROPERTIES
                               LANGUAGE CUDA)
 
-  set(umpire_device_depends camp cuda)
+  set(umpire_device_depends camp cuda umpire_util)
```
mcfadden8, line 52: Why `umpire_util`?  The helper seems to be calling `umpire::fmt::format`, not util.  Shouldn't this be:
```suggestion
  set(umpire_device_depends camp cuda umpire_tpl_fmt)
```

kab163: I just tested that and it works with that as well. @davidbeckingsale should I change it to `umpire_tpl_fmt` instead?

davidbeckingsale: The error classes are provided under umpire_util so I think that is what should be used. The fmt dependency will get brought in transitively.

PR #738, opened 2022-04-20
Refactor CUDA operations

davidbeckingsale:


src/umpire/op/CudaAdviseOperation.hpp
```diff
@@ -7,13 +7,17 @@
 #ifndef UMPIRE_CudaAdviseAccessedByOperation_HPP
 #define UMPIRE_CudaAdviseAccessedByOperation_HPP
```
mcfadden8: To be consistent with ending comment on line 37, I suggest
```suggestion
#define UMPIRE_CudaAdviseOperation_HPP
```

src/umpire/op/CudaAdviseOperation.hpp
```diff
@@ -7,13 +7,17 @@
 #ifndef UMPIRE_CudaAdviseAccessedByOperation_HPP
```
mcfadden8: To be consistent with ending comment on line 37, I suggest

```suggestion
#ifndef UMPIRE_CudaAdviseOperation_HPP
```

mcfadden8 (commented):
Looks good.  Only had two minor suggestions.

mcfadden8 (dismissed):

kab163:
Maybe edit title so that it notes that these are the cuda (mem?) advise operations that are getting a makeover?

davidbeckingsale:
> Maybe edit title so that it notes that these are the cuda (mem?) advise operations that are getting a makeover?

Copy is also refactored...

kab163:
> > Maybe edit title so that it notes that these are the cuda (mem?) advise operations that are getting a makeover?
> 
> Copy is also refactored...

Ah, ok. I guess I'm not as familiar with these cuda operation funcs, so it wasn't as clear.

davidbeckingsale:
> Ah, ok. I guess I'm not as familiar with these cuda operation funcs, so it wasn't as clear.

This PR exposes the "kind" of copy as a parameter, so we can remove the CopyTo/CopyFrom files

https://github.com/LLNL/Umpire/pull/738/files#diff-c702d4779a558f630ed596d43f49ee87d39c1a6abaaf6d6869404350f21dc40d

mcfadden8 (dismissed):

mcfadden8 (approved):

PR #739, opened 2022-04-21
[Task UM998] [Woptim#5] update uberenv camp

adrienbernede:
Opening on behalf of @kab163 for testing purposes.


adrienbernede:
@kab163 @davidbeckingsale,

The initial issue was with uberenv providing a `config.yaml` file that was not up-to-date with recent spack requirements (change in sections declared).

The current issue is with the `hip` package locally defined in umpire: it is outdated compared to the upstream spack version, which translates into the error
```
==> Error: Package 'rocclr' not found in repository '/dev/shm/ruby1/629008/spack/var/spack/repos/builtin'
```
Because `rocclr` is now named `hip-rocclr`.

There probably used to be a reason for this local version on the `hip` package. We should find out if we couldn’t use the spack version now. Do you guys recall why we had to duplicate this package?

kab163:
> @kab163 @davidbeckingsale,
> 
> The initial issue was with uberenv providing a `config.yaml` file that was not up-to-date with recent spack requirements (change in sections declared).
> 
> The current issue is with the `hip` package locally defined in umpire: it is outdated compared to the upstream spack version, which translates into the error
> 
> ```
> ==> Error: Package 'rocclr' not found in repository '/dev/shm/ruby1/629008/spack/var/spack/repos/builtin'
> ```
> 
> Because `rocclr` is now named `hip-rocclr`.
> 
> There probably used to be a reason for this local version on the `hip` package. We should find out if we couldn’t use the spack version now. Do you guys recall why we had to duplicate this package?

Ah!! You found the reason for that error. I couldn't even find what rocclr was (probably because of wrong name). I am all good with using upstream Spack version, though I think David is the better person to ask. The reason why that was originally added was before my time! Thanks for moving this along Adrien!

adrienbernede:
@davidbeckingsale @kab163
We can see the CI is still using the clingo concretizer. This has been bugging me for a while. Because of Marty’s question, I have been working on getting clingo installed on lassen. There too, I found some troubles, then made progress.

- Regarding the concretizer choice, we point to a version of spack that is "too recent" and already dropped the support of the original concretizer...
Since the role here is to support very recent versions of our software, we should move to the new concretizer.

- Regarding the concretizer install, it seems the errors I was facing at first were fixed in spack upstream by the update of clingo version while I was investigating it. Since I point to spack@develop, things started to work without me being able to say what I changed. That is to say: pointing to `develop` is a source of surprise, good or bad.

In the end, clingo is installed on lassen for `umdev` service account!

adrienbernede:
Tasks left:
- [x] point to a fixed commit of spack.
- [x] fix new issues with concretization, showing up when requiring cuda
```
==> Error: umpire@develop%clang@ibm.10.0.1+cuda+ipc_shmem+mpi dev_path=/usr/WS2/umdev/gitlab-runner/umdev/builds/T5vctq-D/015/gitlab/radiuss/Umpire ^cuda@10.1.168 ^spectrum-mpi is unsatisfiable, errors are:
  External package does not satisfy external spec
  no version satisfies the given constraints
    To see full clingo unsat cores, re-run with `spack --show-cores=full`
    For full, subset-minimal unsat cores, re-run with `spack --show-cores=minimized
    Warning: This may take (up to) hours for some specs
```
- [x] fix config to remove the following warning:
```
==> Warning: config:module_roots has been replaced by modules:[module set]:roots and is ignored
```

adrienbernede:
@davidbeckingsale @kab163,

I am pleased to announce the completion of this task. I know I made you wait more than expected... 😅

Among the many small things that changed:
- Update of Spack commit to point to the brand new merge of Raja & Co releases.
- Update of Uberenv to use new fixes, in particular allowing not to install clingo with Uberenv.
- Update of Radiuss-spack-config to update the spack config with a format compatible with spack@0.18.0, and remove the use of a cuda variant that was causing issues. (no more allow-unsupported-compilers)
- Creation of a repo to manage the install of clingo for umdev service user, with control over spack commit, spack location, config etc. (https://github.com/LLNL/radiuss-clingo-install).

I can now document things a bit I guess.

davidbeckingsale (approved):

PR #741, opened 2022-04-21
FixedMallocPool fixes for Windows

### LINKED ISSUE #729: Use-after-free in FixedMallocPool on Windows
**Describe the bug**

Several tests are failing on Windows due to a use-after-free in `util::FixedMallocPool`.

The use-after-free can occur when a `util::FixedMallocPool` object is copy-constructed from another `util::FixedMallocPool` object. This copy construction is shallow, producing two identical `util::FixedMallocPool::m_pool` vectors of `util::FixedMallocPool::Pool` objects that point to the same allocated data. This data is freed when one of the `util::FixedMallocPool` objects is destroyed, leaving the other object with a dangling pointer.

The copy construction occurs on Windows in `strategy::QuickPool`, where `util::FixedMallocPool` is used to implement `strategy::QuickPool::pool_allocator`, which is used as an allocator for an STL container. Microsoft's STL requires that the allocator is copy-constructable, hence the extra code:

https://github.com/LLNL/Umpire/blob/3966120a693c91ea7076653477414139044206fc/src/umpire/strategy/QuickPool.hpp#L116-L120

**To Reproduce**

Please see the "compilers" section below for the CMake and compiler details. The short version is:
```
>cmake ..\Umpire
>cmake --build .
>.\bin\Debug\strategy_tests
```

The tail of the GTest output:
```
[==========] 111 tests from 19 test suites ran. (1488 ms total)
[  PASSED  ] 103 tests.
[  FAILED  ] 8 tests, listed below:
[  FAILED  ] StrategyTest/6.AllocateDeallocateBig, where TypeParam = class umpire::strategy::QuickPool
[  FAILED  ] StrategyTest/6.MultipleAllocateDeallocate, where TypeParam = class umpire::strategy::QuickPool
[  FAILED  ] StrategyTest/6.GetSize, where TypeParam = class umpire::strategy::QuickPool
[  FAILED  ] StrategyTest/6.get_allocator_records, where TypeParam = class umpire::strategy::QuickPool
[  FAILED  ] StrategyTest/6.getCurrentSize, where TypeParam = class umpire::strategy::QuickPool
[  FAILED  ] StrategyTest/6.getActualSize, where TypeParam = class umpire::strategy::QuickPool
[  FAILED  ] ReleaseTest/2.ReleaseCheck, where TypeParam = class umpire::strategy::QuickPool
[  FAILED  ] MixedPool.Host
```

**Expected behavior**

No use-after-free.

**Compilers & Libraries**

Configure:
```
cmake ^
-DCMAKE_CXX_STANDARD=17 ^
-DCMAKE_INSTALL_PREFIX=.\_install ^
-DBLT_CXX_STD= ^
-DUMPIRE_ENABLE_C=On ^
-DUMPIRE_ENABLE_FILESYSTEM=On ^
-DUMPIRE_ENABLE_FILE_RESOURCE=Off ^
..\Umpire
```

Compiler:
- Visual Studio 16 2019
- Windows SDK version 10.0.19041.0
- MSVC 19.27.29111.0

**Additional context**

Umpire version: 3966120a693c91ea7076653477414139044206fc

A simple fix is to delete `util::FixedMallocPool`'s copy constructor, but doing that will require a Windows-specific implementation for `strategy::QuickPool::pool_allocator`. Would that be OK?

degrbg:
This PR closes #729. There are two main changes:

1. Prevent `FixedMallocPool` from being misused by disabling copy construction/assignment (can cause a use-after-free) and by validating the requested allocation size (can change when a `pool_allocator` object is copied, see below).

2. Modify `pool_allocator` to support Windows Debug builds. A `std::shared_ptr` is used to allow `pool_allocator` to be copied without copying the `FixedMallocPool` object. Also, note that `pool_allocator<T>` may be copy constructed from `pool_allocator<U>`, changing the allocation size that is requested from the `FixedMallocPool` object.

   Specifically, `std::multimap<std::size_t, Chunk*, std::less<std::size_t>, pool_allocator<std::pair<const std::size_t, Chunk*>>>` actually constructs an allocator object of type `pool_allocator<std::_Tree_node<std::pair<const std::size_t, Chunk*>, void*>` (type size = 48; this is the size used to construct the `FixedMallocPool` object). This allocator object gets copied to a second allocator object of type `pool_allocator<std::_Container_proxy>` (type size = 16) that shares the same `FixedMallocPool` object. This is safe as long as the copied `pool_allocator`'s type has a size less than or equal to the first `pool_allocator`'s type, but it's good to verify this in `FixedMallocPool`.

Cc: @ctanis


davidbeckingsale (commented):
Hey @degrbg this looks good. Just wondered if there was a reason you added the `allocate_impl` method?

degrbg:
@davidbeckingsale I moved the recursive part of the old `allocate` method to `allocate_impl` to avoid repeatedly `UMPIRE_ASSERT`ing at the beginning of `allocate`.

There is probably a way to do this without recursion, if it is preferable.

davidbeckingsale:
> @davidbeckingsale I moved the recursive part of the old `allocate` method to `allocate_impl` to avoid repeatedly `UMPIRE_ASSERT`ing at the beginning of `allocate`.
> 
> There is probably a way to do this without recursion, if it is preferable.

It's okay, I assumed it was because of the assert but just wanted to confirm. I'm going to make a local branch and PR for this so that we can run it through our full CI pipeline - we can then merge it from there.

degrbg:
@davidbeckingsale It looks like you updated #746 before I could push this update that changes recursion to iteration.

davidbeckingsale:
No worries! Recursion is okay - closing this since #746 was merged.

degrbg:
Thanks!

PR #746, opened 2022-05-02
FixedMallocPool fixes for Windows

### LINKED ISSUE #729: Use-after-free in FixedMallocPool on Windows

davidbeckingsale:
Local copy of #741 
Closes #729


src/umpire/util/FixedMallocPool.hpp
```diff
@@ -42,6 +53,7 @@ class FixedMallocPool {
 
   void newPool();
   void* allocInPool(Pool& p) noexcept;
+  void* allocate_impl(std::size_t bytes);
```
mcfadden8, line 56: What is the purpose for `allocate_impl` instead of `allocate`?  It seems different from the rest of the patterns we use

mcfadden8, line 58: What is the purpose for `allocate_impl` instead of `allocate`?  It seems different from the rest of the patterns we use.  If there is a reason for the difference, a comment would be helpful.

davidbeckingsale: @degrbg answered this on the original PR #741 - it was to avoid repeated assert statement in the recursive allocate call.

mcfadden8: Thanks, please add a comment describing this so that someone isn't tempted to switch it back in the future.

mcfadden8 (approved):

PR #747, opened 2022-05-03
is_device_allocator now returns false if no DeviceAllocator exists yet

kab163:
If no DeviceAllocator has been created yet and a user calls is_device_allocator, it will now return false instead of throwing an error. There is also a check added to the DeviceAllocator test for this.


davidbeckingsale (dismissed):
Looks good, thank you for adding the test case!

davidbeckingsale (approved):

kab163:
I did not realize this didn't have auto-merge turned on! *facepalm*

PR #751, opened 2022-05-23
Fixing plot_allocations script and example

kab163:
Currently, the plot_allocations script is outdated and doesn't work as shown in the docs. This PR updates the plot_allocations tools and the provided example. I also update the documentation that explains how to use the script and the resulting image produced.


mcfadden8 (approved):
Thanks for updating this!

PR #752, opened 2022-05-23
Removing GCC compiler FMT note

kab163:
An unnecessary note gets printed out countless times during some builds using gcc. We thought we got rid of all of those occurrences, but when tools and backtrace is turned on, they showed up again. This PR is to remove that annoying note and then create a CI test that checks for +tools and +backtrace during a gcc + cuda build which was when this issue was first detected.


mcfadden8 (approved):
Looks good to me, thanks!

PR #758, opened 2022-06-16
Add NamedAllocationStrategy support in Replay

kab163:
Some application memory usage involves the NamedAllocationStrategy. In trying to get a replay file from their code, I realized Replay doesn't know how to handle that strategy. This PR attempts to add support within replay so that I can generate my replay files!


tools/replay/ReplayFile.hpp
```diff
@@ -50,7 +50,7 @@ class ReplayFile {
         char accessing_allocator[max_name_length];
       } advisor;
       struct {
-        char named_alloc[max_name_length];
+        int alloc_id;
```
mcfadden8: Since this particular Allocator does not have any additional parameters beyond its name and base allocator, there is now need for this structure.  The entire `struct {} named` may be removed.

kab163: Ah, yes, I was wondering about this. Thanks, I'll remove it!

kab163:
Am I missing an output.omp.replay file? Is it ok that the sqlite replay output file is a binary file?

davidbeckingsale (dismissed):

mcfadden8 (approved):
Looks great to me.  Thanks @kab163 !

mcfadden8:
> Am I missing an output.omp.replay file? Is it ok that the sqlite replay output file is a binary file?

I don't know about the omp replay.  Is not (another) one that is not being tested at the moment?  I am fine with leaving it if so.  If you could perhap write up a Jira ticket on it so we don't forget about it completely , that would be great.

Yes, the sqlite replay is expected to be binary.  It is a database.

kab163:
> > Am I missing an output.omp.replay file? Is it ok that the sqlite replay output file is a binary file?
> 
> I don't know about the omp replay. Is not (another) one that is not being tested at the moment? I am fine with leaving it if so. If you could perhap write up a Jira ticket on it so we don't forget about it completely , that would be great.
> 
> Yes, the sqlite replay is expected to be binary. It is a database.

I will do that next week, along with  adding a link to this PR to the replay notes! Thanks Marty!

PR #761, opened 2022-06-22
[Woptim#27] Install Clingo in temp location with user_cache_path

adrienbernede:
This branch relates to https://github.com/LLNL/RAJA/pull/1284.

We set `user_cache_path` to a temporary and unique location (in shared memory) and install Clingo there too.
Clingo is then reliably installed by each job of the CI, using the binary cache. This solution proves to be fast enough.

We also update Uberenv to fix log ordering.

We update Spack to fix issues with Clingo install.

Notes:
- [x] CI will for one lassen job because of a bug in spack reported there: https://github.com/spack/spack/issues/31214
- [x] I don’t know what I did to deserve so many failures in non-gitlab CI :(

### DIFF
--- a/.gitlab/lassen-build-and-test-extra.yml
+++ b/.gitlab/lassen-build-and-test-extra.yml
@@ -23,7 +23,7 @@ clang_10_0_1_cuda_mpi_shmem:
 # Extras.
 clang_9_0_0_libcpp:
   variables:
-    SPEC: "~fortran %clang@9.0.0+libcpp cxxflags='-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0' cflags='-DGTEST_HAS_CXXABI_H_=0'"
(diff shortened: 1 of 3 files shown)


adrienbernede:
@davidbeckingsale I had tried several things, like double quotes too.
Interesting that quoting the full spec works.

De we agree that we don’t want to apply that change to uberenv though?

davidbeckingsale:
> @davidbeckingsale I had tried several things, like double quotes too. Interesting that quoting the full spec works.
> 
> De we agree that we don’t want to apply that change to uberenv though?

I think quoting the whole spec is correct, but I'm not sure the best place to make that fix.

davidbeckingsale:
@adrienbernede is this ready to go?

adrienbernede:
Oh, I forgot about the workaround.
Let me see if I can get that merged into Uberenv, so that we can point to mainstream commit there.

adrienbernede:
@davidbeckingsale I changed my mind, I opened https://github.com/LLNL/uberenv/pull/84, but the practical approach is to merge this here and update Uberenv if the submitted PR ever gets merged before the bug is fixed in Spack.

davidbeckingsale (approved):

PR #767, opened 2022-08-01
update umpire to match camp and raja exports

trws:
I'm trying this as a solution to the build failure in the upstream spack update: https://gitlab.spack.io/spack/spack/-/jobs/2875546

Thinking we need some more downstream testing too, especially with BLT-based projects.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -37,7 +37,7 @@ stages:
     include:
       - local: '.gitlab/custom-jobs.yml'
       - project: 'radiuss/radiuss-shared-ci'
-        ref: task/remove-intel-18
+        ref: v2022.08.0
         file: '${CI_MACHINE}-build-and-test.yml'
       - local: '.gitlab/${CI_MACHINE}-build-and-test-extra.yml'
     strategy: depend
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -37,7 +37,7 @@ stages:
     include:
       - local: '.gitlab/custom-jobs.yml'
       - project: 'radiuss/radiuss-shared-ci'
-        ref: v2022.06.0
+        ref: task/remove-intel-18
         file: '${CI_MACHINE}-build-and-test.yml'
       - local: '.gitlab/${CI_MACHINE}-build-and-test-extra.yml'
     strategy: depend
--- a/cmake/SetupUmpireThirdParty.cmake
+++ b/cmake/SetupUmpireThirdParty.cmake
@@ -90,8 +90,19 @@ if (NOT BLT_EXPORTED)
   blt_import_library(NAME          blt_stub EXPORTABLE On)
   set_target_properties(blt_stub PROPERTIES EXPORT_NAME blt::blt_stub)
   install(TARGETS blt_stub
-          EXPORT               bltTargets)
+    EXPORT               bltTargets)
   blt_export_tpl_targets(EXPORT bltTargets NAMESPACE blt)
   install(EXPORT bltTargets
-    DESTINATION  lib/cmake/camp)
+    DESTINATION  lib/cmake/umpire)
+elseif (UMPIRE_ENABLE_MPI)
+  # If the target is EXPORTABLE, add it to the export set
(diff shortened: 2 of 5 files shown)


trws:
@davidbeckingsale, @adrienbernede, is there a way to prevent icpc 18 from breaking here without going to the radiuss shared ci stuff?

davidbeckingsale:
Does Intel 18 work if you build as c++14 instead of 17? Seems like the it's working for RAJA so I would expect it should work here without major changes.

trws:
It has been removed from raja already as part of this process, we're explicitly supporting 19 but not 18 for some time now was my understanding.  I don't know, is the direct answer to your question however.

davidbeckingsale:
Oh, in that case, just ditch it here too :)

trws:
So, I went to do that, but it's being pulled in from a separate RADIUSS repo AFAICT?

davidbeckingsale:
Good point - that's the "shared" CI jobs. I've made a branch, should be able to point Umpire at it and check everything works. The `ref` for the shared include is here: https://github.com/LLNL/Umpire/blob/develop/.gitlab-ci.yml#L40 and you can change it to `task/remove-intel-18` (https://github.com/LLNL/radiuss-shared-ci/pull/2)

trws:
Thanks for that @davidbeckingsale!  I was having trouble finding the target repo embarassingly enough, but the last set of tests got to all but a couple on XL thanks to that.  Hopefully with the bltTargets include protection we'll make it and be able to close this out.

trws:
Looks like we're green here. @rhornung, @davidbeckingsale do you want to do other testing here or with the whole unit before we merge to umpire/raja?

davidbeckingsale:
I'm good with this.

trws:
Cool, then feel free to approve and/or merge as desired.

.gitlab-ci.yml
```diff
@@ -37,7 +37,7 @@ stages:
     include:
       - local: '.gitlab/custom-jobs.yml'
       - project: 'radiuss/radiuss-shared-ci'
-        ref: v2022.06.0
+        ref: task/remove-intel-18
```
adrienbernede: Let’s merge `task/remove-intel-18`, and maybe tag the merge `v2022.06.1` before we merge things here.
@trws I know it seems like running down a rabbit hole, am I thankful for your hard work.

cmake/SetupUmpireThirdParty.cmake
```diff
@@ -85,7 +85,24 @@ if (UMPIRE_ENABLE_MPI OR UMPIRE_ENABLE_HIP OR UMPIRE_ENABLE_OPENMP OR UMPIRE_ENA
...
+  # If the target is EXPORTABLE, add it to the export set
+  get_target_property(_is_imported mpi IMPORTED)
+  if(NOT ${_is_imported})
+    install(TARGETS              mpi
+      EXPORT               ${arg_EXPORT})
+    # Namespace target to avoid conflicts
+    set_target_properties(mpi PROPERTIES EXPORT_NAME blt::mpi)
+    install(EXPORT bltTargets
+      DESTINATION  lib/cmake/umpire)
+  endif()
```
adrienbernede: Wouldn’t this be better in the the scope of the previous `if`, i.e. only when BLT tpls are needed ?

.gitlab-ci.yml
adrienbernede: Done

cmake/SetupUmpireThirdParty.cmake
trws: yes, probably.

adrienbernede: OK, done.

.gitlab-ci.yml
trws:
```suggestion
        ref: v2022.06.1
```

adrienbernede: I already applied the required change.

trws: Yup, saw after I did that, planning to remove this in a sec.  Thanks BTW!

davidbeckingsale (approved):

PR #773, opened 2022-08-15
Prefix UMPIRE_ to ENABLE_MPI and ENABLE_IPC_SHARED_MEMORY

mcfadden8:
This prefix was inconsistently applied ENABLE_MPI and
ENABLE_IPC_SHARED_MEMORY which caused some build confusion.

This is documented in https://rzlc.llnl.gov/jira/browse/UM-1016

### DIFF
--- a/scripts/spack_packages/umpire/package.py
+++ b/scripts/spack_packages/umpire/package.py
@@ -229,7 +229,7 @@ class Umpire(CachedCMakePackage, CudaPackage, ROCmPackage):
         spec = self.spec
 
         entries = super(Umpire, self).initconfig_mpi_entries()
-        entries.append(cmake_cache_option("UMPIRE_ENABLE_MPI", '+mpi' in spec))
+        entries.append(cmake_cache_option("ENABLE_MPI", '+mpi' in spec))
 
         return entries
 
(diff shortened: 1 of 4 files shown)


scripts/spack_packages/umpire/package.py
```diff
@@ -229,7 +229,7 @@ def initconfig_mpi_entries(self):
         spec = self.spec
 
         entries = super(Umpire, self).initconfig_mpi_entries()
-        entries.append(cmake_cache_option("ENABLE_MPI", '+mpi' in spec))
+        entries.append(cmake_cache_option("UMPIRE_ENABLE_MPI", '+mpi' in spec))
```
davidbeckingsale: I think this needs to remain `ENABLE_MPI`

mcfadden8: Thank you for the careful review.  I was wondering about this actually.

mcfadden8: Although, aren't we expecting `cmake` users to specify `-DUMPIRE_ENABLE_MPI=On`?  I'm not sure I follow

mcfadden8: What is the intended difference between `ENABLE_MPI` and `UMPIRE_ENABLE_MPI`?

davidbeckingsale (approved):

PR #775, opened 2022-08-19
Only call backtrace once when building a backtrace

mcfadden8:


davidbeckingsale (approved):

davidbeckingsale:
@samuel-schwartz - can you try making this change in your branch and seeing if that speeds things up?

samuel-schwartz:
> @samuel-schwartz - can you try making this change in your branch and seeing if that speeds things up?

Yep! Will do!

PR #777, opened 2022-08-24
Do not call backtrace_symbols if UMPIRE_ENABLE_BACKTRACE_SYMBOLS is off.

mcfadden8:


samuel-schwartz (commented):
Looks good! Merged into Summer2022

PR #779, opened 2022-08-29
[Woptim] update to xl 16 1 1 11, match raja config

adrienbernede:
In order to match RAJA CI configuration, we update xl@16.1.1.10 to xl@16.1.1.11.
We don’t expect any change in the CI results.


adrienbernede:
Closed in favor of #780

adrienbernede:
Closed in favor of #780

PR #788, opened 2022-10-18
Adding HighWatermark Coalescing Heuristic Tuning

kab163:
After the experiments done for the poster, we are going to add a high_watermark option to our heuristics. The user will see options for percent_releasable, percent_releasable_hwm, blocks_releasable, and blocks_releasable_hwm. 

From what we have seen, using the high watermark value when reallocating memory after a coalesce can cause less memory overhead, but could also increase overall runtime (the coalesce function can be called a bit more). 

More testing on this would have to be done to know for sure how it will impact each particular application (and how it will impact the DynamicPoolList as our poster experiments were only done on QuickPool). The defaults for pool coalescing heuristics remain the same.


src/umpire/strategy/DynamicPoolList.cpp
```diff
@@ -130,12 +130,37 @@ void DynamicPoolList::coalesce() noexcept
...
+{
+  return percent_releasable(percentage, false);
+}
+
+PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::percent_releasable_hwm(int percentage)
+{
+  return percent_releasable(percentage, true);
+}
+
+PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::percent_releasable(int percentage, bool hwm_switch)
```
mcfadden8: Can we rename `hwm_switch` to something like `constrain_to_hwm` or `coalesce_to_hwm` or `resize_to_hwm`?  I'm not sure I understand what `hwm_switch` does by its name.

kab163: That is a good point! I like `coalesce_to_hwm` or `resize_to_hwm`.

kab163: @davidbeckingsale any preference here?

davidbeckingsale: `resize_to_hwm`

src/umpire/strategy/DynamicPoolList.hpp
```diff
@@ -34,8 +34,17 @@ namespace strategy {
  */
 class DynamicPoolList : public AllocationStrategy {
  public:
+  /*!
+   * \brief Coalescing Heuristic functions for Percent-Releasable and Blocks-Releasable. Both have
+   * the option to reallocate to High Watermark instead of actual size of the pool (actual size is
+   * currently the default).
+   */
   static PoolCoalesceHeuristic<DynamicPoolList> percent_releasable(int percentage);
+  static PoolCoalesceHeuristic<DynamicPoolList> percent_releasable_hwm(int percentage);
```
mcfadden8, line 43: Should we combine these two functions by using default parameters?

```suggestion
  static PoolCoalesceHeuristic<DynamicPoolList> percent_releasable(int percentage, bool resize_to_hwm = false);
```

mcfadden8 (commented):
We should also add tests to determine whether the two different kinds of heuristics are working as expected.

src/umpire/strategy/DynamicPoolList.hpp
kab163: As per our 11/22 Umpire meeting, we decided this needed more discussion. Having the two separate functions like that is kind of "self documenting" in a nice way, but then again won't scale well if we decide further down the line to have some other size to coalesce to. So, we may need to figure something else out and the fact that we are dealing with lambdas in these functions makes it tough. So: I will resolve this conversation for now, but will open a JIRA ticket for further discussion.

kab163: See UM-1022 on Jira!

kab163:
- [x] Add docs about the new "_hwm" coalescing heuristic option and what it means
- [ ] include a test to make sure that the new coalescing heuristics are allocating to the size we want (might involve locally changing an example file to test)
- [ ] Figure out if more tests are needed than just the one!

kab163:
I changed the `recipe_dynamic_pool_heuristic` so that it allocates 1024 bytes but the minimum size of all other allocations is 8192. 
Without making any changes to the currrent example, the output is:
Pool has 9216 bytes of memory. 3072 bytes are used. 5 blocks are in the pool. 0 bytes are releaseable.
Pool has 9216 bytes of memory. 2048 bytes are used. 5 blocks are in the pool. 0 bytes are releaseable.
Pool has 9216 bytes of memory. 1024 bytes are used. 4 blocks are in the pool. 0 bytes are releaseable.
Pool has 4096 bytes of memory. 0 bytes are used. 1 blocks are in the pool. 0 bytes are releaseable.
With the `_hwm` coalescing function the output is:
Pool has 9216 bytes of memory. 3072 bytes are used. 5 blocks are in the pool. 0 bytes are releaseable.
Pool has 9216 bytes of memory. 2048 bytes are used. 5 blocks are in the pool. 0 bytes are releaseable.
Pool has 9216 bytes of memory. 1024 bytes are used. 4 blocks are in the pool. 0 bytes are releaseable.
Pool has 9216 bytes of memory. 0 bytes are used. 1 blocks are in the pool. 0 bytes are releaseable.
(The last line seems to be the main difference - is there a better way to test this?)

kab163:
Edited documentation pages rendered:
https://umpire.readthedocs.io/en/task-um-1018-add-hwm-coalesce-funcs/sphinx/cookbook/coalesce_pool.html
https://umpire.readthedocs.io/en/task-um-1018-add-hwm-coalesce-funcs/sphinx/cookbook/dynamic_pool_heuristics.html

docs/sphinx/cookbook/coalesce_pool.rst
```diff
@@ -4,10 +4,49 @@
 Coalescing Pool Memory
 ======================
 
-The :class:`umpire::strategy::QuickPool` provides a
-:func:`umpire::strategy::QuickPool::coalesce` that can be used to release
+Umpire's memory pools provide a more performant way to allocate large
+amounts of memory with fewer calls to the underlying API. 
+Memory pools allow developers to allocate all needed memory
```
mcfadden8:
```suggestion
Memory pools allocate and manage a pool of large memory blocks.  When application
allocations are requested, the memory pool allocator will service the allocation from
these memory blocks rather than making (many) allocation requests of the underlying API
which can ...
```

docs/sphinx/cookbook/coalesce_pool.rst
```diff
@@ -4,10 +4,49 @@
...
 
-The :class:`umpire::strategy::QuickPool` provides a
-:func:`umpire::strategy::QuickPool::coalesce` that can be used to release
+Umpire's memory pools provide a more performant way to allocate large
+amounts of memory with fewer calls to the underlying API. 
+Memory pools allow developers to allocate all needed memory 
+at once instead of making multiple, smaller memory allocations which can 
+become quite expensive. However, the performance of memory pools 
+varies widely depending upon how the memory blocks within the pool are managed.
+Each allocation within a memory pool must fit within a `memory block`.
```
mcfadden8:
```suggestion
Each application allocation within a memory pool must fit within a `memory block`.
```

docs/sphinx/cookbook/coalesce_pool.rst
```diff
@@ -4,10 +4,49 @@
...
+As memory is allocated and deallocated, those blocks must be properly 
+adjusted in order to handle new, incoming memory allocations.
+
+Umpire's solution for managing blocks of memory within a pool is to use a
+`coalescing function`. The coalescing function will deallocate a certain amount
+of unused memory and reallocate it, creating a new memory block which can handle new allocations.
+
```
mcfadden8: I would replace the above paragraph with something like the following:
```suggestion
When Umpire is unable to find a block with sufficient memory for the applications' allocation
request, it will allocate a new block from the allocation resource and add that block to the pool.
As application memory use grows, the pool will fill up with many blocks.  Umpire has a `coalescing function` that will deallocate all of the unused (free) blocks in its pool and then reallocate a single block of memory that is large enough to contain all of the freed blocks.
```

mcfadden8 (dismissed):
I added some suggestions for the documentation.  Otherwise, I approve.

Thanks!

davidbeckingsale (approved):

PR #792, opened 2022-11-04
[WOPTIM] Use ci specific reservation on ruby

adrienbernede:
LC created a CI reservation on ruby. I tested it and it’s working.

I suggest we make it the default and see if it works on a regular basis in umpire.


adrienbernede:
@davidbeckingsale we should merge this too. I always forget that it’s not merged already...

davidbeckingsale (approved):

PR #799, opened 2022-12-16
Properly set ID for device allocator

### LINKED ISSUE #798: Using Umpire with MPI and Multiple GPUs "always" uses device 0
**Describe the bug**

I am running my application with MPI and bind a single GPU to each MPI rank as follows:
```
    int myRank= -1;
    MPI_Comm_rank( MPI_COMM_WORLD, &myRank );

    int nDevices=-1;
    cudaGetDeviceCount(&nDevices);

    const int deviceId = myRank % nDevices;
    cudaSetDevice(deviceId);
```

However, when allocating device memory, all allocations from all ranks go to device 0. 

I think this is a consequence of the following lines of code:
https://github.com/LLNL/Umpire/blob/93b1441aaa258c1dcd211a552b75cff6461a2a8a/src/umpire/resource/CudaDeviceMemoryResource.cpp#L26-L29

- Is this correct?
- Is there a way to work around this? Do I need to create a different device allocator to use instead of relying on the default that Umpire provides?

**Expected behavior**
Memory allocations should occur on the device that the host application has set.

**Additional context**

If I bind the GPU when launching the job and set `CUDA_VISIBLE_DEVICES` accordingly, the code works as expected. However, we do ship code out and this approach is not something that we can always rely on.

gzagaris:
# Summary 

The ``umpire::resource::resource_to_device_id()`` method has logic to detect the device ID based on the name of the memory resource. However, if the calling code has set a particular GPU to the process, e.g., via calling ``cudaSetDevice()`` or ``hipSetDevice()``, the default device allocator that Umpire provides may have the wrong device ID. Consequently, all memory allocations/operations would be directed to the wrong GPU on the system.

This commit fixes this issue by adding logic to query the CUDA or HIP runtime for the correct device ID to use if the caller has not encoded the device ID to the name of the memory resource.

Resolves #798

### DIFF
--- a/src/umpire/resource/MemoryResourceTypes.hpp
+++ b/src/umpire/resource/MemoryResourceTypes.hpp
@@ -10,7 +10,16 @@
 #include <cstddef>
 #include <string>
 
+#include "umpire/config.hpp"
 #include "umpire/util/error.hpp"
+
+#if defined(UMPIRE_ENABLE_CUDA)
+#include <cuda_runtime_api.h>
+#endif /* UMPIRE_ENABLE_CUDA */
+
+#if defined(UMPIRE_ENABLE_HIP)
+#include <hip/hip_runtime.h>
+#endif /* UMPIRE_ENABLE_HIP */
 
 namespace umpire {
 namespace resource {
@@ -99,6 +108,16 @@ inline int resource_to_device_id(const std::string& resource)
   int device_id{0};
   if (resource.find("::") != std::string::npos) {
     device_id = std::stoi(resource.substr(resource.find("::") + 2));
+  }
+  else {
+    // get the device bound to the current process
+    #if defined(UMPIRE_ENABLE_CUDA)
+      cudaGetDevice(&device_id);
+    #endif /* UMPIRE_ENABLE_CUDA */
+
+    #if defined(UMPIRE_ENABLE_HIP)
+      hipGetDevice(&device_id);
+    #endif /* UMPIRE_ENABLE_HIP */
   }
   return device_id;
 }


gzagaris:
@davidbeckingsale, @mcfadden8, at your convenience, can you take a quick look at this? I think this is more proper behavior instead of defaulting to device 0 for the default device allocator that Umpire creates (see #798), but, I would be interested to hear your thoughts and suggestions.

davidbeckingsale:
Hey @gzagaris - absolutely, this looks good. However, I think we may have users relying on this behavior.

I think that the typical use case is to use something like CUDA_VISIBLE_DEVICES to expose the correct GPU which will then be "0" even though if would have a different ID if you can see all GPUs on the node.

Let me check and confirm my suspicions.

gzagaris:
> Hey @gzagaris - absolutely, this looks good. However, I think we may have users relying on this behavior.
> 
> I think that the typical use case is to use something like CUDA_VISIBLE_DEVICES to expose the correct GPU which will then be "0" even though if would have a different ID if you can see all GPUs on the node.
> 
> Let me check and confirm my suspicions.

Hi @davidbeckingsale -- I believe these changes should be compatible with the mode where the GPU is selected by setting `CUDA_VISIBLE_DEVICES`. I think that If `CUDA_VISIBLE_DEVICES=1`, `cudaGetDevice()` returns `0`. But, please, let me know if you find out otherwise or if there is another use case that I am not thinking of.

Alternatively, we could add a method to the allocator, e.g., `Allocator::setDevice()` to allow the host application to set the device. I am happy to do that if you think that's a better path forward.

gzagaris:
Hi @davidbeckingsale, just wanted to follow up. Any update or additional thoughts on this?

davidbeckingsale:
No additional thoughts - looks good, will get it merged this week. Thanks!

gzagaris:
> No additional thoughts - looks good, will get it merged this week. Thanks!

Sounds good. Thank you for your help!

gzagaris:
Hey @davidbeckingsale -- just got bitten by this again. 

- What are the plans for getting this merged? 
- Is an Umpire release eminent? 

Thanks again for all your help.

davidbeckingsale:
@gzagaris I have created a local branch for this so we can run internal testing. Will merge as soon as it passes.

Thank you again and sorry for the delay!

gzagaris:
> @gzagaris I have created a local branch for this so we can run internal testing. Will merge as soon as it passes.
> 
> Thank you again and sorry for the delay!

Thanks @davidbeckingsale!

PR #804, opened 2023-01-10
Use blt_convert_to_system_includes instead of blt_patch_target

adayton1:
`blt_patch_target` was causing very strange CMake errors upstream when packages were including Umpire.

```
CMake Error in axom/CMakeLists.txt:
Target "axom" INTERFACE_INCLUDE_DIRECTORIES property contains path:

"/g/g17/dayton8/ale3d/update_compiler_branch/imports/axom/src/axom/core/include"

which is prefixed in the source directory.
```

Using `blt_convert_to_system_includes` fixes the issue.

### DIFF
--- a/src/tpl/CMakeLists.txt
+++ b/src/tpl/CMakeLists.txt
@@ -64,8 +64,7 @@ install(TARGETS
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib)
 
-blt_patch_target(NAME umpire_tpl_json
-                 TREAT_INCLUDES_AS_SYSTEM ON)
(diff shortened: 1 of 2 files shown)


adayton1:
Nevermind, not using system includes at all fixes the issue. It looks like somehow the INTERFACE_INCLUDE_DIRECTORIES ends up with a relative path, but INTERFACE_SYSTEM_INCLUDE_DIRECTORIES ends up with an absolute path.

adayton1:
@white238 said he would submit a fix.

davidbeckingsale:
Fixed by #805

Issue #802, opened 2023-01-10
Generated targets cause problems upstream


adayton1:
**Describe the bug**

```
CMake Error in src/main/CMakeLists.txt:
  Target "main" contains relative path in its
  INTERFACE_INCLUDE_DIRECTORIES:

    "include/umpire"
```

**To Reproduce**

Not sure exactly what triggers this, but I'm seeing similarly weird errors when building Axom with the latest Umpire.

**Additional context**

I'm pretty sure the error is somehow coming from the generator expressions here (we do have Fortran turned off):

```
target_include_directories(
  umpire_interface
  PUBLIC
  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
  $<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:Fortran>:${CMAKE_Fortran_MODULE_DIRECTORY}>>
  $<INSTALL_INTERFACE:$<$<COMPILE_LANGUAGE:Fortran>:include/umpire>> # for Fortran module files
  $<INSTALL_INTERFACE:include>)
```

adayton1:
I'm using CMake 3.21.1.

adayton1:
This is probably a different issue, but it still comes from Umpire and may be related: https://github.com/LLNL/axom/issues/989

adayton1:
```
INTERFACE_INCLUDE_DIRECTORIES "/usr/tce/packages/mvapich2/mvapich2-2.3.6-intel-oneapi.2022.2/include;/g/g17/dayton8/ale3d/update_raja_stack/lib/toss3_intel.oneapi.2022/camp/2022.10.1.a/compiler-intel-oneapi.2022.2-gcc-10.2.1/cpp-14/blt-0.5.2.a/cmake-3.21.1/cuda-none/include;${_IMPORT_PREFIX}/include;\$<\$<COMPILE_LANGUAGE:Fortran>:include/umpire>;${_IMPORT_PREFIX}/include"
```

This comes from the generated umpire-targets.cmake file. I would have thought that the generator expression would have been evaluated by this point, or there would have been a full path rather than a relative path.

adayton1:
Here is the relevant portion, I believe:

`\$<\$<COMPILE_LANGUAGE:Fortran>:include/umpire>`

adayton1:
[...]

PR #805, opened 2023-01-10
Bugfix: Exported include directories were not being relative to Umpire's install prefix

white238:
Due to some generator expression shenanigan's, it is not safe to mark `INSTALL_INTERFACE` generator expressions as system includes. What happens is in the exported target you end up with `include` instead of `${_IMPORT_PREFIX}/include`.  The problem with this is in Axom, it would add `<axom src>/axom/<component name>/include`, which does not exist and is inside of the build directory, then CMake errors out.

Also apparently having the nested generator expression of `COMPILE_LANGUAGE` inside `INSTALL_INTERFACE`, as opposed to the other way around, makes the required relative path of the include directory not exported prefixed with `${_IMPORT_PREFIX}`.

Learn something new everyday.

### DIFF
--- a/src/tpl/CMakeLists.txt
+++ b/src/tpl/CMakeLists.txt
@@ -50,7 +50,13 @@ target_include_directories(
   umpire_tpl_json
   INTERFACE
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/tpl>
-  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
+  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>)
+
+blt_convert_to_system_includes(TARGET umpire_tpl_json)
+
+target_include_directories(
+  umpire_tpl_json
+  INTERFACE
   $<INSTALL_INTERFACE:include>)
 
 install(FILES
@@ -64,8 +70,6 @@ install(TARGETS
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib)
 
-blt_patch_target(NAME umpire_tpl_json
-                 TREAT_INCLUDES_AS_SYSTEM ON)
 
 #
 # CLI11 Option Parsing Headers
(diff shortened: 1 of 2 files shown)


davidbeckingsale (approved):

nselliott:
Trying to build SAMRAI with umpire 2022.10.0, I'm hitting the same issue reported in #802 and fixed here where it only finds an invalid relative path
`  Target "SAMRAI_algs" contains relative path in its`
 ` INTERFACE_INCLUDE_DIRECTORIES:`
`    "include/umpire" `

Since this fix is not in 2022.10.0, is there anything I can do to build with that release, or do I need to wait for an upcoming release?

aaroncblack:
I hit same error as @nselliott, so am waiting on new release.

Until then, I used my rather amateur CMake skills to hack it on my side:
```
   get_target_property(UMPIRE_INCLUDES umpire INTERFACE_INCLUDE_DIRECTORIES)
   string(FIND "${UMPIRE_INCLUDES}" "\$<\$<COMPILE_LANGUAGE:Fortran>:include/umpire" FOUND)
   if (NOT FOUND EQUAL -1)
     message(WARNING "Found bad include path in Umpire target.  This is reported in https:://github.com/LLNL/Umpire/issues/802")
     message(WARNING "Umpire include dirs: ${UMPIRE_INCLUDES}")
     string(REPLACE "\$<\$<COMPILE_LANGUAGE:Fortran>:include/umpire" "\$<\$<COMPILE_LANGUAGE:Fortran>:${UMPIRE_ROOT}/include/umpire" UMPIRE_INCLUDES_FIXED "${UMPIRE_INCLUDES}")
     message(WARNING "Umpire include dirs, after scrubbing: ${UMPIRE_INCLUDES_FIXED}")
     set_target_properties( umpire PROPERTIES
        INTERFACE_INCLUDE_DIRECTORIES "${UMPIRES_INCLUDES_FIXED}"
     )
   endif()
```

My hack only works if your CMake code knows where Umpire is ( the dev sets it via UMPIRE_ROOT in my CMake ).  Although maybe there is a way to fish out the Umpire path from its exported properties.

PR #819, opened 2023-03-21
sycl: support 2023 oneapi release

bd4:
The CL/sycl.hpp header has been removed in the 2023 release. Note that the SYCL 2020 standard prescribes sycl/sycl.hpp as location, so this change in oneapi brings it in accordinace with the standard.

dpcpp has also been depreciated, using icpx with -fsycl is now preferred and dcpp will be removed. This attempts to preserve support for older oneapi releases via the sycl_compat.hpp header.


bd4:
Needed for https://github.com/wdmapp/gtensor/pull/257. For gtensor, I try to always support the latest public oneapi release, and make some effort to support the previous release. I am not sure it is worth the effort to support older releases, as there are major improvements in each release and they are often required to take advantage of the emerging Intel HPC class GPUs.

davidbeckingsale:
@bd4 sorry for the delay, this looks good. One change - could you move the sycl_compat file to the util subdirectory?

bd4:
> @bd4 sorry for the delay, this looks good. One change - could you move the sycl_compat file to the util subdirectory?

Sure pushed the rename, it builds for me with oneAPI 2023.1.0.

PR #821, opened 2023-04-17
Add Synchronizer strategy

davidbeckingsale:


MrBurmark (dismissed):

mcfadden8 (changes requested):
For testing, I think it would be good to have a test (that could also serve as an example) that simply creates an allocator of this kind and uses it for a couple of allocations and deallocations.  This would be a minimum test to make sure that everything links up and the calls all successfully pass through without error.

src/umpire/event/event.hpp
```diff
@@ -15,10 +15,45 @@
...
 #include "camp/camp.hpp"
+#include "camp/resource.hpp"
 #include "umpire/event/recorder_factory.hpp"
 #include "umpire/util/Macros.hpp"
 
+namespace camp {
+namespace resources {
+inline namespace v1 {
+
+inline std::string to_string(camp::resources::Resource& r)
```
MrBurmark, line 26: We should fix the constness of get_platform in camp.

davidbeckingsale:
Will be superseded by other features.

PR #823, opened 2023-04-28
adding _hwm replay option for both blocks and free_percentage heuristics

kab163:
The purpose of this PR is to add a _hwm option to replay. I would like to have a .replay or .stats file and be able to use the replay tool to tell me what would happen if I had used a highwatermark option with my pool. For example, replay should now have
```
Options:
  -h,--help                   Print this help message and exit
  -i,--infile TEXT:FILE REQUIRED
                              Input file
  -q,--quiet                  Only errors will be displayed.
  -t,--time-run               Display time information for replay running operations
  -d,--dump                   Dump ULTRA memory usage trace for each Allocator
  -s,--stats                  Track/Display pool allocation size statistics
  --no-demangle               Disable demangling of replay file
  --skip-operations           Skip Umpire Operations during replays
  -r,--recompile              Force recompile replay binary
  -p,--use-pool TEXT          Specify pool to use: List or Quick
  --use-heuristic TEXT        Heuristic: Block, Block_hwm, FreePercentage, or FreePercentage_hwm
  --heuristic-parm INT:INT in [0 - 100]
                              Heuristic parameter to use
```
And I could use `--use-heuristic Block_hwm` to test the Blocks releasable heuristic with a highwatermark tuning on my pool.


kab163:
@davidbeckingsale @mcfadden8 - I reverted the replay output files back to develop versions since that was not needed

mcfadden8:
Why weren't they needed?  Is there not a replay test for running with the HWM heuristic?  (it has been a while since I last looked at this)

kab163:
> Why weren't they needed? Is there not a replay test for running with the HWM heuristic? (it has been a while since I last looked at this)

See MM chat

mcfadden8 (approved):
Looks great, thanks @kab163!

PR #824, opened 2023-05-03
fixing the name of kernel called in cuda test

kab163:
Typo discovered in name of the test for the DeviceAllocator when CUDA enabled.


mcfadden8 (approved):
ok

Issue #652, opened 2021-09-16
Link hypre with umpire using host linker


liruipeng:
Hi, Umpire developers,

We now use host linker for the drivers in hypre. With Umpire, we got the following error,
```
Building ij ... 
mpixlC -o ij ij.o -L/g/g92/li50/workspace/hypre/test/src/hypre/lib -lHYPRE -Wl,-rpath,/g/g92/li50/workspace/hypre/test/src/hypre/lib           -lm  -L/usr/tce/packages/cuda/cuda-10.1.243/lib64 -lcudart -lcusparse -lcurand      -L/usr/workspace/li50/Umpire-git/Umpire/install_xlC_lassen/lib  -lumpire 
/usr/workspace/li50/Umpire-git/Umpire/install_xlC_lassen/lib/libumpire.a(Allocator.cpp.o): In function `__sti____cudaRegisterAll()':
/var/tmp/li50/tmpxft_000189aa_00000000-5_Allocator.cudafe1.cpp:(.text+0xac0): undefined reference to `__cudaRegisterLinkedBinary_44_tmpxft_000189aa_00000000_6_Allocator_cpp1_ii_a17095a1'
/usr/workspace/li50/Umpire-git/Umpire/install_xlC_lassen/lib/libumpire.a(Replay.cpp.o): In function `__sti____cudaRegisterAll()':
/var/tmp/li50/tmpxft_000189ab_00000000-5_Replay.cudafe1.cpp:(.text+0x5d4): undefined reference to `__cudaRegisterLinkedBinary_41_tmpxft_000189ab_00000000_6_Replay_cpp1_ii_5eca6429'
/usr/workspace/li50/Umpire-git/Umpire/install_xlC_lassen/lib/libumpire.a(ResourceManager.cpp.o): In function `__sti____cudaRegisterAll()':
/var/tmp/li50/tmpxft_000189ac_00000000-5_ResourceManager.cudafe1.cpp:(.text+0x22134): undefined reference to `__cudaRegisterLinkedBinary_50_tmpxft_000189ac_00000000_6_ResourceManager_cpp1_ii_42a9a1b2'
/usr/workspace/li50/Umpire-git/Umpire/install_xlC_lassen/lib/libumpire.a(Umpire.cpp.o): In function `__sti____cudaRegisterAll()':
[...]

PR #827, opened 2023-05-30
Make `umpire_device` separate library

msimberg:
Attempt to address this comment: https://github.com/LLNL/Umpire/issues/652#issuecomment-1527288786.

This makes `umpire_device` a separate library that has to be linked explicitly in addition to linking to `umpire` if using device allocators. The motivation for this is that in many use cases `umpire_device` is not used, but bundling it into `umpire` adds a requirement for downstream projects (even those that are not directly using Umpire) to enable separable compilation on their CMake targets. This option can't be required transitively from Umpire itself or projects using Umpire directly, so this PR at least makes `umpire_device` opt-in only for those cases that it's used.

If device allocators are actually used, downstream projects still have to enable separable compilation explicitly; this does not change that.

The only slightly non-obvious change here is the change to sequence destruction of the device allocators before destruction of the resource manager. This is currently done by relying on the fact that function local statics are destroyed in the opposite order of construction. This assumes that `make_device_allocator` is the only function that allocates the memory needed for device allocators. The destruction of the device allocators was moved to `umpire_device` so that `umpire` does not need to handle device allocator destruction (`umpire_device` depends on `umpire`, not vice-versa). Let me know if this is understandable or if you think another approach would be cleaner.

I've fixed what I assume to be a typo in the `device_allocator` test, but I don't know if that is the correct fix.

All tests except one test passes for me locally. The failing test `memory_resource_traits_tests` also fails on `develop` for me though, so I'm assuming I didn't break that (the failure looks like this: https://gist.github.com/msimberg/4960d7e32aa22fdaeaae40221c7a1ff7).


msimberg:
Ping? Is this something you'd still be interested in?

msimberg:
Hi there, this is still something I'd be interested in having in umpire. However, are you interested in having it? It'd be nice to get a response whether this is something I can expect to get in at some point (even if you may want changes to it before that).

davidbeckingsale:
> Hi there, this is still something I'd be interested in having in umpire. However, are you interested in having it? It'd be nice to get a response whether this is something I can expect to get in at some point (even if you may want changes to it before that).

Yes! Definitely interested - I'm currently in the process of surveying other users to make sure this doesn't cause any unintended impacts. Thank you

msimberg:
Thanks @davidbeckingsale for getting back to me! All right, that sounds good. At least then I know this hasn't slipped through the cracks, and I fully understand that you want to make sure this doesn't break anything.

PR #828, opened 2023-05-31
Add transfer function

davidbeckingsale:


davidbeckingsale:
cc @MrBurmark

src/umpire/ResourceManager.cpp
```diff
@@ -359,6 +359,23 @@ util::AllocationRecord ResourceManager::deregisterAllocation(void* ptr)
   return m_allocations.remove(ptr);
 }
 
+void* ResourceManager::transfer(void* ptr, Allocator to)
+{
+  auto alloc_record = findAllocationRecord(ptr);
+
+  if (!(alloc_record->strategy->treatAsPassthrough() && to.getAllocationStrategy()->treatAsPassthrough())) {
```
MrBurmark, line 366: So this does a direct transfer when the underlying allocators are the same, but moves when they are different? Wouldn't this allow a direct transfer between named allocators built on different base allocators? Should this take a resource to use for the copy?

MrBurmark (commented):
This looks reasonable as a start.
I have a couple of questions about certain cases. I don't remember all of the constraints on this so the implementation can remain simple.
Should there be a special case for transferring to and from the same allocator?
What about transferring from a named allocator to the underlying allocator and vice-versa?
Does this mess up the stats/counters?

PR #831, opened 2023-07-10
Fixed coalescing call for blocks releasable heuristic

asidhu0:
The bodies of the two functions, `blocks_releasable(std::size_t nblocks)` and `blocks_releasable_hwm(std::size_t nblocks)`, were swapped so I switched the code of these two functions. Now `blocks_releasable` uses the actual size and `blocks_releasable_hwm` uses the high-watermark. In addition, I passed the right value into `blocks_releasable(std::size_t nblocks)` and `blocks_releasable_hwm(std::size_t nblocks)` by calling the heuristic before the coalesce call happens.

### DIFF
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -122,7 +122,10 @@ bool DynamicPoolList::tracksMemoryUse() const noexcept
 void DynamicPoolList::coalesce() noexcept
 {
   UMPIRE_LOG(Debug, "()");
-  
+  umpire::event::record([&](auto& event) {
+    event.name("coalesce").category(event::category::operation).tag("allocator_name", getName()).tag("replay", "true");
+  });
+    
   std::size_t suggested_size{m_should_coalesce(*this)};
   if (0 != suggested_size) {
     UMPIRE_LOG(Debug,
@@ -136,14 +139,14 @@ void DynamicPoolList::coalesce() noexcept
 PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::blocks_releasable(std::size_t nblocks)
 {
   return [=](const strategy::DynamicPoolList& pool) {
-    return pool.getReleasableBlocks() > nblocks ? pool.getActualSize() : 0;
+    return pool.getReleasableBlocks() >= nblocks ? pool.getActualSize() : 0;
   };
 }
 
 PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::blocks_releasable_hwm(std::size_t nblocks)
 {
   return [=](const strategy::DynamicPoolList& pool) {
-    return pool.getReleasableBlocks() > nblocks ? pool.getHighWatermark() : 0;
+    return pool.getReleasableBlocks() >= nblocks ? pool.getHighWatermark() : 0;
   };
 }
 
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -125,7 +125,7 @@ void DynamicPoolList::coalesce() noexcept
   umpire::event::record([&](auto& event) {
     event.name("coalesce").category(event::category::operation).tag("allocator_name", getName()).tag("replay", "true");
   });
-    
+
   std::size_t suggested_size{m_should_coalesce(*this)};
   if (0 != suggested_size) {
     UMPIRE_LOG(Debug,
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -122,10 +122,15 @@ bool DynamicPoolList::tracksMemoryUse() const noexcept
 void DynamicPoolList::coalesce() noexcept
 {
   UMPIRE_LOG(Debug, "()");
-  umpire::event::record([&](auto& event) {
-    event.name("coalesce").category(event::category::operation).tag("allocator_name", getName()).tag("replay", "true");
-  });
-  dpa.coalesce(dpa.getActualSize());
+  
+  std::size_t suggested_size{m_should_coalesce(*this)};
+  if (0 != suggested_size) {
+    UMPIRE_LOG(Debug,
+               "Heuristic returned true, "
+               "performing coalesce operation for "
+                   << this << "\n");
+    dpa.coalesce(suggested_size);
+  }
 }
 
 PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::blocks_releasable(std::size_t nblocks)
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -122,7 +122,7 @@ bool DynamicPoolList::tracksMemoryUse() const noexcept
 void DynamicPoolList::coalesce() noexcept
 {
   UMPIRE_LOG(Debug, "()");
-  
+
   std::size_t suggested_size{m_should_coalesce(*this)};
   if (0 != suggested_size) {
     UMPIRE_LOG(Debug,
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -125,13 +125,10 @@ void DynamicPoolList::coalesce() noexcept
   umpire::event::record([&](auto& event) {
     event.name("coalesce").category(event::category::operation).tag("allocator_name", getName()).tag("replay", "true");
   });
-    
+
   std::size_t suggested_size{m_should_coalesce(*this)};
   if (0 != suggested_size) {
-    UMPIRE_LOG(Debug,
-               "Heuristic returned true, "
-               "performing coalesce operation for "
-                   << this << "\n");
+    UMPIRE_LOG(Debug, "coalesce heuristic true, performing coalesce, suggested size is " << suggested_size);
     dpa.coalesce(suggested_size);
   }
 }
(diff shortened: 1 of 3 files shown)


mcfadden8:
Thanks @asidhu0 -

Can we update the test fail such that it will make sure that these heuristics are producing the desired behavior?

kab163:
Update:
TODOs for review:
- [x] Create a test for the heuristics in pool_heuristics_tests.cpp to make sure these heuristics have desired behavior
- [x] Change PR name (make it shorter)

src/umpire/strategy/DynamicPoolList.cpp
```diff
@@ -122,10 +122,15 @@ bool DynamicPoolList::tracksMemoryUse() const noexcept
 void DynamicPoolList::coalesce() noexcept
 {
   UMPIRE_LOG(Debug, "()");
-  umpire::event::record([&](auto& event) {
```
mcfadden8, line 125: This `umpire::event::record` line should not be removed as it is needed for the `replay` tool.

src/umpire/strategy/QuickPool.cpp
```diff
@@ -312,11 +312,11 @@ void QuickPool::coalesce() noexcept
 {
   UMPIRE_LOG(Debug, "()");
 
-  umpire::event::record([&](auto& event) {
-    event.name("coalesce").category(event::category::operation).tag("allocator_name", getName()).tag("replay", "true");
```
mcfadden8, line 313: This is needed by the `replay` tool and should not be removed.

tests/unit/strategy/pool_heuristics_tests.cpp
```diff
@@ -119,3 +119,32 @@ TYPED_TEST(PoolHeuristicsTest, BlocksReleasable)
   ASSERT_EQ(a.second->getReleasableBlocks(), 1);
   ASSERT_EQ(a.second->getTotalBlocks(), 1);
 }
+
```
mcfadden8, line 159: I suggest that you leave a comment in here about the assumptions this test is making about the size of the blocks and what is being done to account for any alignment that the underlying allocator is doing.  I read through the code and can see how it would work, but leaving a few breadcrumb comments might help me when/if I look at it again a few years from now :-).

asidhu0:
@mcfadden8 @kab163 @davidbeckingsale 

Another thing I observed is that if we set the heuristic to be `blocks_releasable_hwm(2)`, for example, the coalescing will automatically happen only when 3 blocks are releasable, not 2. 

https://github.com/LLNL/Umpire/blob/27594e401b7c0633b8849875ccdda4a73f5eadbc/src/umpire/strategy/QuickPool.cpp#L343-L348

This is the current behavior because on line 346 the greater than sign is used. So if there are 2 blocks releasable and our heuristic is `blocks_releasable_hwm(2)` then both `pool.getReleasableBlocks()` and `nblocks` are 2. And because of the greater than sign, 0 is returned and the pool does not coalesce although there are 2 releasable blocks.

src/umpire/strategy/QuickPool.cpp
asidhu0: Thanks for letting me know! I'll bring it back.

src/umpire/strategy/DynamicPoolList.cpp
asidhu0: Will do!

tests/unit/strategy/pool_heuristics_tests.cpp
asidhu0: For sure! For my test I also allowed the pool to coalesce itself and did not manually call the coalesce. function But to do this I had to make 3 blocks releasable not 2 although the heuristic I used was `blocks_releasable_hwm(2)`. I explain this discrepancy in my comment below.

mcfadden8:
I believe that this is a bug.  IIRC, `blocks_releasable...` was to return true when there were N or more blocks available.  So I think the heuristic  should be testing with '>=' instead.  Good catch.

asidhu0:
Should I make the change of having the heuristics use the '>=' instead to this branch?

mcfadden8:
I think so yes.  But I'd like to have @davidbeckingsale chime in as well before you do.

src/umpire/strategy/QuickPool.cpp
mcfadden8: It looks like the (failing) tests are letting you know as well.  I believe if you put these lines back, the replay tests will succeed again.

src/umpire/strategy/DynamicPoolList.cpp
```diff
@@ -125,20 +125,28 @@ void DynamicPoolList::coalesce() noexcept
...
     event.name("coalesce").category(event::category::operation).tag("allocator_name", getName()).tag("replay", "true");
   });
-  dpa.coalesce(dpa.getActualSize());
+
+  std::size_t suggested_size{m_should_coalesce(*this)};
+  if (0 != suggested_size) {
+    UMPIRE_LOG(Debug,
+               "Heuristic returned true, "
+               "performing coalesce operation for "
+                   << this << "\n");
```
mcfadden8: What does the output of this debug log look like?  You can enable it by running with the `UMPIRE_LOG_LEVEL` environment variable set to `Debug` and then look at the debug file output for this particular `UMPIRE_LOG`.

tests/unit/strategy/pool_heuristics_tests.cpp
```diff
@@ -110,12 +110,40 @@ TYPED_TEST(PoolHeuristicsTest, BlocksReleasable)
 
   for (int i{max_blocks}; i > 0; i--) {
     ASSERT_NO_THROW(a.first.deallocate(ptrs[i - 1]););
-
-    if (i % 2)
-      ASSERT_EQ(a.second->getReleasableBlocks(), 1);
-    else
-      ASSERT_EQ(a.second->getReleasableBlocks(), 2);
+    ASSERT_EQ(a.second->getReleasableBlocks(), 1);
```
mcfadden8, line 153: This looks so much better.

asidhu0: Yup, if we use `>=` there can't be 2 releasable blocks since they should coalesce automatically. So either a block will be releasable right after its been deallocated or right after its been coalesced (in both cases there is only 1 releasable block).

src/umpire/strategy/DynamicPoolList.cpp
asidhu0: There was no output for that part of the debug log.

mcfadden8: You should see some if that code gets executed and a coalesce happening.  I'm concerned because I do not know what is expected with `<< this <<` in the input.  What is is that you are trying to log here?  Is it necessarily?  If so, think you need to change it somewhat to display the information that will be useful for debugging.  

Also, if you do decide to keep it, we also need to change `"\n"` to `std::endl`

You might want to work with @kab163 on this because you should being seeing output in the file when this code is executed and the environment variable is set.

mcfadden8 (dismissed):
Thanks!  This looks good to me.

asidhu0:
@mcfadden8 @davidbeckingsale 

Any ideas on why some of the tests are failing and how to debug them? 

The error first shows up on line 1304 on this [link](https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/1389843). The tests that I wrote passed but I think there might be a memory leak happening during the tear down.

davidbeckingsale:
> @mcfadden8 @davidbeckingsale
> 
> Any ideas on why some of the tests are failing and how to debug them?
> 
> The error first shows up on line 1304 on this [link](https://lc.llnl.gov/gitlab/radiuss/Umpire/-/jobs/1389843). The tests that I wrote passed but I think there might be a memory leak happening during the tear down.

Yup looks like the test introduced a leak. You could try replacing the vector<void*> with a regular array (you know how many items it will hold so the vector is unnecessary). 

There's a line in the CI output that will show you how to reproduce the configuration, include enabling address sanitizer, so you can test your changes locally. Lines 72-81.

kab163:
Should the test also include a Percent_Releasable_hwm test?

asidhu0:
@davidbeckingsale @mcfadden8 @kab163 

An observation I had for my `PercentReleasableHWM` test, was that after coalescing when I wanted to deallocate the rest of the pool (to avoid a memory leak), I had to call `release` first and then I could deallocate the rest of the memory. If I didn't call `release` then the test for `QuickPool` would result in an infinite loop where `coalesce` would keep calling `deallocate` and vice versa. However, this problem was not observed in `DynamicPool`. I think the reason why is because in `DyanmicSizePool.hpp` the `deallocate` function does not call its corresponding `coalesce` function while in `QuickPool.cpp`, `deallocate` and `coalesce` continuously call each other if certain conditions are met. This might be intended behavior because the two pools are different but I just wanted to state what I saw happening.

kab163:
> @davidbeckingsale @mcfadden8 @kab163
> 
> An observation I had for my `PercentReleasableHWM` test, was that after coalescing when I wanted to deallocate the rest of the pool (to avoid a memory leak), I had to call `release` first and then I could deallocate the rest of the memory. If I didn't call `release` then the test for `QuickPool` would result in an infinite loop where `coalesce` would keep calling `deallocate` and vice versa. However, this problem was not observed in `DynamicPool`. I think the reason why is because in `DyanmicSizePool.hpp` the `deallocate` function does not call its corresponding `coalesce` function while in `QuickPool.cpp`, `deallocate` and `coalesce` continuously call each other if certain conditions are met. This might be intended behavior because the two pools are different but I just wanted to state what I saw happening.

Ok so I think this is something that we should look more in to later, but is ok for now. Why don't you create a Github Issue about this behavior so we can track it. Just add a good description like you did here to the issue and any other details you think would be helpful for reproducing this later when we have more time to address it. I'll take one last pass over this today and approve it.

kab163 (approved):

PR #832, opened 2023-07-12
[woptim] Update uberenv to use environments

adrienbernede:
In this PR I intend to update Uberenv to use environments in Spack.

There is a issue with flags stacking I am trying to reproduce.


adrienbernede:
replaced by https://github.com/LLNL/Umpire/pull/875

PR #834, opened 2023-07-18
Allow using external fmt

adayton1:


src/umpire/event/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ set (umpire_event_sources
   json_file_store.cpp
   recorder_factory.cpp)
 
-set (umpire_event_depends umpire_tpl_fmt umpire_tpl_json camp)
+set (umpire_event_depends fmt::fmt umpire_tpl_json camp)
```
adayton1, line 20: Do we want to use the header only library?

src/umpire/util/error.hpp
```diff
@@ -10,7 +10,7 @@
 #include <stdexcept>
 #include <string>
 
-#include "umpire/fmt/fmt.hpp"
+#include "fmt/format.h"
```
adayton1, line 13: Are we OK with exposing fmt in a header?

adayton1:
@davidbeckingsale, what should I do about the spack package for Umpire needing fmt? There's a circular dependency between radiuss-spack-packages and Umpire.

adayton1:
I know I need to add something like:

`depends_on("fmt@9.1:10", when="@develop")`

src/umpire/util/error.hpp
adayton1: Looks like it's not even used in this file. So now it shouldn't bleed out of Umpire's headers.

adayton1: Oh, but it will propagate into other headers, now.

src/umpire/event/CMakeLists.txt
adayton1: The library is built, so I won't worry about it for now.

src/umpire/util/error.hpp
adayton1: It would be nice to avoid this, but I think that's better left as a future exercise.

adayton1:
It passed!

kab163 (approved):
Looks good. I no longer see all the fmt notes/warnings when building this branch with the same cmake command as the one from my issue!

PR #837, opened 2023-07-20
Make allocate API threadsafe when needed

mcfadden8:
The `ThreadSafeAllocator` strategy was only placing a mutex around the allocate and deallocate calls and tracking and zero-byte-allocation calls were left unprotected since they were being performed out of scope from the `ThreadSafeAllocator` strategy.

The fix was to move the mutex usage up to the `Allocator` API level where the zero-byte-allocation, tracking, and allocate/deallocate calls were being made.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ else
 	DebugArgs=
 endif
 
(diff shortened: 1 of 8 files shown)


mcfadden8:
@davidbeckingsale - This PR is ready for review.  It passes our CI and the problem reporter has confirmed that the change has addressed their concern.

davidbeckingsale (approved):

PR #839, opened 2023-07-26
Add print of repro instructions in CI script

davidbeckingsale:


rhornung67 (commented):
I like this. Do you mind if I steal it and use it in RAJA projects?

scripts/gitlab/build_and_test.sh
```diff
@@ -253,7 +253,15 @@ then
     # make clean
 fi
 
+cd ${project_dir}
 echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
 echo "~~~~~ Build and test completed"
 echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
+echo "~~~~~ To reproduce this build:"
+echo ""
+echo " #`git rev-parse HEAD` on host ${truehostname}"
```
mcfadden8: Thanks @davidbeckingsale - This looks great!  Would you mind changing the following 3 lines to:
```suggestion
echo "On host ${truehostname}, run:"
echo ""
echo "git checkout `git rev-parse HEAD` && git submodule init && git submodule update"
echo "SPACK_DISABLE_LOCAL_CONFIG=\"\" SPACK_USER_CACHE_PATH=\"../ci_spack_cache\" python3 scripts/uberenv/uberenv.py \"${spec}\" --prefix=\"../ci_uberenv\""
echo "mkdir build && cd build"
echo "cmake .. -C ../${truehostname}-\*.cmake"
echo "make && make test"
```

mcfadden8: This would make the output look something like:
```
On host corona, run:

git checkout fe91815d1850310ec894e2d2eae1d6b89cabafca && git submodule init && git submodule update
SPACK_DISABLE_LOCAL_CONFIG="" SPACK_USER_CACHE_PATH="../ci_spack_cache" python3 scripts/uberenv/uberenv.py "~shared +fortran +device_alloc tests=basic  +rocm amdgpu_target=gfx906 %rocmcc@5.5.0 ^hip@5.5.0 " --prefix="../ci_uberenv"
mkdir build && cd build
cmake .. -C ../corona-\*.cmake
make && make test
```

scripts/gitlab/build_and_test.sh
```diff
@@ -253,7 +253,15 @@ then
...
+echo ""
+echo "SPACK_DISABLE_LOCAL_CONFIG=\"\" SPACK_USER_CACHE_PATH=\"ci_spack_cache\" python3 scripts/uberenv/uberenv.py \"${spec}\""
```
mcfadden8: The "--prefix" option needs to be added in order for the subsequent cmake to work.  I would suggest the following change which I have tested locally and found it to work.  The idea is that the uberenv built stuff needs to be in a directory outside of the umpire source directory (see email from Adrien to Umpire group):
```suggestion
echo "SPACK_DISABLE_LOCAL_CONFIG=\"\" SPACK_USER_CACHE_PATH=\"ci_spack_cache\" python3 scripts/uberenv/uberenv.py \"${spec}\" --prefix=\"../ci_uberenv\""
```

davidbeckingsale:
> I like this. Do you mind if I steal it and use it in RAJA projects?

No problem! We're making a few tweaks but I think that whatever we merge you can just drop right in.

mcfadden8 (approved):
Looks great, thanks!

adrienbernede:
@davidbeckingsale @rhornung67 @mcfadden8 Don’t you like the reproducer already printed by RADIUSS Shared CI ? In facts, it looks like it’s the build_and_test script you may not like that much...

I can understand the need to use /tmp instead of /dev/shm, and I can integrate that in my reproducer.

Last week, I actually improved the build_and_test script so that it allows to easily turn off /dev/shm (not merged yet). I could actually have /dev/shm off by default and activate it in CI.

IMHO, it’s a win if we use the same script in CI as for dev, so I welcome feedback.

(Note: Printing the reproducer at the end of the script will only show the reproducer on successful runs).

mcfadden8:
@adrienbernede, What does the reproducer line look like in the current CI and where is it located in the logs?

adrienbernede:
The reproducer is as early in the logs as I could put it, right after the runners initialization logs.

```
### Tioga reproducer
xsu umdev
working_dir="/usr/workspace/${USER}/Umpire/woptim-shared-ci-2023-08-0-317490-$(date +%s)"
mkdir -p ${working_dir} && cd ${working_dir} 
git clone https://github.com/LLNL/Umpire.git --single-branch --depth=1 
cd Umpire
git fetch origin --depth=1 1efe221c5be3d697d3ef594528b1c8d2a0be4890 
git checkout 1efe221c5be3d697d3ef594528b1c8d2a0be4890 
git submodule update --init --recursive 
SPEC="~shared +fortran tests=basic %cce@16.0.0" MODULE_LIST="" flux mini run --time-limit=15m --nodes=1 --begin-time=+5s ./scripts/gitlab/build_and_test.sh
### End of reproducer
```

mcfadden8:
Cool!  I didn't realize that this even existed!  I always just did a search for "SPEC" and copy/pasted it in to my own uberenv command.  This is even better since it will avoid me having to interact directly with uberenv and it will enable me to get the exact same checkout of code.

adrienbernede:
It’s an exact reproducer of the CI job. The main drawback is that it builds the deps and the project in /dev/shm, which I am working on improving. Using /dev/shm is ideal for performance (responsive CI), but not for debugging (no persistence).

For the records there are actually 3 directories that must be set somehow:
1. Spack user cache (isolate spack from your environment)
2. Spack install dir (for dependencies)
3. Project build dir.

It looks like you agreed that 1. should be in /tmp. Do you have a preference for the location of 2. and 3. ?

mcfadden8:
Could 2 and 3 be placed in an ignored (by git) directory at the top of the Umpire directory that was checked out?

adrienbernede:
2 cannot be placed in the Umpire directory, otherwise the dependencies would be installed as part of the Umpire source tree, which causes errors.

The Axom way is to place 2 and 3 in a directory next to the Umpire clone. (e.g. `/path/to/Umpire/../Umpire_deps_and_builds/`) What do you think @mcfadden8  ?

mcfadden8:
Placing them in peer directory is fine with me.

PR #841, opened 2023-08-04
Fix rocm compiler errors

mcfadden8:


mcfadden8:
@trws and @davidbeckingsale, I made the [two one-line changes](https://github.com/LLNL/Umpire/pull/841/files#diff-e7de5bcfca5b979b02d7937362cf08e7bb15a2e2b29affd50329bac1a71e171f) in this PR so that rocm 5.6 would work (previous versions and other compilers worked fine).  But I am not sure I understand why this would have been a problem in the first place.  Can either of you explain what the difference is with using `new` instead of `std::make_shared` in this case?  Does one work better than the other?  Which way is actually correct/preferred?

davidbeckingsale (approved):

trws:
Either `std::make_shared` or `std::make_unique` are always preferred where appropriate.  I'm not sure why using `new` there would *fail*, but it's still not recommended. The reason is for exception safety.  If you use new, and an exception is thrown in creating the shared_ptr or unique_ptr from it, then the memory is leaked or in some cases may potentially produce UB.  Using `make_*` ensures that case never happens.

mcfadden8:
Thanks @trws !

PR #843, opened 2023-08-16
fixed getReleasableSize()

asidhu0:
In order to fix the getReleasableSize() function for both `QuickPool.cpp` and `DynamicSizePool.hpp` I wrapped the coalescing function with a conditional that checks if there is more than a single block in the pool.


mcfadden8 (approved):
I think this looks good.  Thanks for doing this!

davidbeckingsale (approved):

kab163 (approved):

PR #856, opened 2023-10-31
Use GNUInstallDirs to set installation directories

### LINKED ISSUE #855: Avoiding hardcoded install paths (e.g. `lib` vs `lib64`)

Currently, `Umpire` hardcodes the subpaths of install targets (e.g. `lib`). In a vacuum this may or may not be problematic, but when pulled in as a dependency, the inability to change these types of paths can be problematic (see e.g. https://github.com/ValeevGroup/tiledarray/issues/428, where the locally defined `LIBDIR` may clash with the install location of `Umpire` if not careful)

It would be desirable for `Umpire` to at least respect `CMAKE_INSTALL_LIBDIR` to allow for consistent install locations when pulled in as a dependency. Typically we set this via `GNUInstallDirs.cmake`,  but that's less of a concern if at least the behavior is overridable via the aforementioned variable. 

N/A

**Additional context**

N/A

davidbeckingsale:
Fixes #855


src/tpl/CMakeLists.txt
```diff
@@ -57,18 +57,17 @@ blt_convert_to_system_includes(TARGET umpire_tpl_json)
 target_include_directories(
   umpire_tpl_json
   INTERFACE
-  $<INSTALL_INTERFACE:include>)
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_FULL_INCLUDEDIR}>)
```
white238: Are you sure this is full?

davidbeckingsale: Maybe... :D

davidbeckingsale:
CC @wavefunction91 - does this change meet your needs?

kab163 (approved):
lgtm

PR #858, opened 2023-11-09
Making umpire_device a separate library 

kab163:
This would make the `umpire_device` library (containing device code, i.e. the device allocator that can be used in GPU kernels) a separate library.

Umpire already requires rdc when the device allocator feature is enabled. (By default it is off.) This will break it out into a separate lib so if a downstream user doesn't require that feature, they won't need RDC. 

We would still like to get input from users before merging.

Closes #827


white238 (dismissed):

MrBurmark (dismissed):
lgtm

davidbeckingsale (approved):

PR #863, opened 2024-01-11
Fix typo in HostSharedMemoryResourceImpl causing infinite loop

davidbeckingsale:


mcfadden8 (approved):
Thanks for finding/fixing this @davidbeckingsale!

PR #889, opened 2024-06-27
Use radiuss docker images

davidbeckingsale:
Trims down CI configurations

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,8 +44,8 @@ jobs:
         docker_target: gcc
       clang:
         docker_target: clang
-      cuda:
-        docker_target: cuda
+      # cuda:
+      #   docker_target: cuda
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -40,22 +40,14 @@ jobs:
...
-      clang11:
-        docker_target: clang11
-      clang12:
-        docker_target: clang12
-      sycl:
-        docker_target: sycl
+      gcc:
+        docker_target: gcc
+      clang:
+        docker_target: clang
```
adayton1, line 46: What about an intel build?

davidbeckingsale: Coming soon, need to fix the Intel image which currently fails to build :'(

kab163 (approved):

PR #897, opened 2024-07-23
Release/2024.07.0

kab163:


rhornung67 (approved):

adayton1:
Should there be a branch targeting main? Or does umpire follow a different release process than RAJA?

kab163:
> Should there be a branch targeting main? Or does umpire follow a different release process than RAJA?

I don't think so, no. I imagine Umpire's release process is different from RAJA's. We stopped using the main branch after v6.0.0

PR #900, opened 2024-07-29
provide finalize_io method

### LINKED ISSUE #616: ability shut down umpire's I/O cleanly

`umpire::initialize_io` grabs `std::cerr`'s rdbuf and flushes it when at the end of lifetime of the corresponding static object (`s_error_buffer`). This assumes that the buffer is still around. In our app we reset `std::cerr`'s rdbuf. Unfortunately there is no way to access `s_error_buffer`.

Ideally there should be a way to control the lifetime of ResourceManager. Or there should be a `finalize_io` function that can be called by the user.

Not messing with `std::cerr`'s buffer ... but I don't see anything in the standard forbidding this.

**Additional context**

n/a.

evaleev:
This is needed to clear references to buffers of standard streams in case `initialize_io` captures references to buffers of standard streams that are custom and have finite lifetime.

resolves https://github.com/LLNL/Umpire/issues/616

### DIFF
--- a/src/umpire/util/io.cpp
+++ b/src/umpire/util/io.cpp
@@ -52,10 +52,26 @@ std::ostream& error()
 
 namespace util {
 
+namespace detail {
+
+OutputBuffer& s_log_buffer_accessor()
+{
+  static OutputBuffer buffer;
+  return buffer;
+}
+
+OutputBuffer& s_error_buffer_accessor()
+{
+  static OutputBuffer buffer;
+  return buffer;
+}
+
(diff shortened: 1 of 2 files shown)


evaleev:
`Apply Style` seems to fail with PRs from forks: https://github.com/LLNL/Umpire/actions/runs/10146645731/job/28055190803?pr=900#step:2:589

possible solution (and relevant issue in any case): https://github.com/actions/checkout/issues/455#issuecomment-792228083

cc: @davidbeckingsale

PR #904, opened 2024-08-15
Fix pre-main crash in CUDA builds

adayton1:
* Avoid using regex in CUDA builds
* Separate out logic for CUDA and Hip

### DIFF
--- a/src/umpire/resource/MemoryResourceTypes.hpp
+++ b/src/umpire/resource/MemoryResourceTypes.hpp
@@ -8,7 +8,6 @@
 #define UMPIRE_MemoryResourceTypes_HPP
 
 #include <cstddef>
-#include <regex>
 #include <string>
 
 #include "umpire/config.hpp"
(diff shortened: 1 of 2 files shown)


src/umpire/resource/MemoryResourceTypes.hpp
```diff
@@ -106,22 +107,27 @@ inline MemoryResourceType string_to_resource(const std::string& resource)
...
+  int device_id{0};
+
+#if defined(UMPIRE_ENABLE_CUDA)
+  if (resource.find("::") != std::string::npos) {
+    device_id = std::stoi(resource.substr(resource.find("::") + 2));
+  } else {
+    // get the device bound to the current process
+    cudaGetDevice(&device_id);
+  }
+#elif defined(UMPIRE_ENABLE_HIP)
```
davidbeckingsale: CUDA and HIP are not the only device types, so I would just leave this as an `else` rather than `elif`

davidbeckingsale (changes requested):

src/umpire/resource/MemoryResourceTypes.hpp
adayton1: Done.

davidbeckingsale (approved):

adayton1:
@davidbeckingsale, the static analysis check seems to have been frozen for a couple days now. Does it need to be restarted?

robinson96 (approved):
LGTM

davidbeckingsale:
It doesn't look like any of the GitHub Actions have run, but not sure why

PR #908, opened 2024-09-06
Fix high water mark coalescing in memory pools

### LINKED ISSUE #906: Pathological reallocation in memory pools
**Describe the bug**

With certain allocation patterns and sizes QuickPool never coalesces a backing allocation to be large enough, leading to unnecessary reallocations.

**To Reproduce**

Compile the following file with anything from 2024.02.01 to current develop (older versions likely also affected, but have not tested):
```c++
#include <umpire/ResourceManager.hpp>
#include <umpire/strategy/QuickPool.hpp>

#include <iostream>
#include <cstddef>

int main() {
    auto alloc = umpire::ResourceManager::getInstance().getAllocator("DEVICE");
    auto pooled_alloc =
        umpire::ResourceManager::getInstance().makeAllocator<umpire::strategy::QuickPool>(
            "DEVICE_pool", alloc,
            16,  // initial pool size important only in the sense that allocations must be larger than the initial size, otherwise no additional allocations are required
            16,  // next size is not important to reproduce
            16); // alignment important: if all allocations are multiples of alignment all is good

    const std::size_t niter = 5;
    for (std::size_t iter = 0; iter < niter; ++iter) {
            std::cerr << "iteration " << iter << '\n';
            auto* p1 = pooled_alloc.allocate(1024 + 8);
            auto* p2 = pooled_alloc.allocate(1024 + 8); // need at least two allocations not multiples of alignment
            pooled_alloc.deallocate(p2);
            pooled_alloc.deallocate(p1);
    }
}
```

Just to print the allocations I ran this through gdb with the following gdb script:
```
start

break cudaMalloc
commands
continue
end

break cudaFree
commands
continue
end

continue
```
and then
```
gdb -batch -x gdbscript umpire_test
```
produces:
```
...
iteration 0
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
iteration 1
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
iteration 2
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
iteration 3
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
iteration 4
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 2, 0x0000fffff59fac64 in cudaMalloc () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
Thread 1 "miniapp_umpire_" hit Breakpoint 3, 0x0000fffff59fb2d0 in cudaFree () from /capstor/scratch/cscs/simbergm/src/DLA-Future/build/spack/src/libDLAF.so
...
```
Sizes are not visible above, but the first iteration first allocates twice, once for the initial buffer and once more to accommodate the second allocation. It then coalesces, freeing the two backing allocations above, and allocating one larger block. However, this block does not take overallocation into account, and is too small for the next iteration. The pool then always allocates for the second user-triggered allocation, frees both backing allocations, and attempts to coalesce. Repeat.

If one of the user-requested allocations is 16-byte aligned the reproducer correctly allocates only on the first iteration and then never again.

I suppose using umpire's logging should produce the same results, I'm just not familiar enough with it.

I used `DEVICE` only to make it easier to hook gdb into `cudaMalloc` and avoid printing potential other allocations. The same behaviour should happen with any backing memory.

I expect this affects DynamicPoolList as well.

**Expected behavior**

I expect that repeating the same allocation and deallocation pattern multiple times (no matter what initial size, next size, alignment, and user-requested allocation sizes are used), letting the pool empty between iterations to allow for coalescing (with the default coalescing strategy), would lead to allocations only during the first iteration and when coalescing after the first iteration. After that all allocations should fit in the pool without any further allocations.

**Compilers & Libraries (please complete the following information):**
 - Compiler & version: GCC 12.3.0
 - CUDA version (if applicable): CUDA 12.2.1 (not relevant to reproduce though)

**Additional context**

If I understand the pools correctly, the reason for this behaviour is that the actual size and the high water mark don't include the overallocation. They only sum up the sizes requested by the user, not the sizes actually allocated eventually, so there can be cases like the above where the pool never grows big enough to fit all required allocations.

msimberg:
Fixes #906.

This adds two methods, `getAlignedSize` and `getAlignedHighwaterMark` to `QuickPool` and `DynamicPoolList`. The `m_current_bytes`/`m_current_size` members have been renamed `m_aligned_bytes`.

`getAlignedHighwaterMark` is now used instead of `getHighWatermark` in the `hwm` coalescing heuristics.

### DIFF
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -83,6 +83,16 @@ std::size_t DynamicPoolList::getActualHighwaterMark() const noexcept
   return dpa.getActualHighwaterMark();
 }
 
+std::size_t DynamicPoolList::getAlignedSize() const noexcept
+{
+  return dpa.getAlignedSize();
+}
+
+std::size_t DynamicPoolList::getAlignedHighwaterMark() const noexcept
+{
+  return dpa.getAlignedHighwaterMark();
+}
+
 std::size_t DynamicPoolList::getReleasableSize() const noexcept
 {
   std::size_t SparseBlockSize = dpa.getReleasableSize();
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -186,7 +186,7 @@ PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::percent_releasable_hwm(i
     return [=](const strategy::DynamicPoolList& pool) {
       // Calculate threshold in bytes from the percentage
       const std::size_t threshold = static_cast<std::size_t>(f * pool.getActualSize());
-      return pool.getReleasableSize() >= threshold ? pool.getActualHighwaterMark() : 0;
+      return pool.getReleasableSize() > threshold ? pool.getActualHighwaterMark() : 0;
     };
   }
 }
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -143,7 +143,7 @@ PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::blocks_releasable(std::s
 PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::blocks_releasable_hwm(std::size_t nblocks)
 {
   return [=](const strategy::DynamicPoolList& pool) {
-    return pool.getReleasableBlocks() >= nblocks ? pool.getHighWatermark() : 0;
+    return pool.getReleasableBlocks() >= nblocks ? pool.getActualHighwaterMark() : 0;
   };
 }
 
@@ -178,14 +178,15 @@ PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::percent_releasable_hwm(i
   if (percentage == 0) {
     return [=](const DynamicPoolList& UMPIRE_UNUSED_ARG(pool)) { return 0; };
   } else if (percentage == 100) {
-    return
-        [=](const strategy::DynamicPoolList& pool) { return pool.getCurrentSize() == 0 ? pool.getHighWatermark() : 0; };
+    return [=](const strategy::DynamicPoolList& pool) {
+      return pool.getCurrentSize() == 0 ? pool.getActualHighwaterMark() : 0;
+    };
   } else {
     float f = (float)((float)percentage / (float)100.0);
     return [=](const strategy::DynamicPoolList& pool) {
       // Calculate threshold in bytes from the percentage
       const std::size_t threshold = static_cast<std::size_t>(f * pool.getActualSize());
-      return pool.getReleasableSize() >= threshold ? pool.getHighWatermark() : 0;
+      return pool.getReleasableSize() >= threshold ? pool.getActualHighwaterMark() : 0;
     };
   }
 }
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -64,13 +64,6 @@ std::size_t DynamicPoolList::getTotalBlocks() const noexcept
   return dpa.getTotalBlocks();
 }
 
-std::size_t DynamicPoolList::getCurrentSize() const noexcept
-{
-  std::size_t CurrentSize = dpa.getCurrentSize();
-  UMPIRE_LOG(Debug, "() returning " << CurrentSize);
-  return CurrentSize;
-}
-
 std::size_t DynamicPoolList::getActualSize() const noexcept
 {
   std::size_t ActualSize = dpa.getActualSize();
--- a/src/umpire/strategy/DynamicPoolList.cpp
+++ b/src/umpire/strategy/DynamicPoolList.cpp
@@ -153,7 +153,7 @@ PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::blocks_releasable(std::s
 PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::blocks_releasable_hwm(std::size_t nblocks)
 {
   return [=](const strategy::DynamicPoolList& pool) {
-    return pool.getReleasableBlocks() >= nblocks ? pool.getActualHighwaterMark() : 0;
+    return pool.getReleasableBlocks() >= nblocks ? pool.getAlignedHighwaterMark() : 0;
   };
 }
 
@@ -189,14 +189,14 @@ PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::percent_releasable_hwm(i
     return [=](const DynamicPoolList& UMPIRE_UNUSED_ARG(pool)) { return 0; };
   } else if (percentage == 100) {
     return [=](const strategy::DynamicPoolList& pool) {
-      return pool.getCurrentSize() == 0 ? pool.getActualHighwaterMark() : 0;
+      return pool.getCurrentSize() == 0 ? pool.getAlignedHighwaterMark() : 0;
     };
   } else {
     float f = (float)((float)percentage / (float)100.0);
     return [=](const strategy::DynamicPoolList& pool) {
       // Calculate threshold in bytes from the percentage
       const std::size_t threshold = static_cast<std::size_t>(f * pool.getActualSize());
-      return pool.getReleasableSize() > threshold ? pool.getActualHighwaterMark() : 0;
+      return pool.getReleasableSize() >= threshold ? pool.getAlignedHighwaterMark() : 0;
     };
   }
 }
--- a/src/umpire/strategy/DynamicSizePool.hpp
+++ b/src/umpire/strategy/DynamicSizePool.hpp
@@ -372,13 +372,6 @@ class DynamicSizePool : private umpire::strategy::mixins::AlignedAllocation {
     return m_actual_bytes;
   }
 
-  // TODO: Keep old behaviour with m_aligned_bytes or remove and use definition
-  // from AllocationStrategy which returns m_current_size?
-  std::size_t getCurrentSize() const
-  {
-    return m_aligned_bytes;
-  }
-
   std::size_t getActualHighwaterMark() const noexcept
   {
     return m_actual_highwatermark;
--- a/src/umpire/strategy/DynamicSizePool.hpp
+++ b/src/umpire/strategy/DynamicSizePool.hpp
@@ -40,8 +40,9 @@ class DynamicSizePool : private umpire::strategy::mixins::AlignedAllocation {
   struct Block *freeBlocks{nullptr};
 
   // Total size allocated (bytes)
+  std::size_t m_aligned_bytes{0};
+  std::size_t m_aligned_highwatermark{0};
   std::size_t m_actual_bytes{0};
-  std::size_t m_current_size{0};
   std::size_t m_actual_highwatermark{0};
 
   // Minimum size of initial allocation
@@ -320,7 +321,10 @@ class DynamicSizePool : private umpire::strategy::mixins::AlignedAllocation {
     best->next = usedBlocks;
     usedBlocks = best;
 
-    m_current_size += rounded_bytes;
+    m_aligned_bytes += rounded_bytes;
+    if (m_aligned_bytes > m_aligned_highwatermark) {
+      m_aligned_highwatermark = m_aligned_bytes;
+    }
     UMPIRE_UNPOISON_MEMORY_REGION(m_allocator, usedBlocks->data, bytes);
 
     // Return the new pointer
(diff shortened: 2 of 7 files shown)


src/umpire/strategy/DynamicPoolList.cpp
```diff
@@ -178,14 +178,15 @@ PoolCoalesceHeuristic<DynamicPoolList> DynamicPoolList::percent_releasable_hwm(i
...
+    return [=](const strategy::DynamicPoolList& pool) {
+      return pool.getCurrentSize() == 0 ? pool.getActualHighwaterMark() : 0;
+    };
   } else {
     float f = (float)((float)percentage / (float)100.0);
     return [=](const strategy::DynamicPoolList& pool) {
       // Calculate threshold in bytes from the percentage
       const std::size_t threshold = static_cast<std::size_t>(f * pool.getActualSize());
-      return pool.getReleasableSize() >= threshold ? pool.getHighWatermark() : 0;
+      return pool.getReleasableSize() > threshold ? pool.getActualHighwaterMark() : 0;
```
msimberg: Here and in `QuickPool.cpp`, I had to change the comparison to `>` from `>=` to avoid infinite recursion. With `getActualHighwaterMark` https://github.com/LLNL/Umpire/blob/bf2df8d46f6e8918b7402a45b0f95127c3121478/tests/unit/strategy/pool_heuristics_tests.cpp#L116 ends up exhausting the stack when coalescing because `getReleasableSize() == threshold` after coalescing. Since the last operation of coalescing is deallocating (https://github.com/LLNL/Umpire/blob/bf2df8d46f6e8918b7402a45b0f95127c3121478/src/umpire/strategy/QuickPool.cpp#L335), which then checks for coalescing again, you end up with infinite recursion.

If the original behaviour should be kept then I suppose allocation and creation of new blocks could be separated, such that `do_coalesce` could release memory, allocate a new block and return, instead of going via `allocate` and `deallocate` to create a new block.

What do you think is best here?

tests/integration/primary_pool_tests.cpp
```diff
@@ -358,7 +358,7 @@ TYPED_TEST(PrimaryPoolTest, coalesce)
   ASSERT_EQ(pool->getBlocksInPool(), 1);
 
   ASSERT_EQ(this->m_allocator->getCurrentSize(), 0);
-  ASSERT_LT(pool->getActualSize(), old_actual_size);
+  ASSERT_LE(pool->getActualSize(), old_actual_size);
```
msimberg: `getActualSize()` and `old_actual_size` now seem to always be equal in the test. I can't judge if this is a meaningful semantic change or not.

tests/integration/primary_pool_tests.cpp
```diff
@@ -579,7 +579,7 @@ TYPED_TEST(PrimaryPoolTest, heuristic_75_percent_hwm)
   ASSERT_NO_THROW({ alloc.deallocate(a[2]); }); // 50% releasable
   ASSERT_EQ(pool->getBlocksInPool(), 4);
   ASSERT_NO_THROW({ alloc.deallocate(a[1]); }); // 75% releasable
-  ASSERT_EQ(pool->getBlocksInPool(), 2);        // Collapse happened
+  ASSERT_EQ(pool->getBlocksInPool(), 4);
```
msimberg: This happens because of the change from `>=` to `>` in the coalescing heuristic. Instead of coalescing at exactly 75% releasable it'll only coalesce when strictly more than 75% memory is free.

src/umpire/strategy/DynamicPoolList.cpp
msimberg: Perhaps https://github.com/LLNL/Umpire/blob/bf2df8d46f6e8918b7402a45b0f95127c3121478/src/umpire/strategy/QuickPool.cpp#L325 should be `getReleasableBlocks() > 1` instead? If there's only one or zero blocks there's no coalescing that can be done.

msimberg:
Thinking about this a bit more, this isn't quite right either. There's a missing counter, which is the high water mark of "current bytes". In QuickPool, `m_current_bytes` is what is used for `getCurrentSize` which correctly accounts for overallocation. `m_current_size` from `AllocationStrategy` does not take into account overallocation. Should this rather introduce a new counter in `QuickPool` which does the high water mark for `m_current_bytes` instead of `m_current_size`?

davidbeckingsale:
> Thinking about this a bit more, this isn't quite right either. There's a missing counter, which is the high water mark of "current bytes". In QuickPool, `m_current_bytes` is what is used for `getCurrentSize` which correctly accounts for overallocation. `m_current_size` from `AllocationStrategy` does not take into account overallocation. Should this rather introduce a new counter in `QuickPool` which does the high water mark for `m_current_bytes` instead of `m_current_size`?

I had started looking at this, and I think that's the right approach. Let me know if you would like to tackle this. Thanks!

msimberg:
> I had started looking at this, and I think that's the right approach. Let me know if you would like to tackle this. Thanks!

Good to hear, thanks! If you already have something started and you have specific thoughts on how to do it, by all means go for it. You do know the codebase better than I do after all! I'm happy to continue in this direction if it helps though, I probably just need a bit more time and support, so if you're happy to wait for me to get this into a good state I'll keep at it.

src/umpire/strategy/DynamicSizePool.hpp
```diff
@@ -368,16 +372,28 @@ class DynamicSizePool : private umpire::strategy::mixins::AlignedAllocation {
     return m_actual_bytes;
   }
 
+  // TODO: Keep old behaviour with m_aligned_bytes or remove and use definition
+  // from AllocationStrategy which returns m_current_size?
```
msimberg: This still needs addressing.

src/umpire/strategy/QuickPool.cpp
```diff
@@ -262,9 +265,11 @@ std::size_t QuickPool::getActualSize() const noexcept
   return m_actual_bytes;
 }
 
+// TODO: Keep old behaviour with m_aligned_bytes or remove and use definition
+// from AllocationStrategy which returns m_current_size?
```
msimberg: This also needs addressing, same as https://github.com/LLNL/Umpire/pull/908/files#r1749860382.

tests/integration/primary_pool_tests.cpp
msimberg: This is no longer necessary after the new changes and has been reverted.

tests/integration/primary_pool_tests.cpp
msimberg: This is also reverted with the latest changes.

src/umpire/strategy/DynamicPoolList.cpp
msimberg: I've reverted the `>=` to `>` change with the latest additions. It's no longer an issue when using `getAlignedHighwaterMark` instead of `getActualHighwaterMark`.

The second comment above is still relevant, but is not a must for this PR.

msimberg:
@davidbeckingsale I've pushed some updates hopefully in the direction that you had in mind.

One open question is if the non-`hwm` heuristics should use `getAlignedSize` instead of `getActualSize`? It's not really a affected by #906 since `getActualSize` is is definitely big enough for the user-facing allocations, but perhaps it should change for consistency? I'm not sure...

src/umpire/strategy/DynamicSizePool.hpp
davidbeckingsale: I think that this overload can actually be removed, and the getCurrentSize provided by AllocationStrategy will return the current size (current total of bytes requested by user, un-aligned)

src/umpire/strategy/QuickPool.cpp
davidbeckingsale: Same comment as above.

davidbeckingsale (changes requested):

davidbeckingsale:
> @davidbeckingsale I've pushed some updates hopefully in the direction that you had in mind.
> 
> One open question is if the non-`hwm` heuristics should use `getAlignedSize` instead of `getActualSize`? It's not really a affected by #906 since `getActualSize` is is definitely big enough for the user-facing allocations, but perhaps it should change for consistency? I'm not sure...

Looks good. I think the non-hwm can stick with actualSize since that is already including the alignment. (actual size is the _actual_ amount of bytes allocated by the pool).

src/umpire/strategy/DynamicSizePool.hpp
msimberg: Removed in https://github.com/LLNL/Umpire/pull/908/commits/58a685af3f7ab6e3002995cf83e6b5c25e26af11.

src/umpire/strategy/QuickPool.cpp
msimberg: Removed in https://github.com/LLNL/Umpire/pull/908/commits/58a685af3f7ab6e3002995cf83e6b5c25e26af11.

src/umpire/strategy/DynamicSizePool.hpp
msimberg: Note that this test was updated as a result of this change: https://github.com/LLNL/Umpire/pull/908/commits/9c28b1ccef8f12414135082225c4be2758c81275.

davidbeckingsale:
Merged in #917

PR #910, opened 2024-09-10
Adding IPC Shared Memory docs

kab163:


kab163:
Anything else to clarify or add here?

https://umpire.readthedocs.io/en/kab163-task-add-ipc-shared-mem-docs/sphinx/cookbook/shared_memory.html

docs/sphinx/cookbook/shared_memory.rst
```diff
@@ -0,0 +1,73 @@
...
+Important Notes About Shared Memory
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Because we are dealing with shared memory there are a few unique characteristics of the Shared Memory allocators
+which set it apart from other Umpire allocators.
+
+1. Once you allocate shared memory, that block of memory is fixed. If you need a bigger size, you will have to create a new one.
+2. If you want to see how much memory is available for a shared memory allocator, use the ``getActualSize()`` function.
+3. File descriptors are used for the shared memory. These files will be under ``/dev/shm``.
+
```
mcfadden8, line 61: Thanks @kab163  for adding documentation for this.  I'm glad that interest seems to be building and documentation is needed.  

You are correct in your statement made earlier that Umpire does not need to have MPI enabled in order to provide users with IPC Shared Memory.  HOWEVER, if they wish to associated shared memory with MPI communicators, I believe that Umpire will need to also have MPI set to On when it is compiled.  I could be wrong here, but this is how I remember it.  @davidbeckingsale might be able to shed further light or correct me if I am wrong.

davidbeckingsale (approved):

PR #931, opened 2025-02-03
Add cleanup_cached_communicators method

kab163:
Adding function to cleanup cached communicators to avoid memory leak.
If users call `umpire::get_communicator_for_allocator(...)` then it currently causes a memory leak. The addition of this `umpire::cleanup_cached_communicators()` frees that memory to avoid memory leaks.

Thanks to Burl Hall for catching this bug!

### DIFF
--- a/src/umpire/Umpire.cpp
+++ b/src/umpire/Umpire.cpp
@@ -256,16 +256,17 @@ void* find_pointer_from_name(Allocator allocator, const std::string& name)
 
 #if defined(UMPIRE_ENABLE_MPI)
 namespace {
-static std::map<int, MPI_Comm> cached_communicators{};
-
 std::map<int, MPI_Comm>& get_cached_communicators()
 {
+  static std::map<int, MPI_Comm> cached_communicators{};
   return cached_communicators;
 }
 } // namespace
 
 MPI_Comm get_communicator_for_allocator(Allocator a, MPI_Comm comm)
 {
+  std::map<int, MPI_Comm>& cached_communicators = get_cached_communicators();
+
   MPI_Comm c;
   auto scope = a.getAllocationStrategy()->getTraits().scope;
   int id = a.getId();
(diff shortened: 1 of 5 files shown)


src/umpire/Umpire.cpp
```diff
@@ -277,6 +284,15 @@ MPI_Comm get_communicator_for_allocator(Allocator a, MPI_Comm comm)
...
   return c;
 }
+
+void cleanup_cached_communicators()
+{
+  std::map<int, MPI_Comm>& comm = get_cached_communicators();
+
+  for (auto c : comm) {
+    MPI_Comm_free(&c.second);
+  }
```
adayton1, line 295: I think you may want to call clear on this map after freeing all the communicators.

src/umpire/Umpire.cpp
```diff
@@ -255,10 +255,17 @@ void* find_pointer_from_name(Allocator allocator, const std::string& name)
 }
 
 #if defined(UMPIRE_ENABLE_MPI)
-MPI_Comm get_communicator_for_allocator(Allocator a, MPI_Comm comm)
+namespace {
+static std::map<int, MPI_Comm> cached_communicators{};
```
adayton1: I would move this inside the get_cached_communicators() function and have get_communicator_for_allocator call get_cached_communicators. That way it is accessed consistently.

kab163:
See generated updated docs: https://umpire.readthedocs.io/en/bugfix-kab163-fix-mpi-leak/sphinx/cookbook/shared_memory.html

src/umpire/Umpire.cpp
kab163: Ah yes, good idea

src/umpire/Umpire.cpp
kab163: Done!

adayton1 (approved):

davidbeckingsale (approved):

PR #937, opened 2025-03-10
Documenting Windows Build Minimum Requirements

kab163:


CMakeLists.txt
```diff
@@ -134,6 +134,16 @@ include(cmake/SetupCMakeBasics.cmake)
 include(cmake/SetupCompilerFlags.cmake)
 include(cmake/SetupUmpireThirdParty.cmake)
 
+if (WIN32)
+  if ((NOT UMPIRE_ENABLE_FILESYSTEM) OR (NOT "${CMAKE_CXX_STANDARD}" STREQUAL "17"))
```
cyrush: 👍

kab163:
With the cmake fatal error message set, when I turn off the filesystem:
```
-- Setting C standard to 99
-- Checking for std::filesystem
-- std::filesystem NOT found, using POSIX
CMake Error at CMakeLists.txt:139 (message):
      Windows builds require the FileSystem to be enabled     and the C/C++ standard to be set to 17 or higher.     Please re-configure with UMPIRE_ENABLE_FILESYSTEM=ON,     with BLT_CXX_STD=c++17 and CMAKE_CXX_STANDARD=17. 


-- Configuring incomplete, errors occurred!
Error: Process completed with exit code 1.
```
So the cmake stuff seems to work.

CMakeLists.txt
davidbeckingsale: @kab163 this should check that the standard is at least c++17

davidbeckingsale (dismissed):

CMakeLists.txt
```diff
@@ -134,6 +134,16 @@ include(cmake/SetupCMakeBasics.cmake)
 include(cmake/SetupCompilerFlags.cmake)
 include(cmake/SetupUmpireThirdParty.cmake)
 
+if (WIN32)
+  if ((NOT UMPIRE_ENABLE_FILESYSTEM) OR (CMAKE_CXX_STANDARD LESS 17))
+    message(FATAL_ERROR "\
+    Windows builds require the FileSystem to be enabled \
+    and the C/C++ standard to be set to 17 or higher. \
+    Please re-configure with UMPIRE_ENABLE_FILESYSTEM=ON, \
+    with BLT_CXX_STD=c++17 and CMAKE_CXX_STANDARD=17 or greater.")
```
```suggestion
    with BLT_CXX_STD=c++17 or greater.")
```

davidbeckingsale: Check is good but this message is clearer, since use doesn't have to set both variables.

kab163: si senor

davidbeckingsale (approved):

adayton1 (approved):

PR #939, opened 2025-03-17
spack v1.0 support: `%foo +bar` -> `+bar %foo`

haampie:
Spack v1.0 will parse `pkg %gcc +foo` as "pkg with direct dependency gcc
with variant foo enabled" instead of "pkg with variant foo enabled and compiler
gcc". Reorder as `pkg +foo %gcc` to be compatible with Spack v0.x and
v1.0.


adrienbernede (approved):
@davidbeckingsale I confirm this is OK: https://lc.llnl.gov/gitlab/bernede1/umpire-no-mirror/-/pipelines/520878.

PR #943, opened 2025-03-18
Fix HIP constant memory

davidbeckingsale:
Local branch from #920 so we can run all CI.

### DIFF
--- a/.gitlab/jobs/tioga.yml
+++ b/.gitlab/jobs/tioga.yml
@@ -38,3 +38,7 @@ rocmcc_6_3_1_hip_openmp_device_alloc:
     SPEC: "~shared +fortran +openmp +rocm +device_alloc tests=basic amdgpu_target=gfx90a %rocmcc@=6.3.1 ^hip@6.3.1"
   extends: .job_on_tioga
 
+rocmcc_6_3_0_hip_deviceconst:
+  variables:
+    SPEC: "~shared +fortran +deviceconst +rocm tests=basic amdgpu_target=gfx90a %rocmcc@=6.3.0 ^hip@6.3.0"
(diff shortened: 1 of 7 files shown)


gzagaris:
Hey @davidbeckingsale -- I was about to follow up with you on this and noticed this PR. 

When you get a chance let me know if you have any further information on this issue or if there is anything that I can do on my end to help you with this.

I also noticed that in ece2248 you enabled RDC. Is that required? We have observed about [5% - 20%] overhead with RDC (with CUDA) and I have switched to not using it at all.

davidbeckingsale:
The build system piece seems good, just working on a linking issue - I think it might be todo with the fact that it's looking for virtual functions that were in a base class not compiled as hip, which is why I was trying RDC. Looking into it...

gzagaris:
> The build system piece seems good, just working on a linking issue - I think it might be todo with the fact that it's looking for virtual functions that were in a base class not compiled as hip, which is why I was trying RDC. Looking into it...

Thanks @davidbeckingsale!

I'd be a little surprised if RDC was the issue. The virtual functions in the base class, as well as the HIP runtime API calls, are all host functions. To the best of my knowledge RDC is primarily relevant for device (or host-device) functions that are called within a kernel but defined in a separate compilation unit -- which doesn't seem to be the case here. This wasn't necessary for CUDA, at least. But,  of course, it's build-systems and compilers, so ... who knows 😄 

The only device asset is the buffer:
```
__constant__ static char s_umpire_internal_device_constant_memory[MAX_CONST_MEM_SIZE];
```

But, that's not referenced directly anywhere outside the cpp file that defines it. I'm confused. I'll stop rambling. 😕

kab163 (approved):

PR #958, opened 2025-04-25
Add missing strategies to Fortran API

davidbeckingsale:


kab163 (commented):
Do all these `new` calls have a corresponding `delete`? Or is that handled some other way?

davidbeckingsale:
> Do all these `new` calls have a corresponding `delete`? Or is that handled some other way?

Handled by Shroud

kab163 (approved):

PR #967, opened 2025-05-21
Allow both shared memory implementations

davidbeckingsale:


examples/mpi3_shared_memory.cpp
```diff
@@ -24,7 +24,7 @@ int main(int argc, char** argv)
   traits.scope = umpire::MemoryResourceTraits::shared_scope::node;
 
   // Create allocator using MPI3 shared memory
-  auto mpi3_shm_allocator = rm.makeResource("SHARED::mpi3_alloc", traits);
```
adrienbernede: I found this change intriguing because the mpi3 CI jobs sometimes fails.
Is it changing the name of the mpi communicator, the mpi allocation, or an Umpire specific memory allocation ?

kab163: Just the name of the Umpire allocator. We had a user who wanted to be able to give the shared memory allocator the name "SHARED". @davidbeckingsale might want to chime in on Monday too.

kab163:
https://umpire.readthedocs.io/en/task-allow-both-shared-mem/sphinx/cookbook/shared_memory_allocators.html

kab163 (dismissed):

tests/unit/resource/shared_memory_resource_tests.cpp
```diff
@@ -51,12 +51,12 @@ class SharedMemoryTest : public ::testing::Test {
...
     if (shmem_resource == nullptr) {
       auto& rm = umpire::ResourceManager::getInstance();
-      auto traits{umpire::get_default_resource_traits("SHARED")};
+      auto traits{umpire::get_default_resource_traits("SHARED::POSIX")};
       ASSERT_EQ(traits.scope, umpire::MemoryResourceTraits::shared_scope::node);
       ASSERT_EQ(traits.resource, umpire::MemoryResourceTraits::resource_type::shared);
 
       traits.size = m_segment_size;
-      ASSERT_NO_THROW(allocator = rm.makeResource("SHARED::node_allocator", traits););
+      ASSERT_NO_THROW(allocator = rm.makeResource("SHARED::POSIX::node_allocator", traits););
```
kab163, line 60: This is actually REQUIRED for this to work. For the strategy tests to pass, on ruby, with IPC and MPI3 enabled, it kept trying to use the MPI3 shared memory unless I explicitly asked for `rm.makeResource("SHARED::POSIX::some_name", ....` - is that expected?

CMakeLists.txt
```diff
@@ -88,18 +88,39 @@ cmake_dependent_option( UMPIRE_ENABLE_OPENMP "Build Umpire with OpenMP" On
...
+endif()
+
+if(NOT DEFINED UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE)
+  if(UMPIRE_ENABLE_IPC_SHARED_MEMORY AND NOT UMPIRE_ENABLE_MPI3_SHARED_MEMORY)
+    set(UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE "POSIX" CACHE STRING "Default implementation for SHARED memory resource (POSIX)")
+  elseif(UMPIRE_ENABLE_MPI3_SHARED_MEMORY AND NOT UMPIRE_ENABLE_IPC_SHARED_MEMORY)
+    set(UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE "MPI3" CACHE STRING "Default implementation for SHARED memory resource (MPI3)")
+  else()
+    # Either both or neither enabled - default to POSIX
+    set(UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE "MPI3" CACHE STRING "Default implementation for SHARED memory resource (MPI3)")
```
kab163: When this wasn't the default, the allocator in the mpi3 shared memory tests would call the "allocate" function that belonged to IPC shared memory, causing a failure because no named was passed in...

kab163 (commented):
If it really is required to have "SHARED::POSIX::some_name" when creating the allocator, we will need to add that to the docs...

kab163 (dismissed):

CMakeLists.txt
kab163: I have added a comment there to make note that the default can't be changed

tests/unit/resource/shared_memory_resource_tests.cpp
kab163: Comments added - I assume that is expected since I haven't heard otherwise...

CMakeLists.txt
adayton1: I suspect there is a problem with the logic here or in the code if it can't be changed.

docs/sphinx/cookbook/shared_memory_allocators.rst
```diff
@@ -74,14 +74,22 @@ uses the MPI3 API for its Shared Memory mechanisms that allow processes to commu
 To use Umpire's MPI3 Shared Memory allocators, the ``UMPIRE_ENABLE_MPI3_SHARED_MEMORY`` flag 
 should be set to ``On``. Note that if you are using MPI3 Shared Memory, then MPI must be enabled.
 
-See the bottom of this page for a full example of how to use Shared Memory Allocators with Umpire.
+See the bottom of this page for a full example of how to use MPI3 Shared Memory Allocators with Umpire.
 
-Full IPC Shared Memory Recipe
------------------------------
+Ful Shared Memory Examples
```
adayton1: Spelling: Full

docs/sphinx/cookbook/shared_memory_allocators.rst
```diff
@@ -74,14 +74,22 @@ uses the MPI3 API for its Shared Memory mechanisms that allow processes to commu
...
+--------------------------
 
-The following example shows how to create, use, and destruct the IPC Shared Memory Allocator whether using MPI or not. For an example of using
-MPI3 Shared Memory Allocators, users could simply update the included header file. No other code changes are necessary assuming the input
-parameter specifies MPI for the following example code.
+This section shows two full code examples, one for IPC Shared Memory and one for MPI3.
```
adayton1: I would say MPI3 Shared Memory just for full clarity.

docs/sphinx/cookbook/shared_memory_allocators.rst
```diff
@@ -74,14 +74,22 @@ uses the MPI3 API for its Shared Memory mechanisms that allow processes to commu
...
+
+The following example shows how to create, use, and destruct the IPC Shared Memory Allocator whether using MPI or not. Note that this
```
adayton1: I am a bit confused by the statement "whether using MPI or not". That could probably be left off, or made into a separate note saying that IPC shared memory can be used even when MPI and/or MPI3 shared memory are enabled.

examples/mpi3_shared_memory.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+  // Use MPI3 shared memory resource
+  auto traits = umpire::get_default_resource_traits("SHARED::MPI3");
+  traits.size = 1 * 1024 * 1024; // 1 MB
+
+  // Node scope is required for mpi3 shared memory
+  traits.scope = umpire::MemoryResourceTraits::shared_scope::node;
+
+  // Create allocator using MPI3 shared memory
+  auto mpi3_shm_allocator = rm.makeResource("SHARED", traits);
```
adayton1: This example relies on the default shared memory type being MPI3.

tests/unit/resource/shared_memory_resource_tests.cpp
adayton1, line 1: Should this test be renamed to ipc_shared_memory_resource_tests.cpp?

src/umpire/resource/HostSharedMemoryResourceFactory.cpp
adayton1, line 1: Should this file be renamed to HostIpcSharedMemoryResourceFactory.cpp?

src/umpire/util/shared_memory_helper.hpp
```diff
@@ -0,0 +1,39 @@
...
+  const std::string prefix = "SHARED::" + resource_type;
+
+  // Check if name starts with "SHARED::" + resource_type
+  if (name.find(prefix) == 0) {
+    return true;
+  }
+
+  // Check if name starts with "SHARED::" or "SHARED" AND that this resource_type is the default
+  if ((name.find("SHARED::") == 0) || (name == "SHARED")) {
+#ifdef UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE
```
adayton1: I would include "umpire/config.hpp" in this file since that includes the definition for `UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE`. Otherwise you might get some weird behavior depending on the order of when this file is included.

src/umpire/util/shared_memory_helper.hpp
```diff
@@ -0,0 +1,39 @@
...
+    return std::string(UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE) == resource_type;
```
adayton1: Would it make sense to use the variable `umpire::default_shared_memory_resource` defined in config.hpp?

examples/mpi3_shared_memory.cpp
adayton1: Should it be SHARED::MPI3 (with a comment that it can be SHARED if the default shared memory type is MPI3)?

tests/unit/resource/mpi3_shared_memory_resource_tests.cpp
adayton1, line 1: I would change "SHARED" to "SHARED::MPI3" in this file as well.

adayton1: Or at least guard with the macro that specifies the default shared memory type.

tests/unit/resource/shared_memory_resource_tests.cpp
kab163: No, because it can work for either IPC or MPI3 depending on what you use when you call `makeResource` - no other code change is needed. Of course we assume if using MPI3, then MPI is enabled and the example will take the MPI route.

adayton1:
I think you are going to need to update the MemoryResourceType enum in MemoryResourceTypes.hpp. Otherwise, "SHARED" could map to the wrong implementation.

examples/mpi3_shared_memory.cpp
kab163: Yes, I will make the change. We had a user who wanted to be able to just specify "SHARED" and have it do the right thing... but I like your suggestion

tests/unit/resource/shared_memory_resource_tests.cpp
adayton1: I'm confused. It looks like it is now explicitly IPC because of the "SHARED::POSIX".

docs/sphinx/cookbook/shared_memory_allocators.rst
kab163: Made a note for this.

examples/mpi3_shared_memory.cpp
kab163: I think it is really the line that sets up the traits, `auto traits = umpire::get_default_resource_traits("SHARED::MPI3");` that matters most. The name, whether IPC or MPI3, just needs "SHARED" or "SHARED::" prefix, but the traits will need to be more specific... I think

CMakeLists.txt
```diff
@@ -88,18 +88,39 @@ cmake_dependent_option( UMPIRE_ENABLE_OPENMP "Build Umpire with OpenMP" On
...
+  endif()
+endif()
+
+if(NOT DEFINED UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE)
+  if(UMPIRE_ENABLE_IPC_SHARED_MEMORY AND NOT UMPIRE_ENABLE_MPI3_SHARED_MEMORY)
+    set(UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE "POSIX" CACHE STRING "Default implementation for SHARED memory resource (POSIX)")
+  elseif(UMPIRE_ENABLE_MPI3_SHARED_MEMORY AND NOT UMPIRE_ENABLE_IPC_SHARED_MEMORY)
+    set(UMPIRE_DEFAULT_SHARED_MEMORY_RESOURCE "MPI3" CACHE STRING "Default implementation for SHARED memory resource (MPI3)")
+  else()
+    # Either both or neither enabled. NOTE: MUST default to MPI3, otherwise there is confusion on which "allocate" func to call
```
adayton1: What happens if neither are enabled and someone tries to make a SHARED resource?

examples/mpi3_shared_memory.cpp
kab163: The traits could also just have "SHARED" if MPI3 is the default

src/umpire/resource/HostSharedMemoryResourceFactory.cpp
kab163: @davidbeckingsale ? I think that makes sense

CMakeLists.txt
kab163: when they call `rm.makeResource` with "SHARED", it will give a runtime error that "SHARED" can't be found. I have also added messages in the cmake to make it clear which kind of shared memory is enabled/disabled

tests/unit/resource/shared_memory_resource_tests.cpp
kab163: yea, so I guess if it has `SHARED_POSIX` it could do IPC, if it has `SHARED_MPI3` it could do MPI3, but the same example can be used by either one.

kab163: No, I am looking at the wrong file, sorry. You're right, we should rename it (I thought this was referring to the cookbook example.... sorry)

docs/sphinx/cookbook/shared_memory_allocators.rst
```diff
@@ -74,14 +74,61 @@ uses the MPI3 API for its Shared Memory mechanisms that allow processes to commu
 To use Umpire's MPI3 Shared Memory allocators, the ``UMPIRE_ENABLE_MPI3_SHARED_MEMORY`` flag 
 should be set to ``On``. Note that if you are using MPI3 Shared Memory, then MPI must be enabled.
 
-See the bottom of this page for a full example of how to use Shared Memory Allocators with Umpire.
+To create an allocator with the MPI3 Shared Memeory resource, you can do the following:
```
adayton1: Typo: Memeory

Occurs in a couple places.

docs/sphinx/cookbook/shared_memory_allocators.rst
```diff
@@ -74,14 +74,61 @@ uses the MPI3 API for its Shared Memory mechanisms that allow processes to commu
...
+To create these Shared Memeory allocators, you can do the following:
+
+.. code-block:: cpp
+
+   auto mpi3_traits{umpire::get_default_resource_traits("SHARED::MPI3")};
+   // or
+   auto ipc_traits{umpire::get_default_resource_traits("SHARED::POSIX")};
+
+   // then create an allocator:
+   auto mpi3_node_allocator{rm.makeResource("SHARED::mpi3_alloc", traits)};
```
adayton1, line 100: If IPC is the default, would this work? Or do you need "SHARED::MPI3::<some allocator name>"?

docs/sphinx/cookbook/shared_memory_allocators.rst
```diff
@@ -74,14 +74,61 @@ uses the MPI3 API for its Shared Memory mechanisms that allow processes to commu
...
-parameter specifies MPI for the following example code.
+   // and allocate with
+   mpi3_node_allocator.allocate(1024 * sizeof(double));
+   // or
+   ipc_node_allocator.allocate("my_SHARED_alloc", 1024 * sizeof(double));
+
+.. note::
+   It is best practice to use the full name, "SHARED::MPI3" or "SHARED::POSIX", when
+   setting up the `traits` for a shared memory allocator. However, when both IPC and MPI3
+   resources are enabled, using "SHARED" will default to the MPI3 memory resource. Additionally,
```
adayton1, line 112: Maybe mention you can change the default via a CMake option?

adayton1 (dismissed):

CMakeLists.txt
```diff
@@ -88,18 +88,24 @@ cmake_dependent_option( UMPIRE_ENABLE_OPENMP "Build Umpire with OpenMP" On
...
-# NOTE: The following option depends upon UMPIRE_ENABLE_MPI being defined first
-#
-cmake_dependent_option( UMPIRE_ENABLE_IPC_SHARED_MEMORY "Enable Host Shared Memory Resource" ${UMPIRE_ENABLE_MPI}
-                       "NOT WIN32;NOT APPLE" Off )
 
+########
+# Shared Memory Options
+########
+
+cmake_dependent_option( UMPIRE_ENABLE_IPC_SHARED_MEMORY "Enable Host Shared Memory Resource" ${UMPIRE_ENABLE_MPI}
```
kab163:
```suggestion
cmake_dependent_option( UMPIRE_ENABLE_IPC_SHARED_MEMORY "Enable Host IPC Shared Memory Resource" ${UMPIRE_ENABLE_MPI}
```

PR #977, opened 2025-07-08
Add HIP support to AllocationAdvisor

### LINKED ISSUE #973: AllocationAdvisor support for HIP

on AMD GPU hardware, giving memory advice to allocations can be crucial for performance. For unified memory, unless the allocation is declared coarse-grained, there is a significant penalty (and, afaik, cache usage is severely limited).
https://rocm.docs.amd.com/en/docs-6.2.0/conceptual/gpu-memory.html#coherence

The AllocationAdvisor in its current form, howerer, does not support HIP allocations:
https://github.com/LLNL/Umpire/blob/develop/src/umpire/strategy/AllocationAdvisor.cpp

Adding support for an unified memory allocator with an AllocationAdvisor on HIP/ROCm.

Users can perform those memory advices manually afterwards, but there's a performance penalty and it needs to be re-done after each re-allocation.

**Additional context**

Already available for CUDA, so should be quite straightforward to implement?

davidbeckingsale:
Fixes #973

### DIFF
--- a/src/umpire/strategy/AllocationAdvisor.hpp
+++ b/src/umpire/strategy/AllocationAdvisor.hpp
@@ -26,6 +26,7 @@ namespace strategy {
  * - op::CudaAdviseAccessedByOperation
  * - op::CudaAdvisePreferredLocationOperation
  * - op::CudaAdviseReadMostlyOperation
+ * - op::HipAdviseOperation
  *
  * Using this AllocationStrategy when combined with a pool like DynamicPoolList is
  * a good way to mitigate the overhead of applying the memory advice.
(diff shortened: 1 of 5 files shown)


src/umpire/strategy/AllocationAdvisor.hpp
```diff
@@ -26,6 +26,7 @@ namespace strategy {
  * - op::CudaAdviseAccessedByOperation
  * - op::CudaAdvisePreferredLocationOperation
  * - op::CudaAdviseReadMostlyOperation
+ * - op::HipAdviseOperation
```
kab163, line 29:
```suggestion
 * - op::CudaAdviseOperation
 * - op::HipAdviseOperation
```

kab163 (approved):
:(

PR #985, opened 2025-08-21
updating camp to c++17

kab163:
To avoid build warnings from Camp in Umpire builds, we need to update Camp to c++17


adayton1 (approved):

kab163:
Closing in favor of #987

PR #986, opened 2025-08-22
Fix handling of the CUDA language standard

### LINKED ISSUE #984: Fix handling of the CUDA language standard
**Describe the bug**

The CUDA language standard is hardcoded to c++17, preventing anyone from building with a newer language standard.

**To Reproduce**

The following will build the C++ code with c++20 and the CUDA code with c++17

`cmake -DENABLE_CUDA=ON -DBLT_CXX_STD=c++20 ../`

**Expected Behavior**

If I don't explicitly specify CMAKE_CUDA_STANDARD, I think it should match BLT_CXX_STD (which BLT takes care of for us under the hood, see below).

```
# This is what BLT does
if (ENABLE_CUDA)
       if (NOT DEFINED CMAKE_CUDA_STANDARD)
          set(CMAKE_CUDA_STANDARD ${CMAKE_CXX_STANDARD})
       endif()

       message(STATUS "Standard C++${CMAKE_CUDA_STANDARD} selected for CUDA")
    endif()

    if (ENABLE_HIP AND ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.21.0")
       if (NOT DEFINED CMAKE_HIP_STANDARD)
          set(CMAKE_HIP_STANDARD ${CMAKE_CXX_STANDARD})
       endif()

       message(STATUS "Standard C++${CMAKE_HIP_STANDARD} selected for HIP")
endif()
```

If I do explicitly set CMAKE_CUDA_STANDARD, I'm not sure if Umpire should overwrite the value I specified in order to match BLT_CXX_STD, or allow the two to be set separately. It might be convenient to have the greater flexibility in the latter case, though you'd also have to check that CMAKE_CUDA_STANDARD meets the minimum language standard of 17.

adayton1:
The CUDA language standard is hardcoded to c++17, preventing anyone from building with a newer language standard. This fix allows the user to specify CMAKE_CUDA_STANDARD, and if it is not specified, BLT will set it based on the value in BLT_CXX_STD.

Resolves #984

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,16 +23,39 @@ set(UMPIRE_VERSION_RC "")
 
 include(cmake/SetupUmpireOptions.cmake)
 
+# Default to c++17 if not specified by the user.
 set(BLT_CXX_STD "c++17" CACHE STRING "Version of C++ standard")
+
+# If BLT_CXX_STD is set by the user, check that it is at least 17.
+if("${BLT_CXX_STD}" STREQUAL "c++98" OR
+   "${BLT_CXX_STD}" STREQUAL "c++03" OR
+   "${BLT_CXX_STD}" STREQUAL "c++11" OR
+   "${BLT_CXX_STD}" STREQUAL "c++14")
+  message(FATAL_ERROR "Umpire requires minimum C++ standard of c++17. Please set BLT_CXX_STD accordingly.")
+endif()
+
+# If CMAKE_CUDA_STANDARD is set by the user, check that it is at least 17.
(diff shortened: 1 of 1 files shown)


rhornung67 (dismissed):

kab163 (dismissed):

adayton1:
I added some checks for the minimum CUDA and HIP standards since they can be set independently from BLT_CXX_STD.

adayton1:
@rhornung67, @kab163, @davidbeckingsale, any other comments? Since I pushed a change, it dismissed Rich's and Kristi's approvals.

davidbeckingsale (approved):

PR #995, opened 2025-09-08
Fix a communicator leak in HostMpi3SharedMemoryResource.

BradWhitlock:
I was using Umpire shared memory in Axom and I ran *valgrind* on my program and valgrind reported leaks due to arrays associated with `MPI_Win` in `HostMpi3SharedMemoryResource`. The destructor for this class contained a commented-out `MPI_Comm_free()` call as well as a *TODO* note about freeing the communicator later prior to `MPI_Finalize()`. I *assume* there was a good reason for not freeing the communicator in the destructor since it had been commented-out.

This PR takes a stab at freeing the communicator via a callback that is executed during the `MPI_Finalize` process. The `m_shared_comm` member is passed to the callback via a `void*` callback argument (the comm's integer name is stored in the `void*`). This is to avoid potential stack problems. With these changes, valgrind in Axom no longer reported these leaks.

### DIFF
--- a/src/umpire/resource/HostMpi3SharedMemoryResource.cpp
+++ b/src/umpire/resource/HostMpi3SharedMemoryResource.cpp
@@ -25,7 +25,7 @@ HostMpi3SharedMemoryResource::HostMpi3SharedMemoryResource(const std::string& na
   // by turning it into an int (as for Fortran) and then decoding that in the callback.
   int keyval = 0;
   MPI_Comm_create_keyval(MPI_COMM_NULL_COPY_FN, free_comm, &keyval, nullptr);
-  MPI_Comm_set_attr(MPI_COMM_SELF, keyval, (void *)(intptr_t)MPI_Comm_c2f(m_shared_comm));
+  MPI_Comm_set_attr(MPI_COMM_SELF, keyval, (void*)(intptr_t)MPI_Comm_c2f(m_shared_comm));
 }
 
 HostMpi3SharedMemoryResource::~HostMpi3SharedMemoryResource()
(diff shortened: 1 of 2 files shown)


davidbeckingsale (approved):
Thanks for this neat fix @BradWhitlock

adayton1:
Is this cleanup_cached_communicators routine fixing a different resource leak? https://github.com/LLNL/Umpire/blob/6fe14e1deffe9d63f51e0f3dd02e5fd92f2edb51/src/umpire/Umpire.cpp#L338

davidbeckingsale:
> Is this cleanup_cached_communicators routine fixing a different resource leak?
> 
> https://github.com/LLNL/Umpire/blob/6fe14e1deffe9d63f51e0f3dd02e5fd92f2edb51/src/umpire/Umpire.cpp#L338

Yes, communicators cached by this routine: https://github.com/LLNL/Umpire/blob/6fe14e1deffe9d63f51e0f3dd02e5fd92f2edb51/src/umpire/Umpire.cpp#L310

BradWhitlock:
I looked at `cleanup_cached_communicators()` and from what I could tell, it appears to be used for when you get a communicator via `umpire::get_communicator_for_allocator()`. Meanwhile, the `HostMpi3SharedMemoryResource` class creates its own split communicator that does not get put into that managed set of communicators. Another potential way of solving this would be to put `m_shared_comm` into the managed set of communicators and then make sure the user has to call that `cleanup_cached_communicators()` function. The TODO comment in the code seemed to point towards something akin to what I did.

adayton1:
I think your approach is fine, just wanted to make sure I was understanding. I could potentially see wanting a way to clean up the communicator/shared memory allocator before MPI_Finalize, but I don't think that is an immediate need.

adayton1 (approved):

Issue #1022, opened 2025-12-11
UMPIRE_ERROR macro emits 10k warnings when we compile


eherb:
**Describe the bug**

The UMPIRE_ERROR macro in the UMPIRE error-handling utilities uses the GNU extension ##__VA_ARGS__ to support calls with zero variadic arguments:

```
#define UMPIRE_ERROR(type, msg, ...) \
{ \
  type e{msg, std::string{__FILE__}, __LINE__, ##__VA_ARGS__}; \
  UMPIRE_LOG(Error, e.what()); \
  umpire::util::flush_files(); \
  throw e; \
}
```

When compiling in non-GNU or more strictly standard-conforming modes (e.g., with Clang and -Wgnu-zero-variadic-macro-arguments enabled), this triggers a large number of warnings of the form:

token pasting of ',' and VA_ARGS is a GNU extension [-Wgnu-zero-variadic-macro-arguments]

The warning arises at each call site that uses UMPIRE_ERROR with no variadic arguments. In large code bases this results in thousands of warnings, effectively obscuring other diagnostics and making it difficult to keep builds warning-clean.

**To Reproduce**

Build a code base that uses UMPIRE_ERROR(type, msg) both with and without extra arguments, for example:
```
UMPIRE_ERROR(MyErrorType, "basic error");
UMPIRE_ERROR(MyErrorType, "formatted error", some_value);
```
Compile with Clang (or another compiler that warns on GNU variadic macro extensions), using a standard-conforming language mode and default warning set. For example (Clang):

clang++ -std=c++17 -Wall -Wextra -pedantic ...

Observe warnings like:

warning: token pasting of ',' and __VA_ARGS__ is a GNU extension [-Wgnu-zero-variadic-macro-arguments]

emitted at each UMPIRE_ERROR call that omits the variadic arguments.

**Expected behavior**

UMPIRE’s public macros, including UMPIRE_ERROR, should be usable in strictly standard-conforming C++ builds without relying on GNU-specific preprocessor extensions. In particular:

Calling UMPIRE_ERROR(type, msg) (no extra args) and UMPIRE_ERROR(type, msg, extra...) (with extra args) should both compile cleanly without generating -Wgnu-zero-variadic-macro-arguments or similar diagnostics.

Downstream projects should not be required to relax language modes (e.g., switching from -std=c++17 to -std=gnu++17) or disable warnings (-Wno-gnu-zero-variadic-macro-arguments) just to use UMPIRE’s error macro.
[...]

PR #1024, opened 2025-12-12
Warnfix/herbold1/event virt destruct #1022

eherb:
Add a virtual destructor for event_store. This removes a warning on clang to prevent deletion through a base pointer. #1022


davidbeckingsale:
Merged in #1025

PR #1044, opened 2026-01-06
Address TODO Comments - Update FixedSizePool

kab163:


davidbeckingsale:
I think it would be good to see if this impacts performance in a real-world use case

kab163:
Will look in to how a double linked list would compare - with a benchmark to compare the 2 implementations (random allocation vs in order)

kab163:
We will close this for now - if needed, we can revisit later. Like comments say, we will need to benchmark this to make sure there are no unintended side-effects.

PR #1046, opened 2026-01-08
Add `.mailmap` file

tgamblin:
This will clean up the output of commands like `git shortlog` that aggregate work by different users. e.g., Kristi has used 59 different emails in the Umpire project.

It unfortunately looks like GitHub's own stats do not account for this, but if we do analytics on umpire, e.g., with [`contrib`](https://github.com/spack/contrib), this will help.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,5 +1,74 @@
-David Beckingsale <davidbeckingsale@gmail.com> David Beckingsale <beckingsale1@llnl.gov>
-David Beckingsale <davidbeckingsale@gmail.com> David Alexander Beckingsale <beckingsale1@llnl.gov>
-Marty McFadden <mcfadden8@llnl.gov> Marty McFadden <mcfadden8@users.noreply.github.com>
-Marty McFadden <mcfadden8@llnl.gov> Marty Mcfadden <mcfadden8@llnl.gov>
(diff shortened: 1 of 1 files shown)


davidbeckingsale (approved):

rhornung67:
@tgamblin a lot of those emails correspond to machines that no longer exist (lassen, rzansel). is it necessary to add those?  

I guess so, if you want to look at older commits.

adayton1:
Is there some git configuration variable that can be set to avoid all the commits coming from "different" emails?

tgamblin:
@adayton1:

> Is there some git configuration variable that can be set to avoid all the commits coming from "different" emails?

I *think* `.mailmap` is that. It makes `git` translate the emails in commits to the ones in `.mailmap`.  You could also rewrite history, and change the emails in old commits, but it would disrupt users of the repository -- they'd have to re-clone or do some git fanciness to be able to pull again.

@rhornung67:

> I guess so, if you want to look at older commits.

Yes exactly -- the commits are permanent unless you want to rewrite history with `git filter-branch` or something like that. `.mailmap` makes the `git` UI have consistent emails, though.

kab163 (approved):

PR #1053, opened 2026-02-13
Update RSCI and adapt to machine-status-override

adrienbernede:

PR #1055, opened 2026-02-25
Use project specific variable to specify github status token

adrienbernede:
Fix issue with machine outage report.

The issue identified is that token cannot be accessed correctly when defined at group level anymore.

Instead, we defined the token at project level.


davidbeckingsale (approved):

adrienbernede:
@davidbeckingsale this is ready. You may ignore the error which is a bug handled by #1053.

PR #1061, opened 2026-04-01
IPC Shared Memory Release

kab163:


examples/ipc_shared_memory_release.cpp
```diff
@@ -0,0 +1,154 @@
...
+  const std::size_t ps = page_size();
+  for (std::size_t i = 0; i < bytes; i += ps) {
+    buffer[i] = static_cast<std::uint8_t>(buffer[i] + 1);
+  }
+  if (bytes > 0) {
+    buffer[bytes - 1] = static_cast<std::uint8_t>(buffer[bytes - 1] + 1);
+  }
+}
+
+std::size_t get_smaps_rss_bytes_for_mapping(const std::string& mapping_name_substr)
```
davidbeckingsale: We could make this an umpire utility function similar to `get_process_memory_usage`?

examples/ipc_shared_memory_release.cpp
```diff
@@ -0,0 +1,118 @@
...
+  } else {
+    return std::to_string(bytes) + " B";
+  }
+}
+
+void touch_one_byte_per_page(std::uint8_t* buffer, std::size_t bytes)
+{
+  const std::size_t ps = page_size();
+  for (std::size_t i = 0; i < bytes; i += ps) {
+    buffer[i] = static_cast<std::uint8_t>(buffer[i] + 1);
```
adayton1: Why not simplify to `++buffer[i];`?

examples/ipc_shared_memory_release.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  const std::size_t shm_rss_after_free = umpire::get_mapping_memory_usage(allocator_name);
+  if (shm_rss_after_free > 0) {
+    std::cout << "Shared segment RSS after deallocate: " << format_bytes(shm_rss_after_free) << "\n";
+  }
+
+  allocator.release();
+
+  const std::size_t rss_after_release = umpire::get_process_memory_usage();
+  std::cout << "RSS after allocator.release(): " << format_bytes(rss_after_release) << "\n";
```
adayton1: "allocator.release()" should probably be changed to "release" to match the other output.

src/umpire/Umpire.cpp
```diff
@@ -17,7 +19,9 @@
 
 #include "umpire/ResourceManager.hpp"
 #include "umpire/config.hpp"
+#if defined(UMPIRE_ENABLE_IPC_SHARED_MEMORY)
```
adayton1, line 22: Is it a different file included for MPI shared memory?

tests/unit/resource/ipc_shared_memory_resource_tests.cpp
```diff
@@ -163,6 +170,50 @@ class SharedMemoryTest : public ::testing::Test {
...
+    long ps = ::sysconf(_SC_PAGESIZE);
+    return (ps > 0) ? static_cast<std::size_t>(ps) : 4096;
+  }
+
+  void touch_one_byte_per_page(std::uint8_t* buffer, std::size_t bytes)
+  {
+    const std::size_t ps = page_size();
+
+    for (std::size_t i = 0; i < bytes; i += ps) {
+      buffer[i] = static_cast<std::uint8_t>(buffer[i] + 1);
```
adayton1: Why not simplify to `++buffer[i];`?

src/umpire/resource/HostSharedMemoryResourceImpl.hpp
```diff
@@ -301,6 +303,56 @@ class HostSharedMemoryResource::impl {
     return ptr;
   }
 
+  void release()
+  {
+    int err{0};
+    if ((err = pthread_mutex_lock(&m_segment->mutex)) != 0) {
```
adayton1, line 309: Why pthread and not std::mutex and std::lock?

adayton1: I see it's used elsewhere. Don't worry about it.

adayton1:
I've got a use case where I need to read a mesh from a file, make the mesh accessible on every MPI rank, do some processing, then completely free all memory associated with the mesh. Is that accomplished by calling release or some other way?

In a related scenario I could imagine keeping a shared memory allocator around with enough memory for the largest mesh in a collection of meshes. Each mesh is read in, processed, then discarded. Would shared memory usage keep growing without release?

adayton1:
I'm approving, but please address my remaining comments.

adayton1 (dismissed):

src/umpire/Umpire.cpp
kab163: I am starting with IPC Shared memory first - if it seems to work well, I'll expand to MPI3 shared mem

kab163:
> I've got a use case where I need to read a mesh from a file, make the mesh accessible on every MPI rank, do some processing, then completely free all memory associated with the mesh. Is that accomplished by calling release or some other way?
> 
> In a related scenario I could imagine keeping a shared memory allocator around with enough memory for the largest mesh in a collection of meshes. Each mesh is read in, processed, then discarded. Would shared memory usage keep growing without release?

Shared memory would continue growing without the release. In the second scenario, this would hopefully allow you to keep a shared memory allocator around with enough memory for the largest mesh in a collection of meshes, and then just keep reusing the memory for each additional mesh you need (it won't grow since you've made the size large enough for the largest mesh). That's the idea.

examples/ipc_shared_memory_release.cpp
```diff
@@ -0,0 +1,118 @@
...
+//
+// SPDX-License-Identifier: (MIT)
+//////////////////////////////////////////////////////////////////////////////
+
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+#include <string>
+
+#include <unistd.h>
```
adayton1, line 14: Is this a linux header? Are there other things in here that need macro guards? Or should this example only be built if this is a linux build?

examples/ipc_shared_memory_release.cpp
```diff
@@ -0,0 +1,118 @@
...
+  constexpr std::size_t alloc_size = 256ULL * 1024ULL * 1024ULL;
+
+  auto& rm = umpire::ResourceManager::getInstance();
+  auto traits = umpire::get_default_resource_traits("SHARED::POSIX");
+  traits.size = segment_size;
+
+  const std::string allocator_name = "SHARED::POSIX::release_example";
+  umpire::Allocator allocator = rm.makeResource(allocator_name, traits);
+
+  const std::size_t rss_before = umpire::get_process_memory_usage();
```
adayton1, line 77: Should this example use the new node memory utility function that was just added?

tests/unit/resource/ipc_shared_memory_resource_tests.cpp
```diff
@@ -222,7 +273,97 @@ TEST_F(SharedMemoryTest, UnitTests)
...
+    ASSERT_NO_THROW(allocator.release(););
+    MPI_Barrier(MPI_COMM_WORLD);
+    ASSERT_EQ(shmem_resource->getActualSize(), shmem_state->initial_size);
+
+    MPI_Barrier(MPI_COMM_WORLD);
+    ASSERT_NO_THROW(allocator.release(););
+    MPI_Barrier(MPI_COMM_WORLD);
+    ASSERT_EQ(shmem_resource->getActualSize(), shmem_state->initial_size);
+
+    MPI_Barrier(MPI_COMM_WORLD);
```
adayton1, line 287: Is this barrier actually necessary?

tests/unit/resource/ipc_shared_memory_resource_tests.cpp
```diff
@@ -222,7 +273,97 @@ TEST_F(SharedMemoryTest, UnitTests)
     ASSERT_EQ(shmem_resource->getActualSize(), shmem_state->initial_size);
 
     MPI_Barrier(MPI_COMM_WORLD);
+    ASSERT_NO_THROW(allocator.release(););
+    MPI_Barrier(MPI_COMM_WORLD);
+    ASSERT_EQ(shmem_resource->getActualSize(), shmem_state->initial_size);
+
+    MPI_Barrier(MPI_COMM_WORLD);
+    ASSERT_NO_THROW(allocator.release(););
```
adayton1, line 283: Is this intentionally testing what happens when release is called twice on the same allocator? If so, can you add a comment?

tests/unit/resource/ipc_shared_memory_resource_tests.cpp
adayton1: You have a lot of barriers throughout the testing/examples. It would be nice to know when barriers are actually necessary (what function calls and/or writes trigger the need for a barrier).

adayton1: Now that I think about this, isn't this implementation using IPC handles and not MPI?

kab163: The way I went about this, I decided to start with the IPC shared memory first and then move on to MPI3 next if all goes well.

examples/ipc_shared_memory_release.cpp
kab163: It is a POSIX/Unix header, so it'll exist on linux and mac, but not windows - but we don't allow ipc shared mem on windows builds anyway

tests/unit/resource/ipc_shared_memory_resource_tests.cpp
adayton1: But why do the ipc specific tests need mpi barriers?

adayton1: That was probably a dumb question. I don't know much about IPC handles. But I guess there needs to be some form of synchronization and MPI is a convenient way of doing that synchronization.

kab163: Yea, that's pretty much what's happening. It is confusing because I am working on 2 PRs at the same time and one is MPI specific and one is IPC specific - I have been getting it mixed up too. But I think it is ok to use MPI for synchronization in this test

tests/unit/resource/ipc_shared_memory_resource_tests.cpp
kab163: done

examples/ipc_shared_memory_release.cpp
kab163: Added!

adayton1 (dismissed):

tests/unit/resource/ipc_shared_memory_resource_tests.cpp
kab163: I tried removing some of the barriers that I thought might not be necessary, but that ended up breaking some things. So I'll put a pin in that until I have more time to go in and study it a bit more. For now I'll keep the barriers.

adayton1 (approved):

rhornung67 (approved):

PR #1067, opened 2026-04-08
Umpire namespace target updates

kab163:


davidbeckingsale:
Should we use a lowercase namespace? `umpire::`?

davidbeckingsale (approved):

PR #1072, opened 2026-05-04
MPI shared memory socket scope

kab163:


src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,19 +7,174 @@
...
 
 namespace umpire {
 namespace resource {
 
+namespace {
+
+constexpr int IGNORE_KEY{0};
+
+#if defined(__linux__)
+bool try_get_socket_color_for_cpu(int cpu, int& color)
```
adayton1: Could you add some high level documentation for the new functions and their parameters in this file?

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,19 +7,174 @@
...
+      UMPIRE_LOG(Debug, "Using Linux physical_package_id " << color << " for cpu " << cpu);
+      return true;
+    }
+
+    UMPIRE_LOG(Debug, "Could not read physical_package_id for cpu " << cpu);
+  } else {
+    UMPIRE_LOG(Debug, "Could not open physical_package_id for cpu " << cpu);
+  }
+
+#if defined(UMPIRE_ENABLE_NUMA) && defined(__linux__)
```
adayton1: It looks like this check for `defined(__linux__)` is redundant.

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,19 +7,174 @@
...
+               "shared_scope::socket for MPI3 shared memory requires Linux CPU affinity information");
+#endif
+}
+
+MPI_Comm create_shared_communicator(MPI_Comm comm, MemoryResourceTraits::shared_scope scope)
+{
+  MPI_Comm shared_comm{MPI_COMM_NULL};
+
+  if (scope == MemoryResourceTraits::shared_scope::node) {
+    MPI_Comm_split_type(comm, MPI_COMM_TYPE_SHARED, IGNORE_KEY, MPI_INFO_NULL, &shared_comm);
```
adayton1: Consider adding error checking to these MPI calls.

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,26 +7,254 @@
...
+ *
+ * Socket-scoped shared memory requires each rank to be pinned to exactly one
+ * socket so that all ranks with the same socket color can be grouped into one
+ * shared-memory communicator.
+ *
+ * \return Socket color used to split the node-local communicator.
+ */
+int get_socket_color_from_affinity()
+{
+#if defined(__linux__)
```
adayton1: I think this would all be easier to follow if all the linux specific code was inside one macro guard (i.e. this function and the other two above). The `#else` block in this function could be moved inside create_shared_communicator.

adayton1 (dismissed):

adayton1:
Would it be worth considering adding hwloc as a dependency to get the socket info?

kab163:
> Would it be worth considering adding hwloc as a dependency to get the socket info?

I think it would be worth considering... I think for now we should merge this and see how people like it. If there's a need/desire to add hwloc later, we can. (Maybe as an optional dependency - like use hwloc if detected, otherwise use the linux system calls/numa like the branch does now.)

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -10,13 +10,17 @@
...
 #include <fstream>
+#include <limits>
 #include <set>
 #include <sstream>
 #include <string>
 
 #if defined(__linux__)
 #include <sched.h>
 #include <unistd.h>
+const bool is_linux = true;
```
adayton1: constexpr

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -217,6 +229,11 @@ MPI_Comm create_shared_communicator(MPI_Comm comm, MemoryResourceTraits::shared_
   }
 
   if (scope == MemoryResourceTraits::shared_scope::socket) {
+    if(!is_linux) {
```
adayton1: if constextpr

kab163: you're totally right!

kab163:
Update: Hubcast doesn't understand jobs that are allowed to fail. We can either fix the job failure or try to fix within Hubcast somehow.. stay tuned

adayton1:
> Update: Hubcast doesn't understand jobs that are allowed to fail. We can either fix the job failure or try to fix within Hubcast somehow.. stay tuned

You'll need to update branch protections under settings -> branches -> develop and list the CI jobs that are required to pass (instead of ci/gitlab.lc.llnl.gov).

kab163:
> > Update: Hubcast doesn't understand jobs that are allowed to fail. We can either fix the job failure or try to fix within Hubcast somehow.. stay tuned
> 
> You'll need to update branch protections under settings -> branches -> develop and list the CI jobs that are required to pass (instead of ci/gitlab.lc.llnl.gov).

I don't see where I'd be able to do this in the develop branch's settings... (I'm getting copilot to help me...)

.github/hubcast.yml
```diff
@@ -5,7 +5,7 @@ Repo:
 
   check_name: gitlab-ci
 
-  check_types: [child-pipelines, jobs]
+  check_types: [pipelines, child-pipelines]
```
adayton1, line 8: With pipelines, you should be able to change the branch protection rules to just "gitlab-ci" passing. The "gitlab-ci" pipeline will still report a passing status when allowed failures occur.

adayton1: You might still want jobs here as well so you can see exactly which compiler combinations fail.

src/umpire/Umpire.hpp
```diff
@@ -193,19 +193,36 @@ umpire::MemoryResourceTraits get_default_resource_traits(const std::string& name
  */
 void* find_pointer_from_name(Allocator allocator, const std::string& name);
 
-#if defined(UMPIRE_ENABLE_MPI)
-/*!
- * \brief Return the MPI communicator for a shared memory allocator.
- *
+	#if defined(UMPIRE_ENABLE_MPI)
```
adayton1: Why the change in indentation in this file?

examples/mpi3_shared_memory.cpp
```diff
@@ -16,16 +16,37 @@ int main(int argc, char** argv)
...
+  // Node scope is the default for MPI3 shared memory; socket scope is also supported.
+  // Pass `--socket` to request socket scope (falls back to node scope when the rank is not pinned to one socket).
+  const bool request_socket_scope = (argc > 1) && (std::string{argv[1]} == "--socket");
+  traits.scope = request_socket_scope ? umpire::MemoryResourceTraits::shared_scope::socket
+                                      : umpire::MemoryResourceTraits::shared_scope::node;
+
+  if (traits.scope == umpire::MemoryResourceTraits::shared_scope::socket) {
+    std::string reason;
+	  if (!umpire::affinity_maps_to_single_socket(reason)) {
+      if (world_rank == 0) {
```
adayton1, line 36: Indentation is a little weird here.

examples/mpi3_shared_memory.cpp
```diff
@@ -16,16 +16,37 @@ int main(int argc, char** argv)
...
+        std::cerr << "Requested socket-scoped MPI3 shared memory, but CPU affinity does not map to a single socket: "
+                  << reason << "\nFalling back to node-scoped shared memory.\n";
+      }
+      traits.scope = umpire::MemoryResourceTraits::shared_scope::node;
+    } else {
+      if (world_rank == 0) {
+        std::cout << "Running with socket-scope!" << std::endl;
```
adayton1, line 43: I think this output message is wrong. I could be mistaken, but if control flow reaches this point, it is actually node scope.

examples/mpi3_shared_memory.cpp
```diff
@@ -47,6 +68,7 @@ int main(int argc, char** argv)
 
   mpi3_shm_allocator.deallocate(data);
 
+  // Since we called get_communicator_for_allocator...
```
adayton1:
```suggestion
  // Since we called get_communicator_for_allocator(), clean up is needed.
```

src/umpire/Umpire.hpp
kab163: Classic tabs vs spaces stuff going on there

examples/mpi3_shared_memory.cpp
kab163: Line 33 checks to see if socket-scope is requested, from there it checks the affinity. If CPU affinity does not map to a single socket, then it falls back to node, otherwise it continues onward with socket.. that's what I understand this to be doing

.github/hubcast.yml
kab163: Yea, that part is tricky. I'll talk to Alec about this. We have some time to get this sorted out

kab163:
@lc-hubcast approve

kab163 (commented):
@lc-hubcast approve

examples/mpi3_shared_memory.cpp
adayton1: Oh right.

src/umpire/Umpire.hpp
```diff
@@ -204,6 +204,23 @@ MPI_Comm get_communicator_for_allocator(Allocator a, MPI_Comm comm);
...
+ *
+ * When MPI3 shared memory support is disabled (or the platform does not
+ * provide the required affinity information), this function returns false and
+ * populates \p reason with a diagnostic message.
+ *
+ * \param reason Diagnostic message describing why the check failed.
+ *
+ * \return true when the affinity mask is non-empty and maps to a single socket.
+ */
+bool affinity_maps_to_single_socket(std::string& reason);
```
adayton1: Is there a better place for this method?

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,26 +7,346 @@
...
+ * \param comm Parent communicator used to discover ranks that may share host
+ *        memory.
+ * \param scope Requested sharing scope from MemoryResourceTraits. Node scope
+ *        keeps one communicator per node, while socket scope further splits the
+ *        node-local communicator using socket affinity.
+ *
+ * \return Communicator containing exactly the ranks that participate in each
+ * shared-memory allocation for the resource.
+ */
+	MPI_Comm create_shared_communicator(MPI_Comm comm, MemoryResourceTraits::shared_scope scope)
```
adayton1: The indentation seems a little off.

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,26 +7,346 @@
...
+    return shared_comm;
+  }
+
+  if (scope == MemoryResourceTraits::shared_scope::socket) {
+    if constexpr (!is_linux) {
+      UMPIRE_ERROR(runtime_error,
+               "shared_scope::socket for MPI3 shared memory requires Linux CPU affinity information");
+    }
+
+    const int color = get_socket_color_from_affinity();
```
adayton1: I don't think this will compile on platforms other than linux. Won't the socket scope code require a macro guard (e.g. `#if defined(__linux__)`) rather than just an if constexpr?

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,26 +7,346 @@
...
+    return shared_comm;
+  }
+
+  UMPIRE_ERROR(runtime_error,
+               fmt::format("Unsupported shared communicator scope: {}", to_string(scope)));
+}
+
+} // end anonymous namespace
+
+bool affinity_maps_to_single_socket(std::string& reason)
```
adayton1, line 92: Isn't this also defined in umpire.hpp and umpire.cpp?

.github/hubcast.yml
kab163: I made an issue for this

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
kab163: I made some changes for this...

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
kab163: I made some changes for this...

src/umpire/Umpire.hpp
kab163: I made some changes for this...

tests/unit/resource/mpi3_shared_memory_resource_tests.cpp
```diff
@@ -5,10 +5,31 @@
...
+#include <exception>
+
 #include "gtest/gtest.h"
 #include "mpi.h"
 #include "umpire/Umpire.hpp"
 
+#if defined(__linux__)
+namespace {
+
+bool all_ranks_affinity_maps_to_single_socket(std::string& reason)
```
adayton1: This function appears to be duplicated a couple of times.

examples/mpi3_shared_memory.cpp
```diff
@@ -1,31 +1,65 @@
...
 
 #include "umpire/Allocator.hpp"
 #include "umpire/ResourceManager.hpp"
 #include "umpire/Umpire.hpp"
 #include "umpire/config.hpp"
-#include "umpire/resource/HostMpi3SharedMemoryResource.hpp"
 #include "umpire/strategy/NamedAllocationStrategy.hpp"
 #include "umpire/util/MemoryResourceTraits.hpp"
 
+bool all_ranks_affinity_maps_to_single_socket(std::string& reason)
```
adayton1: This is duplicated a bunch.

src/umpire/Umpire.cpp
```diff
@@ -395,6 +402,16 @@ void cleanup_cached_communicators()
 }
 #endif
 
+bool affinity_maps_to_single_socket(std::string& reason)
```
adayton1: Is this wrapper function actually needed?

tests/integration/get_communicator_tests.cpp
```diff
@@ -5,12 +5,42 @@
...
+const std::string alloc_name = "SHARED::MPI3::allocator";
+#elif defined (UMPIRE_ENABLE_IPC_SHARED_MEMORY)
+const std::string trait_name = "SHARED::IPC";
+const std::string alloc_name = "SHARED::IPC::allocator";
+#endif
+
+#if defined(__linux__) && defined(UMPIRE_ENABLE_MPI3_SHARED_MEMORY)
+namespace {
+
+bool all_ranks_affinity_maps_to_single_socket(std::string& reason)
```
adayton1: This is the other duplicate.

tests/integration/get_communicator_tests.cpp
```diff
@@ -38,6 +67,37 @@ TEST(GetCommunicator, SharedAndCached)
...
+  }
+
+  umpire::Allocator allocator;
+  try {
+    allocator = rm.makeResource("SHARED::MPI3::socket_allocator", traits);
+  } catch (const std::exception& e) {
+    GTEST_SKIP() << "Socket-scoped MPI3 shared memory requires ranks bound to a single socket (" << e.what() << ")";
+  }
+
+  auto comm = umpire::get_communicator_for_allocator(allocator, MPI_COMM_WORLD);
```
adayton1, line 72: Don't these communicators need to be cleaned up at the end of the test?

tests/unit/resource/mpi3_shared_memory_resource_tests.cpp
```diff
@@ -88,6 +109,56 @@ TEST_F(MPISharedMemoryTest, SharedMemoryVisibility)
...
+    socket_data[0] = 42;
+    socket_data[1] = nranks;
+  }
+
+  MPI_Barrier(comm);
+
+  ASSERT_EQ(socket_data[0], 42);
+  ASSERT_EQ(socket_data[1], nranks);
+
+  allocator.deallocate(socket_data);
```
adayton1, line 139: Clean up cached communicators?

tests/integration/get_communicator_tests.cpp
kab163: Amazing, thank you for catching that. Yes they do

src/umpire/Umpire.cpp
kab163: Ah, no this can be reworked. I think codex took me a bit too literally here 😵‍💫

kab163: Without the wrapper, we would call "umpire::resource::affinity_maps_to_single_socket" and with it we call "umpire::affinity_maps_to_single_socket".  The point of that wrapper is to make it somewhat easier to call (would be called in the example or test files). But I guess we could also do something else - is there a better way to do this?

kab163: It could also potentially have a shorter name in the wrapper version to make it easier to remember...

adayton1: I don't think it really matters all that much. If this is something used directly by clients, then it's probably OK to keep the wrapper. I'm assuming you aren't defining it at the top level because that would introduce circular dependencies?

tests/unit/resource/mpi3_shared_memory_resource_tests.cpp
adayton1: This still looks duplicated at least twice (once in each test file).

adayton1 (dismissed):

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,26 +7,295 @@
...
+  const int status = MPI_Error_string(error_code, buffer, &length);
+
+  if (status != MPI_SUCCESS) {
+    return fmt::format("MPI error code {} (MPI_Error_string failed with code {})", error_code, status);
+  }
+
+  return std::string{buffer, static_cast<std::size_t>(length)};
+}
+
+void check_mpi_call(int error_code, const char* call_name)
```
davidbeckingsale: this could be moved to the umpire MPI header and applied to all MPI calls elsewhere.

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,26 +7,295 @@
...
 #include "umpire/util/error.hpp"
 
 namespace umpire {
 namespace resource {
 
+namespace {
+
+constexpr int IGNORE_KEY{0};
+
+std::string get_mpi3_resource_error_message(int error_code)
```
davidbeckingsale: Same comment as below.

davidbeckingsale (dismissed):
Looks good. See comments. Should the affinity related code be moved to a utility file (e.g. like we have a numa file with utility functions)

adayton1:
The organization looks a lot better now!

adayton1 (dismissed):

kab163:
All tests should pass except the Windows docker builds due to a camp test failure that is being worked out. If all those tests pass (except Windows) we can force merge.

kab163:
> All tests should pass except the Windows docker builds due to a camp test failure that is being worked out. If all those tests pass (except Windows) we can force merge.

Um ok I guess Windows works too!

davidbeckingsale (approved):

src/umpire/resource/HostMpi3SharedMemoryResource.cpp
```diff
@@ -7,26 +7,106 @@
...
+    return shared_comm;
+  }
+
+  UMPIRE_ERROR(runtime_error,
+               fmt::format("Unsupported shared communicator scope: {}", to_string(scope)));
+}
+
+} // end anonymous namespace
+
+bool affinity_maps_to_single_socket(std::string& reason)
```
adayton1, line 92: I don't think this wrapper adds much.

adayton1 (approved):

PR #1074, opened 2026-06-05
Switch to Hubcast

davidbeckingsale:


kab163 (dismissed):

.github/hubcast.yml
```diff
@@ -0,0 +1,14 @@
+Repo:
+  dest_org: radiuss
+
+  dest_name: Umpire
+
+  check_name: gitlab-ci
+
+  check_types: [pipeline, jobs]
```
kab163: Do we want jobs? Or just pipelines and child pipelines...

PR #1075, opened 2026-06-10
Update radiuss-shared-ci with builtin hubcast tweaks

adrienbernede:
Testing changes for RADIUSS Shared CI 2026.6.0 release.

Plan is to make an RSCI release first.

Optimize alloc on matrix.

TODO:
- [x] release radiuss-shared-ci 2026.6.0
- [x] unknown error in the tests on matrix (https://github.com/llnl/Umpire/pull/1085)


adrienbernede:
@adayton1 @kab163 
Thanks for approving, #1085, diving in the logs on this PR, you could see it fixed the failure of the associated test.
I am now left with a more annoying error: **the replay_tests failure**.

I analyzed the error:
With the new allocation specifying that we want **only 1 gpu** (to get in the machine faster), we **changed the configuration** in which the replay test runs, and therefore the comparison with the reference fails.

"Just update the reference"...
Unfortunately, I see that the **reference file is not tied to cuda** config only. So changing the reference based on the cuda replay test will likely change the reference for the AMD replay test, which will then fail.

Hope...
However, digging deeper, I found in tests/integration/CMakeLists.txt that the **replay test is not run on HIP**.

This all look a little fragile in my opinion.
The fact that the reference does not depend on the device architecture... I don't know, maybe umpire is device agnostic, however, **the reference being used should depend on the number of devices**.

What do you guys think is my way out here ?
_I'd suggest regenerate the reference, push it, and submit an issue to have the reference depend on the number of devices in the future._

adrienbernede:
@adayton1 @kab163 

Updating the reference worked.
Please approve asap, this PR improves access to matrix.

kab163 (approved):

PR #1084, opened 2026-07-06
Util Function to get Available Memory on Node

kab163:


adayton1 (dismissed):

rhornung67 (commented):
Do you want to add a test for this?

rhornung67:
@lc-hubcast approve

adayton1:
I second Rich's suggestion to add a test.

tests/unit/util/node_memory_tests.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+TEST(NodeMemory, GetNodeAvailableMemoryWithCustomDefault)
+{
+  // Test with custom default value
+  double custom_default = 12345.67;
+  double available_memory = umpire::util::get_node_available_memory(custom_default);
+
+#if defined(__linux__)
+  // On Linux, we expect a positive value (actual available memory)
+  // It should NOT be the custom default
```
adayton1, line 38: Can you add EXPECT_NE(available_memory, custom_default)? Or whatever the floating point equivalent gtest macro is?

tests/unit/util/node_memory_tests.cpp
```diff
@@ -0,0 +1,62 @@
...
+  EXPECT_GE(available_memory, 0.0);
+#else
+  // On non-Linux systems, it should return the custom default
+  EXPECT_EQ(available_memory, custom_default);
+#endif
+}
+
+TEST(NodeMemory, GetNodeAvailableMemoryReasonableValue)
```
adayton1: I don't think this test really adds any value.

adayton1: It's already covered by the GetNodeAvailableMemory test.

kab163: > I don't think this test really adds any value.

It's trying to make sure it returned a value that's between 0 and something totally wild.. but yea it may be too odd of a check... I'll adjust this

adayton1: Doesn't the GetNodeAvailableMemory test already do that?

adayton1 (approved):

rhornung67 (approved):

adayton1:
Can you add a release note?

kab163:
> Can you add a release note?

Yes, when I do the release, I'll make sure to make note of this PR and the change of CMake targets (among others)

PR #1085, opened 2026-07-09
Fix memory resource traits device enumeration

adrienbernede:
## Summary
- enumerate only additional per-device memory resources as DEVICE::1 through DEVICE::N-1, leaving DEVICE as the primary device case
- avoid asserting DEVICE::0 has a distinct canonical allocator name when it aliases DEVICE
- initialize memory resource traits test fixture allocator pointers before SetUp can throw

## Background
The original CUDA failure occurred when a one-visible-GPU job generated DEVICE::1, which was not registered. After changing the loop to include DEVICE::0, an AMD GPU job showed that DEVICE::0 can resolve to the canonical DEVICE allocator, so the test should not treat DEVICE::0 as a separate named allocator.

Assisted by Codex with ChatGPT 5.5

### DIFF
--- a/tests/integration/memory_resource_traits_tests.cpp
+++ b/tests/integration/memory_resource_traits_tests.cpp
@@ -77,7 +77,7 @@ std::vector<std::string> memory_resource_strings()
 #if defined(UMPIRE_ENABLE_DEVICE)
   resources.push_back("DEVICE");
   auto& rm = umpire::ResourceManager::getInstance();
-  for (int id = 0; id < rm.getNumDevices(); id++) {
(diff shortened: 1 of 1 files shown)


kab163 (approved):

adayton1 (approved):

kab163:
@adrienbernede - can this PR be closed in favor of your other one (#1075)? It looks like these changes are also included there

adrienbernede:
@kab163 exactly what I was planning to do: this PR uses the old matrix alloc that will have a hard time getting in.

adrienbernede:
closing in favor of #1075

PR #1092, opened 2026-07-23
Stop setting spack-packages builtin repo ref in shared environment

adrienbernede:
In this PR, the definition of the spack-packages reference to use in CI is transferred from the shared environment configuration in Radiuss-Spack-Configs to the uberenv settings file .uberenv_config.json.

This will allow developers to easily change the spack-packages used in the project CI (and build_and_test.sh use in general).

This required an update of radiuss-spack-configs to remove the reference from the shared environment.


adrienbernede:
@davidbeckingsale this is ready to ship!

kab163 (approved):

PR #1099, opened 2026-07-28
Alternative fix for FixedMallocPool touching unused pages

davidbeckingsale:
Uses free list for recycled slots and a bump pointer for fresh slots

### DIFF
--- a/src/umpire/util/FixedMallocPool.cpp
+++ b/src/umpire/util/FixedMallocPool.cpp
@@ -48,7 +48,7 @@ FixedMallocPool::FixedMallocPool(const std::size_t object_bytes, const std::size
   // Free-list links are stored in the slots themselves
   UMPIRE_ASSERT(object_bytes >= sizeof(unsigned int));
   // Slot indices (free-list links and num_used) are unsigned int
-  UMPIRE_ASSERT(objects_per_pool <= std::numeric_limits<unsigned int>::max());
+  UMPIRE_ASSERT(objects_per_pool <= static_cast<std::size_t>(std::numeric_limits<unsigned int>::max()));
   newPool();
 }
 
--- a/src/umpire/util/FixedMallocPool.cpp
+++ b/src/umpire/util/FixedMallocPool.cpp
@@ -6,9 +6,10 @@
 //////////////////////////////////////////////////////////////////////////////
 
 // FixedMallocPool hands out fixed-size slots from malloc()'d pools.
-// Recycled slots are kept on a LIFO free list linked through their own
-// storage; slots that have never been handed out are dispensed in order
-// with a bump pointer and are never read or written until allocated.
+// Recycled slots are kept on a LIFO free list of slot indices linked
+// through the slots' own storage; slots that have never been handed out
+// are dispensed in order with a bump pointer and are never read or
+// written until allocated.
 
 #include "umpire/util/FixedMallocPool.hpp"
 
@@ -28,8 +29,15 @@ inline unsigned char* FixedMallocPool::addr_from_index(const FixedMallocPool::Po
   return p.data + i * m_obj_bytes;
 }
 
+inline unsigned int FixedMallocPool::index_from_addr(const FixedMallocPool::Pool& p, const unsigned char* ptr) const
+{
+  return static_cast<unsigned int>((ptr - p.data) / m_obj_bytes);
+}
+
 FixedMallocPool::Pool::Pool(const std::size_t object_bytes, const std::size_t objects_per_pool)
-    : data(static_cast<unsigned char*>(std::malloc(object_bytes * objects_per_pool))), free_list(nullptr), num_used(0)
+    : data(static_cast<unsigned char*>(std::malloc(object_bytes * objects_per_pool))),
+      free_list(static_cast<unsigned int>(objects_per_pool)),
+      num_used(0)
 {
 }
 
@@ -37,7 +45,7 @@ FixedMallocPool::FixedMallocPool(const std::size_t object_bytes, const std::size
     : m_obj_bytes(object_bytes), m_obj_per_pool(objects_per_pool), m_data_bytes(m_obj_bytes * m_obj_per_pool), m_pool()
 {
   // Free-list links are stored in the slots themselves
-  UMPIRE_ASSERT(object_bytes >= sizeof(unsigned char*));
+  UMPIRE_ASSERT(object_bytes >= sizeof(unsigned int));
   newPool();
 }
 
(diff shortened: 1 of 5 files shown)


MrBurmark (dismissed):

src/umpire/util/FixedMallocPool.cpp
```diff
@@ -26,22 +28,16 @@ inline unsigned char* FixedMallocPool::addr_from_index(const FixedMallocPool::Po
...
-      num_free(objects_per_pool)
+    : data(static_cast<unsigned char*>(std::malloc(object_bytes * objects_per_pool))), free_list(nullptr), num_used(0)
 {
 }
 
 FixedMallocPool::FixedMallocPool(const std::size_t object_bytes, const std::size_t objects_per_pool)
     : m_obj_bytes(object_bytes), m_obj_per_pool(objects_per_pool), m_data_bytes(m_obj_bytes * m_obj_per_pool), m_pool()
 {
+  // Free-list links are stored in the slots themselves
+  UMPIRE_ASSERT(object_bytes >= sizeof(unsigned char*));
```
MrBurmark: This seems to be triggering a test failure, you could take the max of `object_bytes` and `sizeof(unsigned char*)` to always be compatible. Or you could switch the linked list to use integral offsets with a smaller type like int instead of pointers if `objects_per_pool` isn't too high.

adayton1:
These changes look reasonable to me. I'm working with some others to test if it fixes the resident page size growth we were seeing.

adayton1:
For context, one application is seeing Rss growth of about 80 KB per cycle coming from Umpire, which is about 800 MB over the course of a run of 10000 cycles (and some problems run longer). So this is pretty significant.

adayton1:
It looks like the mac build is failing with the same error that was hitting some raja builds. IIRC, it's due to a bug in one of the CMake 4 versions (4.4?) and rolling back to 3.29 or something close to that fixes the issue.

davidbeckingsale:
> For context, one application is seeing Rss growth of about 80 KB per cycle coming from Umpire, which is about 800 MB over the course of a run of 10000 cycles (and some problems run longer). So this is pretty significant.

Thank for this info. We will prioritize fixing prior to release.

rhornung67:
Unfortunately, the change I made to the RAJA GHA build script was not correct. It appeared to fix the issue, but introduced randomness in the build. Steve Smith is working on a fix. @davidbeckingsale please consult Steve's solution so Umpire does not hit the same issue. https://github.com/llnl/RAJA/pull/2068

davidbeckingsale:
> These changes look reasonable to me. I'm working with some others to test if it fixes the resident page size growth we were seeing.

@adayton1 I will wait until you have confirmed this fixes the growth you saw, please let me know once you have confirmation.

adayton1:
> > These changes look reasonable to me. I'm working with some others to test if it fixes the resident page size growth we were seeing.
> 
> @adayton1 I will wait until you have confirmed this fixes the growth you saw, please let me know once you have confirmation.

Initial testing is showing the Rss growth has flattened instead of slowing growing! I want to do a little more testing just to make sure this didn't slow down run time by any significant amount.

davidbeckingsale:
Sounds good, we will wait on your approval to merge.

adayton1:
The performance is pretty bad. I'm seeing a 60% slowdown in cycle time.

src/umpire/util/FixedMallocPool.cpp
```diff
@@ -33,15 +36,16 @@ inline unsigned int FixedMallocPool::index_from_addr(const FixedMallocPool::Pool
...
+      free_list(static_cast<unsigned int>(objects_per_pool)),
+      num_used(0)
 {
 }
 
 FixedMallocPool::FixedMallocPool(const std::size_t object_bytes, const std::size_t objects_per_pool)
     : m_obj_bytes(object_bytes), m_obj_per_pool(objects_per_pool), m_data_bytes(m_obj_bytes * m_obj_per_pool), m_pool()
 {
+  // Free-list links are stored in the slots themselves
+  UMPIRE_ASSERT(object_bytes >= sizeof(unsigned int));
```
MrBurmark, line 49: If you're going to use `unsigned int` for the linked list, and `size_t` for objects_per_pool should you check that `objects_per_pool <= UINT_MAX`.

adayton1:
> The performance is pretty bad. I'm seeing a 60% slowdown in cycle time.

False alarm. I did some fresh builds and retested. I believe the performance is approximately the same. I saw some cases where the cycle time was about 10% faster with this branch, and some cases where it was a few percent slower. But there seems to be some variability in the run time of that problem and this branch seems to fit within the noise. If anything I believe it to be slightly faster.

tests/unit/util/fixed_malloc_pool_tests.cpp
```diff
@@ -0,0 +1,157 @@
...
+  ASSERT_EQ(pool.allocate(object_bytes), b);
+  ASSERT_EQ(pool.allocate(object_bytes), a);
+
+  pool.deallocate(a);
+  pool.deallocate(b);
+}
+
+TEST_F(FixedMallocPoolTest, Churn)
+{
+  // Regression test for the eager-initialization behavior fixed after
```
adayton1: #1082 was never merged, so I think this comment is a little misleading. Can you update it?

adayton1 (dismissed):
I think one test comment needs to be updated, but otherwise this looks good! This fixes the RSS growth we were seeing and the performance is within the noise.

src/umpire/util/FixedMallocPool.cpp
davidbeckingsale: Good catch — added `UMPIRE_ASSERT(objects_per_pool <= std::numeric_limits<unsigned int>::max())` in the constructor alongside the existing `object_bytes` check.

tests/unit/util/fixed_malloc_pool_tests.cpp
davidbeckingsale: Updated — the comment now describes the eager free-list-initialization behavior directly instead of referencing #1082.

adayton1 (dismissed):

src/umpire/util/FixedMallocPool.cpp
```diff
@@ -33,15 +37,18 @@ inline unsigned int FixedMallocPool::index_from_addr(const FixedMallocPool::Pool
...
+  // Slot indices (free-list links and num_used) are unsigned int
+  UMPIRE_ASSERT(objects_per_pool <= std::numeric_limits<unsigned int>::max());
```
MrBurmark:
```suggestion
  UMPIRE_ASSERT(objects_per_pool <= static_cast<std::size_t>(std::numeric_limits<unsigned int>::max()));
```
Would an explicit cast here be better?

adayton1 (approved):

MrBurmark (approved):

PR #1101, opened 2026-07-31
Resolve CMake Target Errors

kab163:


kab163:
@adayton1 - do you think this would resolve the cmake target issues?

adayton1:
> @adayton1 - do you think this would resolve the cmake target issues?

To be honest, I think you are making this too complicated. If you want to copy properties from one target to another, Danny's solution was fine. Though testing both targets is a good idea.

If you are wanting to go another route, you could install the targets twice, once as umpire and once with the umpire:: namespace (there should be a BLT argument somewhere that you can use to install a namespaced target. Changes would be needed in src/umpire/src/CMakeLists.txt for this approach. Then you'd have to import both files.

kab163:
> > @adayton1 - do you think this would resolve the cmake target issues?
> 
> To be honest, I think you are making this too complicated. If you want to copy properties from one target to another, Danny's solution was fine. Though testing both targets is a good idea.
> 
> If you are wanting to go another route, you could install the targets twice, once as umpire and once with the umpire:: namespace (there should be a BLT argument somewhere that you can use to install a namespaced target. Changes would be needed in src/umpire/src/CMakeLists.txt for this approach. Then you'd have to import both files.

Ok, we can take another look next week

kab163:
Closing in favor of #1100

PR #1102, opened 2026-08-04
Export legacy CMake targets

adayton1:
Having namespaced targets such as "umpire::umpire" is the current best practice (because if it is not found, it is an error rather than just inserting -lumpire on the link line).

But we still want to support the legacy "umpire" target for a transition period. We tried making "umpire" an alias to "umpire::umpire", but downstream projects were setting properties on the "umpire" target, so this approach fails because properties can't be set on an alias target.

We tried making a new target and copying properties manually, but that is verbose and error prone. More checking was needed.

An alternative approach is to create an INTERFACE library called umpire and link it to umpire::umpire. But then it gains properties only transitively, not on the target itself.

To have the exact same behavior for the "umpire" target as before, we can export both namespaced and non-namespaced targets fairly easily, and that is what this pull request does.

### DIFF
--- a/.github/hubcast.yml
+++ b/.github/hubcast.yml
@@ -5,7 +5,7 @@ Repo:
 
   check_name: gitlab-ci
 
-  check_types: [pipelines, child-pipelines]
+  check_types: [pipeline, child-pipelines]
 
   delete_closed: true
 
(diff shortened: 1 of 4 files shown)


rhornung67 (approved):
If this works, I think it is a good deprecation transition.

adayton1:
This works for a large application that depends on umpire directly and transitively.

davidbeckingsale (approved):

adayton1 (commented):
@lc-hubcast approve

PR #1103, opened 2026-08-04
Fix hubcast config

adayton1:


kab163 (approved):

alecbcs:
(The error message here is a bit misleading, since it's still reading the config off of the default branch not this PR. The config looks correct to me. This is something we'll work on fixing.)

PR #1105, opened 2026-08-05
Fix legacy CMake targets

adayton1:
Apparently there are issues when trying to export targets twice even if they end up with different names. So now we try creating an INTERFACE library called "umpire" and link it to "umpire::umpire". We also copy some INTERFACE properties from "umpire::umpire" to "umpire". The same process is used for creating an "umpire_device" target from the "umpire::umpire_device" target.

The following is an error from exporting the "umpire" target twice:

```
CMake Error: install(EXPORT "chai-targets" ...) includes target "chai" which requires target "umpire" that is not in this export set, but in multiple other export sets: lib/cmake/umpire/umpire-targets.cmake, lib/cmake/umpire/umpire-non-namespaced-targets.cmake.
 
An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "umpire" target to a single export.
```

If the "umpire" target is not created as an "IMPORTED" library, then CHAI fails to export it's own targets:

```
CMake Error: install(EXPORT "chai-targets" ...) includes target "chai" which requires target "umpire" that is not in any export set.
```

If INTERFACE_INCLUDE_DIRECTORIES are not copied from "umpire::umpire" to "umpire", then downstream clients can't locate the umpire headers.

Link information is supposed to be transitive, so we link "umpire" to "umpire::umpire" and don't copy INTERFACE properties related to linking.

CMake has no internal way to just copy all target properties, or even to get a list of all the properties on a target. We just copy a few of the most common and important ones. INTERFACE_INCLUDE_DIRECTORIES and INTERFACE_COMPILE_DEFINITIONS are known to be needed.

See #1102, #1101, #1100, #1083, #1067

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,13 +203,6 @@ install(
   NAMESPACE umpire::
   DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/umpire)
 
-if (UMPIRE_ENABLE_LEGACY_CMAKE_TARGETS)
-  install(
-    EXPORT umpire-targets
-    FILE umpire-non-namespaced-targets.cmake
-    DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/umpire)
-endif()
-
 if (UMPIRE_ENABLE_TESTS)
   add_subdirectory(tests)
 endif ()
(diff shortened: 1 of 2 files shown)


rhornung67 (dismissed):

kab163 (dismissed):

kab163:
@adayton1 - Did this work for your use case? I want to double check before we merge these changes.

adayton1:
It's still failing in some cases - just trying to work through them.

adayton1:
This is almost ready. If all tests pass here, in the CHAI release candidate, and in a downstream application, then it should be good to go.

adayton1:
Should we make this a patch release or do that in a separate PR?

adayton1 (commented):
@lc-hubcast approve

kab163:
> Should we make this a patch release or do that in a separate PR?

Separete PR

kab163 (approved):

